PR #1, opened 2022-12-08
Restricted search space for hypothesis

jfennick:


bdw2292 (commented):
Why is biobb cwl being changed?

bdw2292 (approved):
Most everything looks like what we discussed yesterday and makes sense!

jfennick:
> 

Since all of our sub-schema are mutually exclusive, I've changed all instance of `oneOf` to `anyOf` for performance reasons. See attached minimal example

```
import unittest

from hypothesis import given, settings
from hypothesis_jsonschema import from_schema

max_num = 1000
schemas_ints_pos = [{'type': 'integer', 'const': n} for n in range(1, 1 + max_num)]
schemas_ints_neg = [{'type': 'number', 'const': -n} for n in range(1, 1 + max_num)]
schema_ints_pos = {'anyOf': schemas_ints_pos}
schema_ints_neg = {'anyOf': schemas_ints_neg}

schema_ints_enum = {'type': 'integer', 'enum': [n for n in range(1, 1 + max_num)] + [-n for n in range(1, 1 + max_num)]}
schema_ints_flat = {'anyOf': schemas_ints_pos + schemas_ints_neg}
schema_ints_tuple_of_lists = {'anyOf': [schema_ints_pos, schema_ints_neg]}
schema_ints_list_of_tuples = {'anyOf': [{'anyOf': [pos, neg]} for pos, neg in zip(schemas_ints_pos, schemas_ints_neg)]}

# NOTE: Since each sub-schema is mutually exclusive, it is logically equivalent
# to use oneOf or anyOf. However, there is a massive performance difference!
# (This is because for oneOf, it needs to validate against ALL of the
# sub-schema and check that one and only one schema matches, thus it is linear
# in the length of the list, whereas anyOf is approximately constant time.)
# There is an additional 10X performance improvement if you can use enum.

strategy = from_schema(schema_ints_enum) # 1.41s (!!!)
#strategy = from_schema(schema_ints_flat) # oneOf 42.22s anyOf 10.72s
#strategy = from_schema(schema_ints_tuple_of_lists) # oneOf 74.16s anyOf 10.76s
#strategy = from_schema(schema_ints_list_of_tuples) # oneOf 93.18s anyOf 11.11s

class TestInts(unittest.TestCase):
    @given(strategy)
    @settings(max_examples=2*max_num) # Do not short circuit
    def test_ints(self, value):
        assert (-max_num) <= value <= max_num

if __name__ == '__main__':
    unittest.main()
```

PR #9, opened 2023-01-28
Gromacs system setup workflow

ndonyapour:
Hello,

This PR contains a  system setup workflow using `Gromacs`.

Thanks, 
Nazanin

### DIFF
--- a/cwl_adapters/python3_calculate_net_charge.cwl
+++ b/cwl_adapters/python3_calculate_net_charge.cwl
@@ -3,28 +3,24 @@ cwlVersion: v1.0
 
 class: CommandLineTool
 
-label: Calculate the Total Charge 
+label: Calculate the net Charge
 
 doc: |-
-  Calculate the total charge of a given ligand
+  Calculate the net charge of a given ligand
 
 baseCommand: python3
-#arguments: ["../scripts/examples/calculate_net_charge.py"]
 
-# hints:
-#   DockerRequirement:
-#     dockerPull: jakefennick/scripts
 requirements:
   InlineJavascriptRequirement: {}
+  DockerRequirement:
+    dockerPull: ndonyapour/scripts
 
 inputs:
   script:
-    type: File
-    format:
-    - edam:format_2330
+    type: string
     inputBinding:
       position: 1
-      #default: ../scripts/examples/calculate_net_charge.py
+
   input_path:
     label: Path to the input file
     doc: |-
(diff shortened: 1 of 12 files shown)


jfennick:
@ndonyapour I have already reviewed this code with you offline numerous times. As I mentioned before, most of this code already exists. Before we can merge, you need to justify why this PR should exist, and then get rid of the duplication.

1. Extract the biobb formal schema improvements into a separate PR.
2. Extract the trivial whitespace indentation changes in modeling.yml into a separate PR.
3. Your code does a vacuum minimization of the ligand only. https://github.com/PolusAI/workflow-inference-compiler/pull/9/files#diff-2a40af1e72e934be593abbba93ad31dd48cbf918ffdb7595012e59569f5f9ba3R2 This is probably unnecessary, and should be a separate PR if justified.
4. Your code adds restraints to the topology, presumably for minimization. https://github.com/PolusAI/workflow-inference-compiler/pull/9/files#diff-359345ace82c79292c373e50bd20496996a30b095480776a36110401ae7d601bR23  Again, this probably unnecessary and should be a separate PR if justified.
5. Regarding acpype:
    1. Your code uses RDKit to calculate the Gasteiger partial charges, then sums them and passes the total to net_charge. Then acpype calculates the partial charges again using ambertools.
    2. The existing code allows the per-atom partial charges to be directly specified with charge_method: user. These approaches can be combined.
    3. More importantly, your ligand setup workflow does not include the acpype workarounds I very painfully discovered here https://github.com/PolusAI/workflow-inference-compiler/blob/master/examples/docking/gen_topol_params.yml#L43 As you can see, for ligands only, the topology file acpype generates doesn’t actually work with explicit solvent! Rather, it only works if you combine it with a protein topology file, which adds the necessary include lines.

Part of the complexity of this PR is because you are trying to rigidly follow this tutorial exactly https://mmb.irbbarcelona.org/biobb/workflows/tutorials/protein-complex_md_setup#ligtop Unless your goal is to 100% exactly reproduce a result, you shouldn't feel obligated to do that, especially if the differences are only w.r.t. minimization.

At a high level, what is the goal of this PR? What are you trying to achieve? *What is the call site?* From our previous discussions, it sounds like you were simply trying to equilibrate a protein ligand complex as a starting point for your other simulation work. But again, tutorial.yml already does that (you just need to specify backend: complex).

@hsidky As the team lead, you need to be aware of what features already exist so we do not spend time duplicating effort.

PR #12, opened 2023-02-14
Updated documentation and added docker runtime check.

jfennick:


jfennick:
test_fuzzy_compile failed simply due to a timeout. You can re-run the action if you'd like, but the occasional timeout due to randomly generating very large workflows is expected and fine. Okay to merge in my opinion.

PR #16, opened 2023-02-22
Added fastapi to wic

mohamedOuladi:
This PR includes the following:
- added api (fastapi) 
- integrated auth
- added Dockerfile.api


jfennick:
FYI I'm still trying to figure out a way to test this new code, if possible.

mohamedOuladi:
> FYI I'm still trying to figure out a way to test this new code, if possible.

you can test it locally by running `uvicorn api:app --reload --port 3000`. You will need to add an access token to your auth headers when making a POST request to route `/compile`.

jfennick:
I should have clarified that I was asking about how to incorporate it into the automated CI. We can probably still merge without, but if there's a way to do it without too much effort that would be excellent.

jfennick (approved):
LGTM. I think we can add the automated testing in the next PR.

api.py
```diff
@@ -0,0 +1,147 @@
...
+
+    Returns:
+        Dict[str, str]: {"message": "The api has 2 routes: compile and inference"}
+    """
+    return {"message": "The api has 2 routes: compile and inference"}
+
+
+@app.post("/compile")
+@authenticate
+async def compile_wf(request: Request) -> Json:
```
abf7d, line 59: Consider using a router to organize your endpoints if you expect the number of endpoints to grow:
look at [section 3](https://levelup.gitconnected.com/5-advance-features-of-fastapi-you-should-try-7c0ac7eebb3e)

abf7d, line 59: Return type should be a strongly typed model instead of just Json. Look into using pydantic to enforce types

auth/settings.py
```diff
@@ -0,0 +1,81 @@
...
+        "name": "bins",
+        "description": (
+            "Endpoints for binning data, including endpoints for accessing masks."
+        ),
+    },
+    {
+        "name": "selections",
+        "description": "Endpoints for selecting and subsetting data.",
+    },
+]
```
abf7d, line 76: I don't think you are binning or selecting data, this was for the cytoexplorer histogram. You should probably remove these

api.py
```diff
@@ -0,0 +1,147 @@
...
+
+        # TODO: get version from ict plugin specs
+        run_val = f'plugin:{stem}:{__version__}'
+        step_name_i = utils.step_name_str(yaml_stem, i, step_key)
+        step_name_i = step_name_i.replace('.yml', '_yml')  # Due to calling remove_dot_dollar above
+        cwl_tree_run['steps'][step_name_i]['run'] = run_val
+
+    compute_workflow = {
+        "name": yaml_stem,
+        # "version": __version__, # no version for workflows
```
abf7d, line 138: Remove commented code

api.py
```diff
@@ -0,0 +1,147 @@
...
+async def compile_wf(request: Request) -> Json:
+    """The compile route compiles the json object fom workflow builder ui which contains steps built in the ui.
+
+    Args:
+        request (Request): request object from workflow builder ui
+
+    Returns:
+        compute_workflow (JSON): workflow json object ready to submit to compute
+    """
+    print('----------Run Workflow!---------')
```
abf7d, line 68: Maybe add logger instead of print statement, see cyto-explorer for an example

abf7d (changes requested):
Some of these can be longer term goals. The things needed to be done are remove commented code and remove Tags.

api.py
jfennick: Camilo is looking into Pydantic types. Does that need to be done for this PR, or can we add Pydantic later?

abf7d: If you guys want to merge this quickly it can be done later.

abf7d (approved):

api.py
mohamedOuladi: I am expecting to have no more than 2 routes for now (second route will be `/inference` with few lines of code in `api.py`). If it gets messier, I will consider using a router.

api.py
mohamedOuladi: An appropriate logger is to be added in a future PR.

PR #20, opened 2023-03-07
Global config

jfennick:
This is the first step towards "conda environments" for wic. There are many tricky path changes, so you will want to actually run this code in your reviews. Also note that for this particular PR, not everything may be caught by the unit tests, so some manual testing may be required.


jfennick:
Can one or more of you please review this PR sometime next week? It's blocking #22

To be more specific, thus far I have only added the autopep8 configuration files to #22. I specifically haven't actually re-formatted any of the python files, because this needs to be merged first so we don't unnecessarily generate massive merge conflicts.

misterbrandonwalker (approved):

PR #28, opened 2023-03-15
Gromacs amber schemas

ndonyapour:
Hi,

This PR includes a set of functions that define schemas for some of Amber and Gromacs Biobb functions.

Thanks!


jfennick (approved):
My only comment is that it is unfortunate that biobb amber uses `forcefield` while biobb gromacs uses `force_field`. Perhaps we can suggest a change upstream but for now I think this looks good to me.

PR #32, opened 2023-03-17
Add explicit dependency on Numpy

ndonyapour:
Hi, 

This PR adds explicit installing of `Numpy` using `conda` to prevent `pytest` from failing.

Thanks,
Nazanin

### DIFF
--- a/conda_devtools.sh
+++ b/conda_devtools.sh
@@ -50,3 +50,9 @@ $CONDA install -y -c conda-forge shellcheck
 $CONDA install -y -c conda-forge compilers
 $CONDA install -y -c conda-forge pip
 #pip cache purge
+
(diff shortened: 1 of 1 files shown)


jfennick (approved):
LGTM. Installing numpy via conda instead of pip building the binary wheel has the benefit of shaving another 3-4 mins off the CI time.

jfennick:
Actually, the story seems to be more complicated when considering Windows. See https://github.com/PolusAI/workflow-inference-compiler/pull/33/files#diff-6ad3080e3a5d9006fcf8e18df47a4fc8a71d5575a9d5cf55a6776843d0f6d5cdR24 That includes this fix and also the fix for windows, so I recommend closing this PR in favor of #33.

ndonyapour:
This works for Linux and Mac system but creates problems for Windows and @jfennick added the fix in another PR.

PR #44, opened 2023-03-28
add box_vector_lengths to gmx_editconf

ndonyapour:
Hello,

This PR adds an option to the `gmx_editconf` wrapper that enables the use of box vector lengths for setting up a system.

Thanks,
Nazanin

### DIFF
--- a/cwl_adapters/gmx_editconf.cwl
+++ b/cwl_adapters/gmx_editconf.cwl
@@ -71,11 +71,20 @@ inputs:
     inputBinding:
       position: 5
       prefix: -box
+      
+  box_vector_angles:
+    type:
+      - "null"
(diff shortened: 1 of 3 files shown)


cwl_adapters/gmx_editconf.cwl
```diff
@@ -52,23 +52,39 @@ inputs:
...
+      position: 5
+      prefix: -box
+      
+  box_vector_angles:
+    type:
+      - "null"
+      - type: array
+        items: float
+    inputBinding:
+      position: 5
```
jfennick, line 81: Should this be position 6? (and then align_principal_axes position 7)

jfennick (changes requested):

PR #59, opened 2023-06-08
update biobb versions

jfennick:


jfennick:
NOTE: This does not update biobb_gromacs. We need to update gromacs carefully because our EC2 instances (i.e. our self-hosted runners) only support CUDA 11.4 which is compatible with gromacs 2121.4 but not gromacs 2022.2

hsidky (approved):

PR #67, opened 2023-06-14
update gromacs from 2021.4 to 2022.2 and 2023.2

jfennick:
**Do NOT merge this yet!**

Our EC2 instances (i.e. our self-hosted runners) are stuck on an older version of CUDA which is not compatible with gromacs 2022.2


jfennick:
See linked PR https://github.com/PolusAI/mm-workflows/pull/6

PR #80, opened 2023-08-18
Change repo-dispatch payload name reference.

ywang271828:
- Fix the wrong variable reference. `mm-workflows` uses `ref` in the `client_payload` of `repository_dispatch` webhook, rather than `ref_name`. The wrong reference evaluated `sender_repo_ref` to `null` and makes the `wic_owner` & `mm_workflows_owner` to `PolusAI` and `wic_ref` (`mm_workflows_ref`) to `master` (`main`).


ywang271828:
Fix the reference error on mm-workflows side. See: https://github.com/PolusAI/mm-workflows/pull/12.

PR #85, opened 2023-08-22
alternative explicit edge syntax

jfennick:


jfennick:
Closed in favor of #86

PR #130, opened 2023-12-21
Fix weekly CI pytest names.

ywang271828:
In `pytest`, `-k` specifies an "expression" (or substring) instead of exact strings of unittest names. If we have two tests with names: `test_a` and `test_a_inlined`, running `pytest -k a` will run both tests since `a` is also a substring of `test_a_inlined`. 

Use `pytest -k "a not inlined"` to exclude the substring `inlined`.


jfennick:
Closing in favor of https://github.com/PolusAI/workflow-inference-compiler/pull/133

Yes, the -k option of pytest is for expressions. But instead of introducing boolean algebra into our test names ... we can simply use unique names. See the existing usage of the _on_push and _weekly suffixes.

PR #131, opened 2023-12-21
Write _inline suffix to inlined workflow.

ywang271828:
In `run_local.py::run_local(): Line 95`, the top-level workflow file is renamed with suffix `_inline` but `write_to_disk()` does not add this suffix when inlining is turned on. This PR modified the `write_to_disk()` to distinguish top-level workflow and whether inline is used.

Note: 
1. Only top-level workflow is suffixed with `_inline`.
2. An alternative solution is to remove line 95 of `run_local.py`, and write inlined version of compiled workflow also in name like `tutorial.cwl`, rather than `tutorial_inline.cwl`.

This will fix a weekly CI error and allow `--run_local` and `--cwl_inline_subworkflows` to be used together.


jfennick:
Closing in favor of https://github.com/PolusAI/workflow-inference-compiler/pull/134

Rather than distribute the _inline appending around, we should keep all of the inline logic together.

PR #147, opened 2024-01-16
Filter array

misterbrandonwalker:


jfennick:
There should not be two nearly identical but separate tools that only differ by the type. Use `cwlVersion: v1.2` and the Any type.

PR #150, opened 2024-01-18
use current value for links; use --run_local

jfennick:


src/wic/api/pythonapi.py
```diff
@@ -223,9 +223,13 @@ def __setattr__(self, __name: str, __value: Any) -> Any:  # pylint: disable=R171
                                 f"links must have the same input type. "
                                 f"cannot link {local_input.name} to {__value.name}"
                             )
-                        tmp = f"{__name}{self.cwl_name}"
-                        local_input._set_value(f"*{tmp}", check=False, linked=True)
-                        __value._set_value(f"&{tmp}", check=False, linked=True)
+                        # Use the current value so we can exactly reproduce hand-crafted yml files.
+                        # (Very useful for regression testing!)
+                        current_value = __value.value
+                        if current_value is None:
```
agerardin, line 229: I think changing the value is fine. 

Now checking for None implies a ordering in our configuration flow (the value we link to must not be None). I defer to you guys on this but note this is an extra constraint in how we build stuff.

agerardin (commented):
See my comments on the added check.

For the rest, looks all good.

src/wic/api/pythonapi.py
agerardin: Actually, after writing some tests, it seems this check actually breaks the api code. Try running this :

```
import wic.api.pythonapi as api
import polus.plugins as pp
from pathlib import Path

bbbc_download_url = "https://raw.githubusercontent.com/saketprem/polus-plugins/bbbc_download/utils/bbbc-download-plugin/plugin.json"
filerenaming_url = "https://raw.githubusercontent.com/PolusAI/polus-plugins/master/formats/file-renaming-plugin/plugin.json"
ome_converter_url = "https://raw.githubusercontent.com/PolusAI/polus-plugins/master/formats/ome-converter-plugin/plugin.json"
montage_url = "https://raw.githubusercontent.com/PolusAI/polus-plugins/master/transforms/images/montage-plugin/plugin.json" 
# need updates
image_assembler_url = "https://raw.githubusercontent.com/agerardin/polus-plugins/new/image-assembler-plugin-1.4.0-dev0/transforms/images/image-assembler-plugin/plugin.json"
precompute_slide_url = "https://raw.githubusercontent.com/agerardin/polus-plugins/update/precompute-slide-fp2/visualization/precompute-slide-plugin/plugin.json"

WIC_STAGING = Path() / "WIC_STAGING"
WIC_STAGING.mkdir(parents=True, exist_ok=True)
CWLTOOL_PATH = (WIC_STAGING / "CWL_TOOLS").resolve()
CWLTOOL_PATH.mkdir(exist_ok=True)
WORKFLOW_PATH = (WIC_STAGING / "WORKFLOWS").resolve()
WORKFLOW_PATH.mkdir(exist_ok=True)
RUNS = (WIC_STAGING / "RUNS").resolve()
RUNS.mkdir(exist_ok=True)


# TODO REMOVE. This is from polus plugins. Polus plugins needs to fix this.
# The problem being that names are rewritten in polus plugins but the manifest is not updated.
# We should either enforce a strict name, generate a standardized handle, or update the manifest
# we send back when submitting plugins.
def name_cleaner(name: str) -> str:
    """Generate Plugin Class Name from Plugin name in manifest."""
    replace_chars = "()<>-_"
    for char in replace_chars:
        name = name.replace(char, " ")
    return name.title().replace(" ", "").replace("/", "_")

def create_step(url: str) -> api.Step:
    manifest = pp.submit_plugin(url)
    plugin_classname = name_cleaner(manifest.name)
    plugin_version = manifest.version.version
    cwl_tool = pp.get_plugin(plugin_classname, plugin_version).save_cwl(
    CWLTOOL_PATH / f"{plugin_classname}.cwl"
    )
    step = api.Step(cwl_tool)
    return step

if __name__ == "__main__":
    bbbc = create_step(bbbc_download_url)
    rename = create_step(filerenaming_url)
    ome_converter = create_step(ome_converter_url)
    montage = create_step(montage_url)
    image_assembler = create_step(image_assembler_url)
    precompute_slide = create_step(precompute_slide_url)

    # BBBCDownload
    bbbc.name = "BBBC001"
    bbbc.outDir = Path(RUNS)

    # FileRenaming config
    rename.filePattern = ".*_{row:c}{col:dd}f{f:dd}d{channel:d}.tif"
    rename.outFilePattern = "x{row:dd}_y{col:dd}_p{f:dd}_c{channel:d}.tif"
    rename.mapDirectory = "map"
    rename.inpDir = bbbc.outDir

    # OMEConverter
    ome_converter.filePattern = ".*.tif"
    ome_converter.fileExtension = ".ome.tif"
    ome_converter.inpDir = rename.outDir

    # Montage
    montage.filePattern = "d1_x00_y03_p{p:dd}_c0.ome.tif"
    montage.layout = "p"
    montage.inpDir = ome_converter.outDir

    # Image Assembler
    image_assembler.imgPath = montage.outDir
    image_assembler.stitchPath = ome_converter.outDir

    # Precompute Slide
    precompute_slide.pyramidType = "Zarr"
    precompute_slide.imageType = "image"
    precompute_slide.inpDir = image_assembler.outDir
    precompute_slide.outDir = RUNS


    steps = [bbbc, rename, ome_converter, montage, image_assembler, precompute_slide]
    workflow = api.Workflow(steps, "experiment", WORKFLOW_PATH)
    workflow._save_yaml()

    # TODO CHECK compiling errors are not propagated back to us, so we don't know where to look
    workflow.compile()

    #TODO CHECK run hangs.
    workflow.run(debug=True)
```

jfennick: I think the reason the None check 'breaks' this is because you are only setting an explicit outDir on the first and last step. For all intermediate steps outDir is uninitialized, and thus None. The reason I raise InvalidLinkError here is because the --outDir flag is indeed a required argument for all of the polus plugins. Note that the plugin manifests are incorrect! They do not even show --outDir as an input, but if you run the docker containers with --help, --outDir is indeed a required argument. So my preference would be to fix the plugin manifests and leave this code as-is. Alternatively, for now I could remove the None check and fallback to the old method. Upto you.

agerardin: For setting intemediary outputs, I may misinterpret, but I thought this line of code was exactly doing that:

```python
                        local_input._set_value(f"*{current_value}", check=False, linked=True)
                        __value._set_value(f"&{current_value}", check=False,linked=True)
```

For the wipp manifests, you are correct, the semantics of the inputs and outputs is slightly different than cwl. Basically, outputs (I guess we should have some required flag there) need to be specified and so should be provided as well to the tool. In my mind, adding them again as inputs is a matter of taste (if the expected behavior is well defined). But in any case, we don't control this spec. But we do control ICT so let's make sure things are defined the way we choose there! I actually not a bad idea to align with cwl since this is our adopted standard.

jfennick: For now, I've added a fallback mechanism which uses the value supplied by the user if any, and then uses an autogenerated name.

PR #152, opened 2024-01-22
feat(pythonapi): add Pydantic V2 support

camilovelezr:
This PR:
* Adds parallel support for Pydantic V2 and V1
* Fixes `Workflow.compile()`:
  * In #150 the function `._savel_all_cwl()` was removed and this is necessary to compile the Workflow
 * Does a general refactoring of `Step` and `Workflow`:
    * `Step.cwl` is now `Step.clt`
    * `Step.cwl_name` is now `Step.clt_name`
    * `Step.cwlVersion` is now `Step.cwl_version`
    * `Workflow` does not need a `path` to be initialized
    * `Workflow.step()` is now called `Workflow.append()` and performs type checking.
  * `Workflow` now automatically looks for the user's WIC directory. It is used to run the `wic --yaml` (compilation) command from that directory. 
  *This is a *workaround* since `cwl_dirs.txt` creates a lot of issues when you run `wic --yaml` from different directories.*
    * Before: `Workflow([step1, step2], name="Workflow_1", path="/Users/wicuser/workflow-inference-compiler")`
    * Current: `Workflow([step1, step2], name="Workflow_1")`


jfennick:
Regarding your last bullet point:

It was never true that wic had to be run from any particular directory, provided your auto-discovery configuration in `~/wic/` is correct.
Please read https://github.com/PolusAI/workflow-inference-compiler/blob/master/docs/userguide.md#auto-discovery

In addition to the pytest and pylint examples, you can think of auto-discovery as similar to PATH or CLASSPATH environment variables. It's really just a default list of locations to look for things.

In particular, if you do not have a `~/wic/` directory (i.e. for new users), wic will automatically create them for you with sensible defaults, and it will repeatedly print the following disclaimers / warnings / instructions:
https://github.com/PolusAI/workflow-inference-compiler/blob/master/src/wic/utils.py#L521-L523
`Please check this file and make sure that the paths in it are correct.`

You probably didn't see these warnings because you have been capturing all stdout:
https://github.com/PolusAI/workflow-inference-compiler/blob/1724d9911133809379e83b07dcb7403fc2681870/src/wic/api/pythonapi.py#L357

In general, it is not possible to know how your machine is setup (remember, auto-discovery is analogous to PATH), so the only sensible thing to do is assume all of your projects are in sibling directories and use `../`.

TL;DR: The **_first_** time you run wic (and the first time only), you should run it from one of the sibling project directories so that you don't have to edit the global auto-discovery config files in `~/wic`. After that, you can run wic in whatever directory you like.

src/wic/api/pythonapi.py
```diff
@@ -155,61 +167,80 @@ def _(val: bool) -> bool:
...
+        if not isinstance(clt_path, (Path, str)):
             raise TypeError("cwl_path must be a Path or str")
-        if isinstance(cwl_path, str):
-            cwl_path_ = Path(cwl_path)
-        else:
-            cwl_path_ = cwl_path
+        clt_path_ = Path(clt_path) if isinstance(clt_path, str) else clt_path
         try:
-            cwl = load_document_by_uri(cwl_path_)
+            clt = load_document_by_uri(clt_path_)
```
jfennick, line 193: This line of code may not do what you think it does. Please see the validate_cwl function here:
https://github.com/PolusAI/workflow-inference-compiler/blob/master/src/wic/plugins.py#L57-L61

It is very easy to look at tutorials and use the seemingly-straightforward code from them, but there are often nasty subtle edge cases that need to be considered (and have already been painfully discovered and solved!).

Moreover, there is an even nastier issue in this case w.r.t. skip_schemas, which you may discover if you turn on your VPN.

So let's try to explore the codebase and make ourselves aware of the existing production-ready code.

src/wic/api/pythonapi.py
```diff
@@ -335,24 +368,30 @@ def yaml(self) -> dict[str, Any]:
...
-        """Compile Workflow using WIC."""
-        # The CommandLineTools should already be on disk, and moreover when
-        # self.path = '.' this will overwrite them!
-        # self._save_all_cwl()
+        """Compile Workflow using WIC.
+
+        Returns path to compiled CWL Workflow.
+        """
+
+        self._save_all_cwl()
```
jfennick, line 394: This line still erroneously overwrites CWL files if they are already on disk. (See comment above which you have deleted.)

This is a problem if the new CWL file contents are loaded from a remote URI (and/or otherwise incorrectly autogenerated) and then the CWL file which has been hand-crafted by an expert (i.e. me) to be correct gets overwritten.

PR #153, opened 2024-01-24
test(pythonapi): add initial tests for Python API

camilovelezr:
*To be merged after #152*
This PR adds an initial set of tests for the Python API

### DIFF
--- a/tests/resources/assembler.cwl
+++ b/tests/resources/assembler.cwl
@@ -1,0 +1,41 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.2
+class: CommandLineTool
+
+requirements:
+  DockerRequirement:
+    dockerPull: polusai/image-assembler-plugin:1.2.1-dev0
+  InitialWorkDirRequirement:
+    listing:
+    - writable: true
+      entry: $(inputs.outDir)
+  InlineJavascriptRequirement: {}
+
+inputs:
+  imgPath:
+    type: Directory
+    inputBinding:
+      prefix: --imgPath
+  outDir:
+    type: Directory
+    inputBinding:
+      prefix: --outDir
+  preview:
+    type: boolean?
+    inputBinding:
+      prefix: --preview
+  stitchPath:
+    type: Directory
+    inputBinding:
+      prefix: --stitchPath
+  timesliceNaming:
+    type: boolean?
+    inputBinding:
+      prefix: --timesliceNaming
+
+outputs:
+  outDir:
+    type: Directory
+    outputBinding:
+      glob: $(inputs.outDir.basename)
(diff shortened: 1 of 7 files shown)


jfennick:
https://workflow-inference-compiler.readthedocs.io/en/latest/dev/gitetiquette.html#pull-requests
Please see the WIC policy w.r.t. pull request size.
This PR is very large (even ignoring the ~250 line diff from the stacked PR), and more importantly I believe there is a fundamentally better approach.
Please give me some time to demonstrate my preferred approach, and then we can revisit this PR.

agerardin:
On my end, I am ok with the general approach of having specific test fixtures and unit tests.
However I would be great to provide a bit more structure and clarify test intents.
A conftest.py with fixtures
One or more directories if needed but at least several files for different  tests (ex: testing the linking logic, the step input validation etc..)
Also better names to clarify the test intent (along the docstring)... but you knew I would mention that 😄

jfennick:
PR #165 is now merged. Please take a look at what it does and how it works. In short:

It clones the image-workflows repo, which allows us to clearly separate both
1. the workflows
2. the underlying CWL CommandLineTools
from the code which tests them.

This allows us to do [property-based testing](https://github.com/PolusAI/workflow-inference-compiler/blob/master/docs/validation.md#property-based-testing)
`on the entire WIC language, for every single tool and workflow simultaneously`.
This in turn reduces the need for [unit-tests](https://github.com/PolusAI/workflow-inference-compiler/blob/master/docs/validation.md#unit-tests).

So similar to how we have done it in the mm-workflows repo
1. We should not be storing workflows in this repository.
2. We should not be storing CWL CommandLineTools in this repository.

Also notice that the CWL CommandLineTools in this PR are subtly different from the ones in [my image-workflows PR](https://github.com/PolusAI/image-workflows/pull/4/). Just two examples (but there are more):
1. [this line](https://github.com/PolusAI/image-workflows/pull/4/files#diff-50a9eeb882e09737b04e96d7a333ddf13184919032008004b85d3d4dd2dd2d07R21) demonstrates that there are additional considerations for specific inputs for legacy tools.
2. [these lines](https://github.com/PolusAI/image-workflows/pull/4/files#diff-99967b2004d431a003ac7e528ad6000fc05fb87333f945b60404e56542dc1003R24-R34) demonstrate that CWL has a different philosophy when it comes to environment variables.

Most importantly, since this information is simply not present in the plugin manifest / ICT spec,
`we cannot auto-generate CWL CommandLineTools.`

Finally, since we are [still overwriting](https://github.com/PolusAI/workflow-inference-compiler/pull/152#discussion_r1484688588) the existing hand-crafted CWL CommandLineTools (i.e. from image-workflows PR 4) then these incorrect auto-generated CWL CommandLineTools are still causing problems for our internal users.

PR #170, opened 2024-02-15
fix(pythonapi): fix int, float in yaml

camilovelezr:
This fixes an issue where integers and floats where put into YML as strings.

### DIFF
--- a/src/wic/api/pythonapi.py
+++ b/src/wic/api/pythonapi.py
@@ -166,7 +166,7 @@ def _(val: int) -> int:
     return val
 
 @_value_str.register
-def _(val: float) -> int:
+def _(val: float) -> float:
     return val
 
 @_value_str.register
--- a/src/wic/api/pythonapi.py
+++ b/src/wic/api/pythonapi.py
@@ -161,13 +161,16 @@ def _value_str(val: Any) -> Union[str, bool, int, float]:
     """Return value of input as str."""
     return str(val)
 
+
 @_value_str.register
 def _(val: int) -> int:
     return val
 
+
 @_value_str.register
 def _(val: float) -> float:
     return val
+
 
 @_value_str.register
 def _(val: bool) -> bool:
--- a/src/wic/api/pythonapi.py
+++ b/src/wic/api/pythonapi.py
@@ -157,22 +157,22 @@ def _is_link(s: str) -> bool:
 
 
 @singledispatch
-def _value_str(val: Any) -> Union[str, bool, int, float]:
-    """Return value of input as str."""
+def _yaml_value(val: Any) -> Union[str, bool, int, float]:
+    """Convert value to YAML compatible value."""
     return str(val)
 
 
-@_value_str.register
+@_yaml_value.register
 def _(val: int) -> int:
     return val
 
 
-@_value_str.register
+@_yaml_value.register
 def _(val: float) -> float:
     return val
 
 
-@_value_str.register
+@_yaml_value.register
 def _(val: bool) -> bool:
     return val
 
@@ -324,7 +324,7 @@ class Step(BaseModel):  # pylint: disable=too-few-public-methods
         d = {
             self.clt_name: {
                 "in": {
-                    inp.name: _value_str(inp.value)
+                    inp.name: _yaml_value(inp.value)
                     for inp in self.inputs
                     if inp.value is not None
                 }
--- a/src/wic/api/pythonapi.py
+++ b/src/wic/api/pythonapi.py
@@ -157,10 +157,17 @@ def _is_link(s: str) -> bool:
 
 
 @singledispatch
-def _value_str(val: Any) -> Union[str, bool]:
+def _value_str(val: Any) -> Union[str, bool, int, float]:
     """Return value of input as str."""
     return str(val)
 
+@_value_str.register
+def _(val: int) -> int:
+    return val
+
+@_value_str.register
+def _(val: float) -> int:
+    return val
 
 @_value_str.register
 def _(val: bool) -> bool:


jfennick:
FYI your second and third commits fix issues that should have been found by the pre-commit hooks.

From the [readme](https://github.com/PolusAI/workflow-inference-compiler?tab=readme-ov-file#quick-start):
`pre-commit install  # Required for developers`

src/wic/api/pythonapi.py
```diff
@@ -157,11 +157,21 @@ def _is_link(s: str) -> bool:
 
 
 @singledispatch
-def _value_str(val: Any) -> Union[str, bool]:
+def _value_str(val: Any) -> Union[str, bool, int, float]:
```
jfennick, line 160: This code is unconditionally converting the input to a string, so the return type should just be str.

Perhaps more importantly, why are we creating one-line wrappers for python standard library functions?

camilovelezr: It is not

camilovelezr:
> FYI your second and third commits fix issues that should have been found by the pre-commit hooks.
> 
> From the [readme](https://github.com/PolusAI/workflow-inference-compiler?tab=readme-ov-file#quick-start): `pre-commit install # Required for developers`

Yes, I am aware of this.
As I have mentioned multiple times, mypy's pre-commit hook locally behaves differently to the Action's mypy, so this is the only way I can guarantee that all checks pass. I have to ignore the local pre-commit hook

src/wic/api/pythonapi.py
jfennick: The problem is that the docstring does not match the implementation.
>     """Return value of input as str."""
>     return str(val)
These two lines are consistent, but adding @singledispatch with @_value_str.register to the other implementations actually causes this to be the identity function on the whitelisted types, and the str() function otherwise.

So why are we doing that?

jfennick: What I'm trying to say is that the one and only call site is in the _yml function
`inp.name: _value_str(inp.value)`
where it appears that you are using the shorthand `name: type` syntax for specifying the **formal** parameters.
However, the _value_str function appears to operate on **arguments**.

https://en.wikipedia.org/wiki/Parameter_(computer_programming)

sameeul: IMO, the function name and the purpose does not align. We should either rename this function to something that can be helpful to infer its purpose. Also, from the one use of this function, it appears to me that it takes an input of some allowed type  and returns it as the input type (without any type conversion). So, `_value_str`  is misleading here.

jfennick:
> > FYI your second and third commits fix issues that should have been found by the pre-commit hooks.
> > From the [readme](https://github.com/PolusAI/workflow-inference-compiler?tab=readme-ov-file#quick-start): `pre-commit install # Required for developers`
> 
> Yes, I am aware of this. As I have mentioned multiple times, mypy's pre-commit hook locally behaves differently to the Action's mypy, so this is the only way I can guarantee that all checks pass. I have to ignore the local pre-commit hook

`which mypy`
If you are using the global system package manager version of mypy, then there could be a problem.
If you are using the version of mypy that gets installed with the wic conda environment (i.e. the same one that gets installed in the github action) then there should not be a problem.

src/wic/api/pythonapi.py
camilovelezr: That is not the purpose of the function.
Its purpose is to help in the generation of a dictionary that can be exported by the `yaml` library to a YAML file. 
You are right that a better name - and an updated docstring - would make things better.
I will rename it and update the docstring.

jfennick: Upon closer inspection, it appears that
`inp.name: _value_str(inp.value)`
is being used to generate [inline inputs](https://workflow-inference-compiler.readthedocs.io/en/latest/userguide.html#inline-inputs), i.e. arguments.

The confusion is that there is also a shorthand `name: type` notation within the inputs: tag (for both CWL Workflow class and CWL CommandLineTool class) which would be parameters.

So let's change the name to something with inline_inputs in it, and let's put the URL to the feature somewhere. Your implementation should also (eventually, not necessarily a blocker for this PR) support all primitive CWL types.

Moreover, note that the inline inputs feature is not limited to CWL primitive types. It supports the full algebra of data types (e.g. nested lists, dicts, etc). (For certain Tools it also supports custom input schemas. See the gromacs mdp options example here https://workflow-inference-compiler.readthedocs.io/en/latest/installguide.html#intellisense-code-completion)

So this code is making assumptions and restrictions that are not required for the underlying yml syntax. This will need to be fixed in future PRs.

camilovelezr: > Upon closer inspection, it appears that `inp.name: _value_str(inp.value)` is being used to generate [inline inputs](https://workflow-inference-compiler.readthedocs.io/en/latest/userguide.html#inline-inputs), i.e. arguments.
> 
> The confusion is that there is also a shorthand `name: type` notation within the inputs: tag (for both CWL Workflow class and CWL CommandLineTool class) which would be parameters.
> 
> So let's change the name to something with inline_inputs in it, and let's put the URL to the feature somewhere. Your implementation should also (eventually, not necessarily a blocker for this PR) support all primitive CWL types.
> 
> Moreover, note that the inline inputs feature is not limited to CWL primitive types. It supports the full algebra of data types (e.g. nested lists, dicts, etc). (For certain Tools it also supports custom input schemas. See the gromacs mdp options example here https://workflow-inference-compiler.readthedocs.io/en/latest/installguide.html#intellisense-code-completion)
> 
> So this code is making assumptions and restrictions that are not required for the underlying yml syntax. This will need to be fixed in future PRs.

This function only referes to things that will be serialized by `yaml`

jfennick: Using this definition for echo
```
cwlVersion: v1.2

class: CommandLineTool

baseCommand: echo

inputs:
  message:
    type: Any  # NOTE: This requires v1.2 above
    inputBinding:
      position: 1

outputs:
  stdout:
    type: File
    outputBinding:
      glob: stdout

stdout: stdout
```
Consider the following **wic** yaml file echo_types.yml:
```
steps:
- echo:
    in:
      message: Hello World
- echo:
    in:
      message: 1
- echo:
    in:
      message: 2.0
# There are many more types than just string, int, and float:
- echo:
    in:
      message: [a, list, of, strings]
- echo:
    in:
      message: [[2D, array], [of, strings]]
- echo:
    in:
      message: ['nested', {'data': ['structures']}]
- echo:
    in:
      message: ['arbitrary': {'json': {'blob': [1, 2.0, {3.14: {}}]}}]
# i.e. there is an entire algebra of data types!
```
```
wic --yaml echo_types.yml --ignore_validation_errors --run_local
...
Success! Output files should be in outdir/
```
Notice that I bolded **wic** yaml above, because the wic python API isn't supposed to just generate any random yaml, it is supposed to generate yaml according to the wic schema. Please [readthedocs](https://workflow-inference-compiler.readthedocs.io/en/latest/validation.html#the-wic-schema).

jfennick: @camilovelezr Since this PR is a hotfix for one of our internal users, we can merge it as-is.

But for future reference, my point is that the _value_str / _yaml_value function is an (incomplete!) identity function. So rather than try to extend a not-quite-identity function to cover more types, we should just delete it.

PR #181, opened 2024-02-23
use pyyaml.YAMLObject for serialization

jfennick:
As discussed in #170, this PR simply removes the _yaml_value function.

### DIFF
--- a/src/wic/api/pythonapi.py
+++ b/src/wic/api/pythonapi.py
@@ -1,7 +1,6 @@
 # pylint: disable=W1203
 """CLT utilities."""
 import logging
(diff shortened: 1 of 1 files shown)


camilovelezr (changes requested):
As discussed with @sameeul removing this function leads to issues when generating YML

jfennick:
> As discussed with @sameeul removing this function leads to issues when generating YML

Please see the latest commit. In any programming language, serialization and deserialization logic should be encapsulated and hidden away from the call site(s). For the pyyaml library, this means inheriting from pyyaml.YAMLObject.

PR #186, opened 2024-02-27
add --allow_partial_failures flag (including tests)

vjaganat90:
This includes new entry point for partial failures


src/wic/plugins.py
```diff
@@ -117,6 +127,51 @@ def get_tools_cwl(homedir: str, validate_plugins: bool = False,
...
+    """Updates outputs optional for every CWL CommandLineTool
+
+    Args:
+        rose_tree (RoseTree): The RoseTree returned from compile_workflow(...).rose_tree
+
+    Returns:
+        RoseTree: rose_tree with output optional updates to every CWL CommandLineTool
+    """
+    n_d: NodeData = rose_tree.data
+    # NOTE: Since only class: CommandLineTool should have dockerPull tags,
```
jfennick, line 163: This comment has been copy & pasted without considering whether the statement still applies for the cwl_update_outputs_optional function.

jfennick: Specifically, test_rand_fail.yml does not contain any subworkflows, so this code path is not exercised. I recommend extracting the fail step into a single-step subworkflow fail.yml

tests/test_examples.py
```diff
@@ -128,6 +128,24 @@ def test_run_inlined_workflows_on_push(yml_path_str: str, yml_path: Path, cwl_ru
     run_workflows(yml_path_str, yml_path, cwl_runner, args)
 
 
+# partial failure tests
+@pytest.mark.slow
```
jfennick, line 132: These tests should finish within seconds. They should not be marked slow.

tests/test_setup.py
```diff
@@ -42,6 +57,8 @@ def get_args(yaml_path: str = '', cwl_inline_subworkflows: bool = False) -> argp
                     for yml_namespace, yml_paths_dict in yml_paths.items()
                     for yml_path_str, yml_path in yml_paths_dict.items()]
 
+yml_paths_partial_failure = [('examples/test_rand_fail.yml', Path('examples/test_rand_fail.yml'))]
```
jfennick, line 60: We should try not to hard-code paths to specific workflow files. We should use the [auto-discovery](https://workflow-inference-compiler.readthedocs.io/en/latest/userguide.html#auto-discovery) mechanism, together with a config_ci.json file to blacklist this file from the regular CI and create a new category for workflows which are expected to fail. For now, this is okay, but we should make the change in a future PR.

tests/test_setup.py
```diff
@@ -31,6 +32,20 @@ def get_args(yaml_path: str = '', cwl_inline_subworkflows: bool = False) -> argp
     return args
 
 
+def get_appended_args(yaml_path: str = '', suppliedargs: List[str] = []) -> argparse.Namespace:
```
jfennick, line 35: We should not have both get_args and get_appended_args. We should keep the body of get_appended_args, name it get_args, and update all of the call sites where cwl_inline_subworkflows = True.

tests/test_setup.py
vjaganat90: Yes I will move this to auto-discovery in a future PR

jfennick:
LGTM

PR #188, opened 2024-02-27
use --relax-path-checks for toil

jfennick:


src/wic/run_local.py
```diff
@@ -243,9 +246,6 @@ def run_local(args: argparse.Namespace, rose_tree: RoseTree, cachedir: Optional[
                 if not cachedir:  # if running on CI
                     print(e)
 
-        proc = sub.run(cmd, check=False)
-        retval = proc.returncode
-
```
sameeul, line 248: Are we not running the final `cmd`?

jfennick: We are already running cmd in either lines 224 or 230. Somehow we are unnecessarily running it twice.

sameeul (approved):

PR #204, opened 2024-03-28
indices text file standard

misterbrandonwalker:


jfennick:
I initially wrote this as a rhetorical question (to force the reader to actually run the command):
`glob: "*.txt"  # What is the order of the output array??`
But for the documentation, I think we should explicitly answer the question by showing the actual output json blob.

PR #251, opened 2024-07-12
Fix PyPy Docker build process

sameeul:
Currently with new release of psutils (6.0.0), cwl-tools dependency of psutil (<6, >3.1) is not met and hence needed separate installation. 
Also, httptools lack PyPy wheel and so needed to be installed via conda.


vjaganat90:
LGTM, Ship it!

PR #259, opened 2024-08-10
Copy cwl adapters to the wheel

sameeul:


sameeul:
Closed in favor of #258

PR #261, opened 2024-08-19
simplify incoming wfb and correct outgoing object syntax

vjaganat90:


sameeul (approved):
Approving for now due to time-crunch. In future, dead/ commented codes need to be tracked in alternate way, rather then in commits :-)

PR #271, opened 2024-09-10
Add ICT to CLT conversion

JesseMckinzie:
- Add function to convert ICT to CLT
- Add ICT spec code

### DIFF
--- a/src/sophios/api/utils/converter.py
+++ b/src/sophios/api/utils/converter.py
@@ -1,10 +1,15 @@
+
 import copy
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Union
 import yaml
+from pathlib import Path
 from jsonschema import Draft202012Validator
 from sophios.utils_yaml import wic_loader
 
 from sophios.wic_types import Json, Cwl
+
+from sophios.api.utils.ict.ict_spec.model import ICT
+from sophios.api.utils.ict.ict_spec.cast import cast_to_ict
 
 SCHEMA: Json = {
     "$schema": "http://json-schema.org/draft-07/schema#",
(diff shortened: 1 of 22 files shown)


src/sophios/api/utils/converter.py
```diff
@@ -239,3 +244,19 @@ def wfb_to_wic(inp: Json) -> Cwl:
...
+    
+    Args:
+        ict (Union[ICT, Path, str, dict]): ICT to convert to CLT. ICT can be an ICT object,
+        a path to a yaml file, or a dictionary containing ICT
+
+    Returns:
+        dict: A dictionary containing the CLT
+    """
+    
+    ict_local = ict if isinstance(ict, ICT) else validate(ict)
```
vjaganat90, line 260: validate isn't implemented, just nuke it and raise an error here. Also see if you can eliminate some of the bloat and boiler plate.

tests/test_ict_to_clt_conversion.py
vjaganat90, line 1: This is needs to redone. Follow the pattern of existing test pattern (See test_rest_core.py and test_examples.py). Also you need to add command to run these tests in runworkflows.yml so that these become part of CI.

tests/data/czi_extract.py
vjaganat90, line 1: Better if all these data objects were .json files as these aren't code. Need more descriptive folder name than just 'data', maybe 'ict_clt_data'

vjaganat90 (changes requested):
Might need one more round of review. Adding tests to CI is necessary.

src/sophios/api/utils/ict/ict_spec/validate.py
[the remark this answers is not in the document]
JesseMckinzie: This is implemented and is needed for casting different types to the ICT object. See the singledispatch function on line 12 with corresponding functions for different argument types on lines 19, 37, and 55.

src/sophios/api/utils/converter.py
JesseMckinzie: Validate is implemented and needed for casting the different types of arguments to the ICT object

vjaganat90 (changes requested):

sameeul:
closed in favor of #275

PR #277, opened 2024-09-27
fix singularity run error on Dali

ndonyapour:


sameeul:
Closed in favor of #278

PR #282, opened 2024-10-08
ICT spec updates and WFB payload fix

JesseMckinzie:
- Update ict spec to not require the UI component
- Update pydantic class to only validate UI when present
- Add function to add missing inputs in WFB payload using the links and ICT data
- Add unit tests to for add missing WFB inputs

### DIFF
--- a/src/sophios/api/utils/converter.py
+++ b/src/sophios/api/utils/converter.py
@@ -193,3 +193,56 @@ def ict_to_clt(ict: Union[ICT, Path, str, dict], network_access: bool = False) -
     ict_local = ict if isinstance(ict, ICT) else cast_to_ict(ict)
 
     return ict_local.to_clt(network_access=network_access)
+
+
+def update_payload_missing_inputs_outputs(wfb_data: dict) -> dict:
+    """Update payload with missing inputs and outputs using links"""
+
+    wfb_data_copy = copy.deepcopy(wfb_data)
+
+    links = wfb_data_copy["state"]["links"]
+    nodes = wfb_data_copy["state"]["nodes"]
+    plugins = wfb_data_copy["plugins"]
+
+    # hashmap of node id to nodes for fast node lookup
+    nodes_dict = {node['id']: node for node in nodes}
+
+    # hashmap of plugins id to nodes for fast plugin lookup
+    plugins_dict = {plugin['pid']: plugin for plugin in plugins}
+
+    # find links corresponding to the node
+    for link in links:
+
+        # link ids
+        target_id: int = link["targetId"]
+        source_id: int = link["sourceId"]
+
+        target_node = nodes_dict[target_id]
+        source_node = nodes_dict[source_id]
+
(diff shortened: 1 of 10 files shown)


sameeul (approved):

src/sophios/api/utils/converter.py
```diff
@@ -161,3 +161,56 @@ def ict_to_clt(ict: Union[ICT, Path, str, dict], network_access: bool = False) -
     ict_local = ict if isinstance(ict, ICT) else cast_to_ict(ict)
 
     return ict_local.to_clt(network_access=network_access)
+
+
+def update_payload_missing_inputs_outputs(wfb_data: dict) -> dict:
```
vjaganat90, line 166: use more restrictive Json type instead of dict

src/sophios/api/utils/ict/ict_spec/cast.py
```diff
@@ -22,6 +23,10 @@ def cast_to_ict(ict: Union[Path, str, dict]) -> ICT:
         else:
             raise ValueError(f"File extension not supported: {ict}")
 
+        data.pop("ui", None)
```
vjaganat90, line 26: This part of the code is not tested

src/sophios/api/utils/ict/ict_spec/model.py
```diff
@@ -24,29 +24,31 @@ class ICT(Metadata):
...
+            io_dict = {"inputs": [], "outputs": []}  # type: ignore
+            ui_keys = [ui.key.root.split(".") for ui in self.ui]
+            for ui_ in ui_keys:
+                io_dict[ui_[0]].append(ui_[1])
+            input_names = [io.name for io in self.inputs]
+            output_names = [io.name for io in self.outputs]
+            inp_bool = [x in input_names for x in io_dict["inputs"]]
+            out_bool = [x in output_names for x in io_dict["outputs"]]
+
+            if not all(inp_bool):
```
vjaganat90, line 43: is this part tested with converting any of the plugins to clt?

vjaganat90 (changes requested):
We need to make sure this works with conversion to clt.

src/sophios/api/utils/converter.py
```diff
@@ -161,3 +161,56 @@ def ict_to_clt(ict: Union[ICT, Path, str, dict], network_access: bool = False) -
...
+        missing_input_key = input_directories[link["inletIndex"]]["name"]
+        missing_output_key = output_directories[link["outletIndex"]]["name"]
+
+        # add the missing input value to the node if needed
+        target_node["settings"]["inputs"][missing_input_key] = source_node["settings"]["outputs"][missing_output_key]
+
+    return wfb_data_copy
+
+
+def is_inlet(binding: dict) -> bool:
```
vjaganat90, line 208: This function definition should be inside update_payload_missing_inputs_outputs. Shouldn't be in the outer namespace of the file.

PR #283, opened 2024-10-16
add user args for workflows built with python api

vjaganat90:
add provisions for args in python api when building a workflow object


src/sophios/compiler.py
```diff
@@ -884,7 +884,7 @@ def compile_workflow_once(yaml_tree_ast: YamlTree,
                 if in_dict['value'].startswith('/'):
                     print("Warning! directory can not start with '/'")
                     print("It is most likely an incorrect path! Can't create directories!")
-                    sys.exit(1)
+                    # sys.exit(1)
```
sameeul, line 887: removing the code is better than commenting it out :-)

vjaganat90: I am removing the whole block in PR 284. It was just a testing comment.

PR #301, opened 2024-12-09
Update linking of the outputs and inputs

sameeul:


src/sophios/api/http/restapi.py
```diff
@@ -101,8 +101,8 @@ async def compile_wf(request: Request) -> Json:
     req = converter.update_payload_missing_inputs_outputs(req)
     wfb_payload = converter.raw_wfb_to_lean_wfb(req)
     # schema non-preserving
-    workflow_temp = converter.wfb_to_wic(wfb_payload)
-    wkflw_name = "generic_workflow_" + str(uuid.uuid4())
```
vjaganat90, line 105: why was removing uuuid from workflow name necessary?

vjaganat90 (changes requested):
The wfb schema needs update as certain fileds "outletIndex", "ui" and "inletIndex" are required fields now.

PR #370, opened 2026-06-18
fix: move away versioning from versioneer and simplify release prep

vjaganat90:


.github/workflows/test_and_publish_pypi.yml
```diff
@@ -3,22 +3,33 @@ name: Test and Publish PyPi
 on:
   push:
     tags:
-      - v[0-9]+.[0-9]+.[0-9]+
-      - v[0-9]+.[0-9]+.[0-9]+-dev[0-9]+
+      - "v*.*.*"
```
sameeul: **[P1] Double-trigger — guaranteed twine failure on every release**

When release-please pushes a tag, this workflow fires **twice**:
1. Via this `push.tags` trigger
2. Via `createWorkflowDispatch` from the `trigger-publish` job in `release-please.yml`

Both runs build the same wheel and call `twine upload`. The second one gets HTTP 400 "File already exists" from PyPI on every release.

**Fix:** Remove the `push.tags` trigger entirely and rely solely on the dispatch from release-please (which already carries the correct tag ref). See `release-please.yml` trigger-publish job.

.github/workflows/release-please.yml
```diff
@@ -0,0 +1,44 @@
...
+
+jobs:
+  release-please:
+    runs-on: ubuntu-latest
+    outputs:
+      release_created: ${{ steps.release.outputs.release_created }}
+      tag_name: ${{ steps.release.outputs.tag_name }}
+      version: ${{ steps.release.outputs.version }}
+    steps:
+      - uses: google-github-actions/release-please-action@v4
```
sameeul: **[P3] Unpinned action with `contents: write` on master**

`@v4` is a mutable tag — a compromised `v4.x.x` patch gets write access to master, can push commits, create tags, and open PRs. `issues: write` is also not needed by release-please.

**Fix:**
1. Pin to a commit SHA: `google-github-actions/release-please-action@<sha>`
2. Remove `issues: write` from the top-level permissions
3. Move permissions to job-level to scope them tightly

.github/workflows/release-please.yml
```diff
@@ -0,0 +1,44 @@
...
+        id: release
+        with:
+          config-file: release-please-config.json
+          manifest-file: .release-please-manifest.json
+
+  trigger-publish:
```
sameeul, line 26: Related to the double-trigger on `test_and_publish_pypi.yml`: this `trigger-publish` job dispatches a second run of the publish workflow after the tag push already triggered one. Recommend keeping this dispatch as the **sole** trigger and removing the `push.tags` block from `test_and_publish_pypi.yml`.

.github/workflows/test_and_publish_pypi.yml
```diff
@@ -3,22 +3,33 @@ name: Test and Publish PyPi
...
+  workflow_dispatch:
```
sameeul, line 4: **[P2] `workflow_dispatch` with no required `ref` input**

Anyone with Actions write permission can trigger this from the GitHub UI. Without a `ref` input, `github.ref` defaults to master. With `no-guess-dev`, setuptools-scm returns the last tag — twine then tries to upload a master-HEAD wheel with the wrong version.

```yaml
workflow_dispatch:
  inputs:
    ref:
      description: 'Tag to publish (e.g. v0.5.1)'
      required: true
```
Then use `${{ github.event.inputs.ref }}` in the checkout step's `ref:` field.

.github/workflows/test_and_publish_pypi.yml
```diff
@@ -3,22 +3,33 @@ name: Test and Publish PyPi
...
+        uses: actions/setup-python@v5
         with:
           python-version: 3.11
 
+      - name: Verify version detection
+        run: |
+          git fetch --force --tags
+          git status --short
+          python -m pip install setuptools-scm
+          python -m setuptools_scm
```
sameeul: **[P2] Verify step exits 0 even when fallback fires**

`python -m setuptools_scm` succeeds with exit code 0 when `fallback_version` kicks in (shallow clone, missing tags, etc). The step passes but the built wheel carries the fallback version.

```yaml
run: |
  python -m pip install setuptools-scm
  python -m setuptools_scm | grep -E '^[0-9]+\.[0-9]+\.[0-9]+' || (echo 'ERROR: version detection fell back to fallback_version'; exit 1)
```

.github/workflows/build_wheel.yml
```diff
@@ -10,13 +10,23 @@ jobs:
...
+        uses: actions/setup-python@v5
         with:
           python-version: 3.11
 
+      - name: Verify version detection
+        run: |
+          git fetch --force --tags
+          git status --short
+          python -m pip install setuptools-scm
+          python -m setuptools_scm
```
sameeul: **[P2] Same verify-step gap as `test_and_publish_pypi.yml`** — exits 0 even on fallback.

```yaml
run: |
  python -m pip install setuptools-scm
  python -m setuptools_scm | grep -E '^[0-9]+\.[0-9]+\.[0-9]+' || (echo 'ERROR: version detection fell back to fallback_version'; exit 1)
```

Also: this step is duplicated verbatim in both CI files — consider extracting to a composite action at `.github/actions/verify-version/action.yml`.

pyproject.toml
```diff
@@ -146,12 +142,17 @@ namespaces = true
...
-style = "pep440-pre"
-versionfile_source = "src/sophios/_version.py"
-versionfile_build = "sophios/_version.py"
-tag_prefix = "v"
+[tool.setuptools_scm]
+fallback_version = "0.5.0"
+version_scheme = "no-guess-dev"
+local_scheme = "no-local-version"
+tag_regex = "^v(?P<version>[0-9]+\\.[0-9]+\\.[0-9]+(?:[.-]dev[0-9]+)?)$"
+write_to = "src/sophios/_version.py"
```
sameeul: **[P2] `write_to` is deprecated in setuptools-scm >= 8**

This project requires `setuptools-scm>=8` but `write_to` (and `write_to_template`) were deprecated in v8 in favour of `version_file`. Generates a deprecation warning on every build.

More importantly: `_version.py` is now gitignored and **never imported** — `__init__.py` uses `importlib.metadata.version()` directly — so this config is generating a file nobody reads.

**Fix:** Remove both `write_to` and `write_to_template` entirely.

pyproject.toml
```diff
@@ -146,12 +142,17 @@ namespaces = true
...
 test = "pytest --workers 8"
 
-[tool.versioneer]
-VCS = "git"
-style = "pep440-pre"
-versionfile_source = "src/sophios/_version.py"
-versionfile_build = "sophios/_version.py"
-tag_prefix = "v"
+[tool.setuptools_scm]
+fallback_version = "0.5.0"
```
sameeul: **[P2] `fallback_version` will silently become stale**

After the next release (e.g., `v0.6.0`), any CI runner building from a shallow clone or without full tag history will still report `0.5.0`. release-please does not update `fallback_version` inside `[tool.setuptools_scm]`.

**Fix:** Change to `fallback_version = "0.0.0.dev0"` — an obviously-wrong sentinel that makes fallback builds visibly broken rather than silently misversioned.

src/sophios/__init__.py
```diff
@@ -1,6 +1,9 @@
+from importlib.metadata import PackageNotFoundError, version
 
-from . import _version
-__version__ = _version.get_versions()['version']
+try:
+    __version__ = version("sophios")
+except PackageNotFoundError:
+    __version__ = "unknown"
```
sameeul: **[P2] Silent fallback to `"unknown"` poisons all generated CWL output**

`auto_gen_header` is computed at import time (line 9), so every CWL file generated from an uninstalled checkout gets `version unknown` in its header. No warning, no log.

Common scenario: cloned repo without `pip install -e .`.

```python
except PackageNotFoundError:
    import warnings
    warnings.warn(
        "sophios is not installed — __version__ will be 'unknown'. Run: pip install -e .",
        stacklevel=2,
    )
    __version__ = "unknown"
```

Also: this path has zero test coverage. A test mocking `importlib.metadata.version` to raise `PackageNotFoundError` would prevent silent regressions.

sameeul:
**[P2] Stale `.gitattributes` entry**

`src/sophios/_version.py export-subst` is still present in `.gitattributes` but the file is now gitignored and untracked. `git archive` won't include it, so the `export-subst` instruction is a no-op. Dead entry — should be removed alongside the `.gitignore` addition.

```bash
# in .gitattributes, remove:
src/sophios/_version.py export-subst
```

sameeul:
**Review follow-up — commit 51f6932**

Checked all 9 findings against the new commit. 8/9 addressed, all well done:

✅ **P1 Double-trigger** — `push.tags` removed from `test_and_publish_pypi.yml`. Dispatch from release-please is now the sole trigger, and it correctly passes `inputs.ref`.

✅ **P2 `workflow_dispatch` required ref** — Added as required input; checkout uses `${{ github.event.inputs.ref }}`.

✅ **P2 Verify step** — Extracted to `.github/actions/verify-version/action.yml`. The implementation is better than I suggested: it detects the `0.0.0.dev0` fallback sentinel, errors by default, and when `expected-ref` is provided it validates the exact version match using `packaging.version.Version`. `build_wheel.yml` uses `allow-fallback: true` (correct), `test_and_publish_pypi.yml` uses `expected-ref` (strict for publish). 

✅ **P2 `write_to` deprecated** — Removed `write_to` and `write_to_template` entirely. The `_version.py` gitignore entry was also correctly removed since the file won't be generated anymore.

✅ **P2 `fallback_version`** — Changed to `"0.0.0.dev0"`.

✅ **P2 Silent `"unknown"` version** — Wrapped in `_resolve_version()` with `warnings.warn(RuntimeWarning, stacklevel=2)`.

✅ **P3 Unpinned action** — Pinned to SHA. `actions/github-script` in trigger-publish is also SHA-pinned. `issues: write` removed. Permissions moved to job-level.

✅ **Testing gap** — `test_version_resolution_warns_when_package_metadata_is_missing` added in `tests/test_setup.py`. Test is correct — monkeypatches `sophios.version` (the module-level name) so `_resolve_version()` picks it up.

---

**Still outstanding (minor):**

⚠️ **`.gitattributes`** — `src/sophios/_version.py export-subst` still present. The file no longer exists in the tree (no longer generated, no longer gitignored), so this entry is dead and does nothing. Low priority but worth removing for cleanliness.

LGTM otherwise — the verify-version composite action is a nice improvement over the raw grep I suggested.

PR #381, opened 2026-08-17
Spec 0: quarantine the contrib zone behind an enforced import boundary

vjaganat90:
Adds the approved core-refactor design at design_docs/core-refactor-design.md and implements its first spec: contrib may import core; core may never import contrib.
  
  - ICT, the Workflow Builder converter, and REST move to src/sophios/contrib/; tests split into tests/core/ and tests/contrib/, with the zone marker derived from directory location so the two cannot drift.
  - Core is deliberately not relocated — that would rename every path clients import and all four console entry points, which the governing constraint forbids. Only contrib import paths change; the REST HTTP surface is untouched.
  - The boundary is enforced by a property over the real module graph rather than a maintained list, so new core modules are covered automatically. Proven by deliberate breach, not by watching it pass.
  - Adds a guard that every console script and container command names an importable module — entry points are strings nothing type-checks, and this repo has now been bitten twice by a rename leaving one stale.
  - Stacks on top of fix_stale_docker branch. One pre-existing failure is knowingly left alone: test_fuzzy_compile intermittently finds a null !ii input raising a bare ValueError, which reproduces on master and belongs to Spec 1.


.github/workflows/lint_and_test.yml
```diff
@@ -224,6 +224,20 @@ jobs:
...
+      - name: PyTest Core/Contrib Boundary
+        if: always()
+        run: cd sophios/ && pytest tests/core/test_zone_boundary.py
+
+      # Contrib surfaces (ICT, Workflow Builder, REST) run in core CI, but as
+      # their own step, so a red build is attributable to a zone without
+      # bisecting. See design_docs/core-refactor-design.md, Spec 0.
+      - name: PyTest Contrib Surfaces
+        if: always()
+        run: cd sophios/ && pytest tests/contrib
```
darkclad: This step will fail CI. This job's matrix is [ubuntu-latest, windows-latest] (line 89) — GitHub-hosted runners. pytest tests/contrib collects test_rest_api.py::test_rest_core_runs_workflow (5 parametrizations), which shells out to cwltool_filterlog ... --disable-pull against Docker images. This workflow has no cwl-docker-extract pre-pull step, so the images aren't present on Linux, and windows-latest can't run Linux containers at all.

Today that test only runs from run_workflows.yml:191 on self-hosted runners that pre-pull. skip_pypi_ci is already this repo's idiom for "needs containers" — build_wheel.yml:34 excludes it for exactly this reason — so:


run: cd sophios/ && pytest tests/contrib -m "not skip_pypi_ci"
That keeps the zone-attributable step the design doc asks for without duplicating the container run that already happens on the docker-capable runners.

.github/workflows/lint_and_test.yml
```diff
@@ -224,6 +224,20 @@ jobs:
         # NOTE: This test MUST be run in serial! See is_isomorphic_with_timeout()
         timeout-minutes: 20 # backup timeout for windows
 
+      # design_docs/core-refactor-design.md, Spec 0: the core/contrib boundary.
+      # Core must never import contrib.
+      # This is a static scan of the module graph; it needs no containers.
+      - name: PyTest Core/Contrib Boundary
+        if: always()
+        run: cd sophios/ && pytest tests/core/test_zone_boundary.py
```
darkclad: test_zone_boundary.py gets a dedicated step, but test_entry_points.py doesn't — it only executes via the whole-repo python -m pytest -m "not skip_pypi_ci" in build_wheel.yml / test_and_publish_pypi.yml. The design doc's Enforcement table lists both as Spec 0 mechanisms, and the stated rationale for dedicated steps ("a red build is attributable to a zone without bisecting") applies equally. It does run on PRs, so this isn't a coverage hole — just an asymmetry that looks unintentional.


run: cd sophios/ && pytest tests/core/test_zone_boundary.py tests/core/test_entry_points.py

tests/core/test_zone_boundary.py
```diff
@@ -0,0 +1,151 @@
+"""Core/contrib zone boundary (CR-001, property P01).
+
+The core compiler must never acquire a dependency on the peripheral surfaces.
+See design_docs/core-refactor-design.md, Spec 0:
+
+    contrib may import core. core may never import contrib.
+
+This is checked over the real module graph rather than a fixed list, so a newly
```
darkclad, line 8: Slightly oversells it — this is a static AST scan, so dynamic imports are invisible. I confirmed that adding importlib.import_module("sophios.contrib.wfb_util") to a core module yields crossings: [] and affected: [], while both from sophios.contrib import x and from ...contrib import x are correctly caught.

That hole is reachable rather than theoretical: src/sophios/api/__init__.py:12 already resolves submodules through import_module. Worth a caveat in the docstring so a future reader doesn't over-trust a green check.

tests/core/test_entry_points.py
```diff
@@ -0,0 +1,87 @@
...
+@pytest.mark.fast
+@pytest.mark.parametrize('source,module,attr', _container_targets())
+def test_container_entry_points_resolve(source: str, module: str, attr: str) -> None:
+    """Every container entry point names an importable module and attribute.
+
+    A container whose CMD names a missing module cannot start, and nothing else
+    in the suite would notice.
+    """
+    try:
+        mod = importlib.import_module(module)
```
darkclad, line 86: This is a tests/core/ module importing sophios.contrib.rest.api, which means the core-only invocation advertised in the new marker help (pytest tests/ -m "not contrib", pyproject.toml:224) still imports contrib and still needs fastapi/uvicorn present. Not wrong for what this test does — validating the container CMD requires importing its target — but the zone model and the marker's promise don't quite line up.

pyproject.toml
```diff
@@ -221,6 +221,7 @@ markers = [
     "slow: (i.e. for medium speed, use 'pytest -m \"not slow\"')",
     "serial: (These tests CANNOT be run in parallel with --workers n > 1)",
     "skip_pypi_ci: tests intentionally excluded from PyPI package-build CI",
+    "contrib: exercises a peripheral surface (ICT, Workflow Builder, REST) rather than the core compiler; run core only with 'pytest -m \"not contrib\"'",
```
darkclad, line 224: The contrib marker is registered and auto-applied by the conftest hook, but nothing actually selects on it — every CI invocation in this PR selects by path (pytest tests/contrib). It's currently decorative. Fine as forward-looking infrastructure; just flagging that the marker/location agreement the conftest docstring describes is untested in CI.

conftest.py
```diff
@@ -11,3 +19,21 @@ def pytest_addoption(parser: pytest.Parser) -> None:
...
+    by path here rather than by placing the hook in a subdirectory.
+
+    Both selection styles therefore agree:
+
+        pytest tests/contrib          # by location
+        pytest -m contrib             # by marker
+        pytest tests/ -m "not contrib"
+    """
+    for item in items:
+        if Path(str(item.fspath)).resolve().is_relative_to(CONTRIB_DIR):
```
darkclad: item.fspath is the deprecated LEGACY_PATH accessor; item.path is the modern equivalent and is already a pathlib.Path, so the round-trip drops out:


if item.path.resolve().is_relative_to(CONTRIB_DIR):

tests/contrib/test_rest_api.py
```diff
@@ -9,7 +9,7 @@
 import pytest
 import yaml
 
-from sophios.api.rest import api
+from sophios.contrib.rest import api
```
darkclad, line 12: tests/contrib/test_rest_api.py:163
Stale docstring path after the move:


inp_file (str): Fixture filename under ``tests/contrib/rest_wfb_objects``.

.github/workflows/lint_and_test.yml
vjaganat90: Agreed and fixed — used `pytest tests/contrib -m "not skip_pypi_ci"` exactly as suggested. This would have broken CI on both matrix legs, and `skip_pypi_ci` is the right existing idiom for it.

.github/workflows/lint_and_test.yml
vjaganat90: Agreed and fixed. Both Spec 0 mechanisms now run in a single step: `pytest tests/core/test_zone_boundary.py tests/core/test_entry_points.py`.

tests/core/test_zone_boundary.py
vjaganat90: Agreed, and went further than a caveat. The literal half of the hole is closable, so the scan now detects `import_module("sophios.contrib.x")` when the target is a string constant — verified by reproducing your exact case, which now reports:

```
Direct boundary crossings (1) — remove these imports:
  sophios.utils -> sophios.contrib.wfb_util
```

The docstring keeps the caveat for the residual case that genuinely cannot be resolved statically — `import_module(f".{name}", __name__)`, as in `api/__init__.py` — and now states plainly that a green result means "no statically visible crossing", which is weaker than "no crossing".

tests/core/test_entry_points.py
vjaganat90: Agreed and fixed. Container entry-point parameters now carry `pytest.mark.contrib` when their target is a contrib module, so `-m "not contrib"` collects 5 of 7 and genuinely avoids importing fastapi/uvicorn. The zone definition stays single-sourced — the test imports `CONTRIB_PREFIXES` rather than restating it.

darkclad (approved):

PR #382, opened 2026-08-20
Spec 1: a specified .wic syntax layer, and one language across both front-ends

vjaganat90:
Adds sophios.lang — a typed, frozen AST and a parser that composes YAML rather than loading it, so diagnostics can name a file, line, and column instead of reporting a schema violation somewhere in a generated document; parsing is total, returning an AST or diagnostics for any input, and the five input forms (!ii, !&, !*, the new !cwl, and bare names) are a closed union. It also adds docs/wic_language_reference.md, the human-readable definition the parser is meant to agree with, whose §6 turned out to matter most: Sophios has two front-ends emitting the same language and they had silently diverged, because the Python API writes the desugared spelling ({wic_inline_input: x}) while the parser understood only the tagged one — so API-written and hand-written .wic parsed to different ASTs. 

Fixing that surfaced two further bugs in the new code: str(_scalar(node) or '') destroyed falsy values, so an edge named false or a step id: 0 silently became empty, and edge names were being type-resolved when they are identifiers. The tagged and desugared spellings are now paired dispatch tables, so adding a construct means adding one row to each and the
equivalence is visible in the source rather than only asserted in prose. wic_raw_cwl is added as the desugared key for !cwl, which had no machine-emittable form. The syntax layer is deliberately environment-independent: resolution and type checking remain separate concerns, so a well-formed document that fails to resolve without theright plugins installed is not a language error.

### DIFF
--- a/src/sophios/lang/parser.py
+++ b/src/sophios/lang/parser.py
@@ -49,16 +49,6 @@ INTERPRETED_STEP_KEYS: Final = frozenset({'scatter', 'scatterMethod', 'when', 'r
 #: `wic:` sidecar step keys have the surface form "(1, step_name)".
 _WIC_STEP_KEY: Final = re.compile(r'^\(\s*(\d+)\s*,\s*(.+?)\s*\)$')
 
-#: How YAML's resolved scalar tags map to Python values. Annotated explicitly
-#: because the converters are heterogeneous and would otherwise infer as
-#: `object`, which is not callable.
-_SCALAR_TAGS: Final[dict[str, Callable[[str], Any]]] = {
-    'tag:yaml.org,2002:null': lambda _: None,
-    'tag:yaml.org,2002:bool': lambda s: s.lower() in ('true', 'yes', 'on'),
-    'tag:yaml.org,2002:int': int,
-    'tag:yaml.org,2002:float': float,
-}
-
 
 @dataclass(frozen=True, slots=True)
 class ParseResult:
@@ -120,7 +110,7 @@ def _document(root: yaml.nodes.MappingNode, file: str, diags: Diagnostics) -> Do
     passthrough: list[tuple[str, OpaqueCwl]] = []
 
     for key_node, value_node in root.value:
-        key = _key_text(key_node)
+        key = _key_text(key_node, file, diags)
         match key:
             case 'steps':
                 steps, steps_as_mapping = _steps(value_node, file, diags)
@@ -146,7 +136,7 @@ def _steps(node: yaml.nodes.Node, file: str, diags: Diagnostics) -> tuple[tuple[
     """
     match node:
         case yaml.nodes.MappingNode():
-            return tuple(_step(_key_text(k), v, file, diags) for k, v in node.value), True
+            return tuple(_step(_key_text(k, file, diags), v, file, diags) for k, v in node.value), True
         case yaml.nodes.SequenceNode():
             return tuple(_sequence_step(item, file, diags) for item in node.value), False
         case _:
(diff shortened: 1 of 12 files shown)


src/sophios/lang/parser.py
```diff
@@ -0,0 +1,504 @@
...
+
+    return WicSidecar(steps=tuple(steps), entries=tuple(entries), span=span)
+
+
+def _step_key(text: str) -> StepKey | None:
+    """Normalise a `"(1, name)"` sidecar key, or None if it is malformed."""
+    match = _WIC_STEP_KEY.match(text)
+    if match is None:
+        return None
+    return StepKey(int(match.group(1)), match.group(2))
```
darkclad: recurses into the step's mapping, but the real shape (per [docs/advanced.md:256](vscode-webview://1ial39mhe523731qvlavj4loua6sstk0r9m0hdpi3nji78ga3sir/sophios-src/docs/advanced.md#L256) and your own reference §5) wraps the child sidecar in a wic: key:


wic:
  steps:
    (1, outer):
      wic:            # <- _sidecar never descends through this
        steps:
          (1, inner): ...
Verified output for that input:


[('(1, outer)', WicSidecar(steps=(), entries=(('wic', {'steps': {'(1, inner)': ...}}),)))]
So (1, inner) stays a raw string buried in opaque entries. StepKey's docstring ("nothing downstream should ever parse that string again") and the reference's "you should never have to parse that string yourself" are both false at depth ≥ 2. test_wic_step_keys_are_normalised only asserts the top-level key, so it doesn't catch this.

Either descend through wic: when recursing, or give the step value its own node type holding both the step's keys and its sidecar — the current annotation tuple[tuple[StepKey, WicSidecar], ...] is claiming something that isn't true.

src/sophios/lang/parser.py
```diff
@@ -0,0 +1,504 @@
...
+    Step,
+    StepKey,
+    UnresolvedName,
+    WicSidecar,
+)
+from .spans import SourceSpan
+
+#: `!cwl expr` — the per-value escape hatch. Additive: files that do not use it
+#: are unaffected, and it is the local form of the global --allow_raw_cwl flag.
+TAG_RAW_CWL: Final = '!cwl'
```
darkclad: TAG_RAW_CWL / KEY_RAW_CWL are defined in [parser.py:37-42](vscode-webview://1ial39mhe523731qvlavj4loua6sstk0r9m0hdpi3nji78ga3sir/sophios-src/src/sophios/lang/parser.py#L37-L42), not beside their siblings in utils_yaml.py, and wic_loader() was never given a constructor for them:


'a: !cwl x/y' -> ConstructorError: could not determine a constructor for the tag '!cwl'
The reference documents !cwl in the five-forms table with no caveat and offers it as a remedy in a diagnostic message — a user who follows it gets a raw PyYAML traceback from the compiler today. Contrast §7, which does explicitly say lang_version is unused. Either register the constructor or disclose it the same way §7 does. Note also that the locked design doc §5.4 enumerates Sophios-owned syntax as !&, !*, !ii, wic: — !cwl is a new construct in the very set the design doc calls exhaustive.

Moving both constants into utils_yaml.py would have made the omission self-evident.

src/sophios/lang/parser.py
```diff
@@ -0,0 +1,504 @@
...
+
+def _output_binding(node: yaml.nodes.Node, file: str, diags: Diagnostics) -> OutputBinding:
+    """Parse one `out:` entry: a bare name, or a name bound to an edge definition."""
+    span = SourceSpan.of(file, node)
+    match node:
+        case yaml.nodes.ScalarNode():
+            return OutputBinding(node.value, None, span)
+        case yaml.nodes.MappingNode() if len(node.value) == 1:
+            key_node, value_node = node.value[0]
+            edge = value_node.tag == TAG_ANCHOR
```
darkclad: anything not !&-tagged yields edge_def=None and the value vanishes with no diagnostic.


- {id: outfile}   -> OutputBinding(name='id',   edge_def=None)   # CWL's id: form misread as the name
- file: something -> OutputBinding(name='file', edge_def=None)   # 'something' gone
No corpus file uses either today, but "preserved so it can be emitted unchanged" is a stated property of the AST. Diagnose it or keep it.

src/sophios/lang/parser.py
```diff
@@ -0,0 +1,504 @@
...
+    @property
+    def ok(self) -> bool:
+        """Whether a document was produced with no errors."""
+        return self.document is not None and not self.diagnostics.has_errors
+
+
+def parse(text: str, filename: str = '<string>') -> ParseResult:
+    """Parse `.wic` source text into a `Document`.
+
+    This function is total: it never raises, for any input. Malformed YAML,
```
darkclad: parse is not total — P04's headline guarantee
The docstring says "This function is total: it never raises, for any input." Two counterexamples, both in _opaque:


parse('requirements: &a [*a]\n')   # -> RecursionError
and eager alias expansion — a 7-level "billion laughs" file takes 1.4 s and grows 9× per level, so a ~10-line file exhausts memory. st.text(max_size=400) cannot generate either shape, so test_p04_parsing_is_total_for_arbitrary_text passes vacuously on this whole class. Worth a recursion/expansion guard plus a Code for it, and a targeted test rather than relying on the fuzzer.

PR #388, opened 2026-08-24
Spec 1: resolve, surface, and override the language version

vjaganat90:
Stacks on #387 (`semrefac_6`); the two commits after its tip are this PR, and Spec 1 closes with it.

`lang_version` goes from a documented intention to a resolved fact. `lang/versions.py` owns what versions exist and how one is chosen: an explicit setting wins, per-file `wic: lang_version:` tags are exact pins, and an untagged tree gets the newest version — never silently anything lower. Conflicting pins are reported, unknown versions are reported naming the known ones, and neither is ever coerced. Exactly one version exists today, so production resolution always lands on 0.0.1; the mechanism is still written and property-tested for the general case, over fabricated version histories, because the first real version bump is precisely when nobody will want to discover the resolver was a stub that worked by coincidence. Resolution happens once at the root over every tag in the merged tree, and subworkflow compilations inherit the result as an explicit setting through a copied options dict — one compilation, one version, tree-wide, with no shared-state mutation.

Inference is invisible by construction, so visibility is a hard requirement rather than a courtesy: the resolved version is printed by the CLI on every compile, carried as `CompiledWorkflow.lang_version` on the Python API, and annotated in the emitted CWL as `sophios:lang_version` — a declared extension field, so `cwltool` still accepts the artifact, which is checked by running `cwltool`. The `--lang_version` flag and `Workflow.compile(lang_version=...)` set the version explicitly and beat any file tag. Like `edam` before it, the `sophios` namespace prefix is now reserved in emitted `$namespaces`, and the reference documents both under the same note.


src/sophios/compiler.py
```diff
@@ -369,6 +378,41 @@ def _finalize_compilation(*,
...
+    A node already on the current path contributes nothing new, so the walk
+    simply stops there; every acyclic region is still visited in full.
+    """
+    if id(node) in _path:
+        return ()
+    path = _path | {id(node)}
+
+    pins: list[str] = []
+    match node:
+        case {'wic': {'lang_version': pin}, **_rest}:
```
darkclad, line 416: **🔴 Blocking — this branch is unreachable from real YAML workflows.**

The fix is in `src/sophios/schemas/wic_schema.py:385` (not in this diff, hence the anchor here). `_wic_tag_schema` builds:

```python
schema_props = {'graphviz': graphviz_schema, 'steps': steps, 'implementation': implementation,
                'default_implementation': default_implementation,
                'version': str_nonempty, 'driver': driver,
                'namespace': namespace, 'inlineable': inlineable}
```

No `lang_version`, and the schema uses `additionalProperties: False`. So the `wic: {lang_version: ...}` pin this PR documents is rejected during wic schema validation before the compiler ever reaches this `case`:

```
ValidationError: Additional properties are not allowed ('lang_version' was unexpected)
```

`read_ast_from_disk` turns that into `Failed to compile <file>`. The `--lang_version` CLI flag and `Workflow.compile(lang_version=...)` are unaffected — only the YAML tag path is dead.

It slipped through because `tests/core/test_lang_version.py` calls `compiler.compile_workflow` directly and never goes through the validator.

Fix is one entry:

```python
'version': str_nonempty, 'lang_version': str_nonempty, 'driver': driver,
```

Please also add a test that carries the pin through the validation path, so this gap can't reopen.

src/sophios/api/python/_workflow_runtime.py
```diff
@@ -446,6 +447,7 @@ def compile_workflow(
     *,
     write_to_disk: bool = False,
     tool_registry: Tools | None = None,
+    lang_version: str | None = None,
```
darkclad, line 450: **🟡 Minor** — the `Args:` block just below documents `workflow`, `write_to_disk`, and `tool_registry`, but not `lang_version`. The two sibling functions (`compiled_workflow` at line 544, `Workflow.compile` in `workflow.py:919`) did get updated, so this one looks like an oversight.

src/sophios/api/python/_workflow_runtime.py
```diff
@@ -522,26 +526,31 @@ def compiled_workflow_from_compiler_info(
         name=workflow.process_name,
         cwl_workflow=cwl_workflow,
         cwl_job_inputs=normalize_rose_tree_job_inputs(rose_tree, sub_node_data.workflow_inputs_file),
+        lang_version=str(cwl_workflow.get(versions.ANNOTATION_KEY, versions.LANG_VERSION)),
```
darkclad: **🟡 Minor** — when the annotation is missing, this reports `versions.LANG_VERSION` — the newest known version — as if it were the document's own.

Unreachable today (the compiler always writes the annotation), but it cuts against the no-silent-selection rule the `versions` module is built around: everywhere else an unspecified version is either resolved by an explicit rule or refused.

Consider `''` — matching `CompiledWorkflow.lang_version`'s default — or raising, so an unannotated document reads as unannotated rather than quietly claiming the newest.

darkclad (commented):
Inline notes below. One blocking issue (the YAML pin is unreachable), two minor.

Note: the blocking finding's actual fix lives in `src/sophios/schemas/wic_schema.py`, which isn't part of this diff, so it's anchored here on the line that consumes the pin.

darkclad (approved):

PR #395, opened 2026-09-01
Spec 1: configuration is a value, parsed once at the boundary

vjaganat90:
`--lang_version` never reached the compiler. Neither did `--allow_raw_cwl`, `--inference_disable`, `--partial_failure_enable`, `--insert_steps_automatically`, or any graph setting. `main` parsed the user's arguments and then asked for a fresh set of defaults, so every flag arrived at the compiler as whatever the parser said it should be when nobody typed anything.

The first fix threaded the parsed `argparse.Namespace` one function deeper. That made the symptom go away and the design worse, and the maintainer rejected it: arguments do not belong in library signatures. This is the fix at the level the defect actually lives.

**The defect.** Configuration was *derived from a command line* rather than *represented as a value*:

```python
with patch.object(sys, 'argv', ['sophios', '--yaml', yaml_path] + supplied):
    args = parser.parse_args()
```

`unittest.mock` on the production path, mutating global process state to compute a value, in a library other people embed. `parse_args` has always accepted an explicit list, so none of it was ever necessary.

The mechanism was the smaller half. The consequence was that **there was no way to name the default configuration** — the only route to it was to fabricate a command line. So four production sites fabricated one, and the pattern it forced was "fake the defaults, then patch back the one option I care about", which is literally how the Python API passes `lang_version` today. A caller holding the user's real arguments could forget to pass them and get defaults instead, silently. `main` was that caller. `contrib/rest/api.py` still is: it parses `args` for the workflow name, discards them, and compiles with `yaml_tag_paths['yaml']` set to the empty string.

**The fix** applies the rule the parser layer already follows — parse at the boundary, pass values inward:

- `get_args` uses `parser.parse_args([...])`. No argv patching, no `unittest.mock` in production.
- `default_compilation_settings()` gives library callers defaults *by name*. Asking for defaults is now something you can say rather than something you fake. The defaults themselves still come from one place, the parser's own `default=` values.
- `get_dicts_for_compilation(args)` is the CLI adapter and **requires** its argument. The old default was a fresh parse of a synthesised argv, which is precisely what let `main` hold the user's flags and compile without them. Forgetting is now a `TypeError` and a mypy failure rather than silence — the guarantee moved out of a test and into the signature.
- **No library signature grows an `args` parameter.** `_build_and_compile_workflow` takes settings and is *narrower* than before: its only use of the namespace was `graph_dark_theme`, which already travels inside `graph_settings`.

**Two scans keep it that way**, each verified to fail when its half is reverted: nothing may install a value over `sys.argv`, and an `argparse.Namespace` may not escape the CLI's own modules. The second is scoped to modules rather than functions on purpose — `main` legitimately reads paths and flags out of what it parsed, and packing those into a five-parameter signature would be less readable, not more. The one helper there that had no business taking one is pinned by name, because that was the actual mistake.

**Behaviour is unchanged for anyone who wanted defaults.** I verified the old and new paths produce identical `compiler_options`, `graph_settings` and `yaml_tag_paths`, so the Python API — and image-workflows and mm-workflows, which compile through it — are untouched, down to not editing a line of their own code. What changes is that flags typed on the command line now do what they say, and REST gets the workflow name it already had.

**Stacking:** branches off `master`; #387 and #388 are rebased on top, so this merges first.


darkclad (commented):
The design is right and the diagnosis is right — `parse_args` always took an explicit list, and the reason the mock survived is exactly the one you name: there was no way to *say* "defaults". Requiring the argument moves the guarantee into the signature, which is the durable version of the fix.

I ran both AST scans standalone over all 66 modules under `src/sophios` — clean, and the matchers see both spellings they claim to. The "deliberately permitted" list checks out: `_tool_builder_support.py:293` constructs a `Namespace` and `cli_watcher` returns one, so neither is a parameter. And the two settings paths are bit-identical for every key except `yaml_tag_paths['yaml']`, so the "unchanged for anyone who wanted defaults" claim holds and the Python API is genuinely untouched.

Nothing blocking. One worth fixing before merge, one worth a decision, three where the code is right but a comment claims more than it does.

Merge order confirmed and stricter than the description says: the heads are strictly linear, `master → 395 → 387 → 388` (`git merge-base --is-ancestor` holds for all three pairs). This also closes the open `--lang_version` finding on #388 — on the rebased head `cli.py:209` carries it into `compiler_options` and `main.py:240` calls `get_dicts_for_compilation(args)`.

tests/core/test_examples.py
```diff
@@ -210,7 +210,7 @@ def run_workflows(
         step_id = StepId(yaml_tree.step_id.stem + '_inline', yaml_tree.step_id.plugin_ns)
         yaml_tree = YamlTree(step_id, yaml_tree.yml)
 
-    compiler_options, graph_settings, yaml_tag_paths = sophios.cli.get_dicts_for_compilation()
+    compiler_options, graph_settings, yaml_tag_paths = sophios.cli.default_compilation_settings()
```
darkclad: **🟡 This is `main`'s bug, surviving in the file that is supposed to catch it.**

`run_workflows` is handed the user's parsed arguments and reads five things out of them — `args.homedir` (192), `args.cwl_inline_subworkflows` (199), `args.yaml` (195), `args.inputs_file` (232), `args.partial_failure_enable` (238) — and then compiles with defaults.

Line 140 supplies `--partial_failure_enable`, but `compiler_options['partial_failure_enable']` is `False` regardless, so the `when` guards at `src/sophios/compiler.py:1026` are never emitted. What still separates `test_run_partial_failures_pass` from `test_run_partial_failures_fail_without_flag` is only the post-compile `successCodes` patch at line 238 — so the test exercises half of partial failure and is named for the whole thing.

`yaml_tag_paths['yaml']` is `''` here too, where `main` supplies the real path, so the repo's principal end-to-end compile test compiles differently from `sophios --yaml` in the one place that reads it (`input_output.py:268`, for `cwl_subinterpreter` steps).

```python
compiler_options, graph_settings, yaml_tag_paths = sophios.cli.get_dicts_for_compilation(args)
```

`args` is already in scope. The three sibling sites (`:293`, `:392`, `test_fuzzy_compile.py:60`) hold defaults-plus-path and so only gain the correct `yaml`, but they should follow for the same reason.

Flagging it at this weight because the PR converted this exact line from `get_dicts_for_compilation()` to `default_compilation_settings()` — which makes the discard explicit and blessed at the one call site where it is wrong.

src/sophios/cli.py
```diff
@@ -132,17 +131,26 @@
                     If set to 'no' (default) passthrough flags won't be sent to the cwl_runner backend.''')
 
 
+def _argv(yaml_path: str = '', suppliedargs: list[str] | None = None) -> list[str]:
+    """The argument list a synthesised parse should read.
+
+    Note there is no program name: `parse_args` takes the arguments *after*
+    argv[0]. Passing the list explicitly is the whole point — this module used
+    to build a full argv and install it over `sys.argv` with
```
darkclad, line 146: **🟡 The module still derives configuration from the host process's `sys.argv` — twelve lines above this docstring.**

```python
parser.add_argument('--yaml', type=str,
                    required=(not ('--generate_config' in sys.argv or '--generate_schemas' in sys.argv)),
```

That runs at import time, against whatever argv the *embedding* program was invoked with. Import sophios from a program that happens to carry `--generate_config` and sophios's `--yaml` silently stops being required. It's the same defect this docstring names — a value computed from global process state, in a library other people embed.

The scan doesn't catch it: `_patches_argv` looks for *writes* to `sys.argv`, and this is a read.

I'd rather that were said out loud than inferred from a green test. Either fix it — drop `required=` and check for a missing `--yaml` in `_main`, where the two generate modes are already known — or note the exemption in the scan's docstring, so its clean bill isn't read as broader than it is.

src/sophios/contrib/rest/api.py
```diff
@@ -84,7 +84,10 @@ async def compile_wf(request: Request) -> Json:
     graph = get_graph_reps(wkflw_name)
     yaml_tree: YamlTree = YamlTree(StepId(wkflw_name, plugin_ns), workflow_can)
 
-    compiler_options, graph_settings, yaml_tag_paths = get_dicts_for_compilation()
+    # From the arguments this endpoint actually built, not a fresh default
+    # parse: the two disagreed, so `yaml_tag_paths['yaml']` was the empty
+    # string instead of the workflow being compiled.
+    compiler_options, graph_settings, yaml_tag_paths = get_dicts_for_compilation(args)
```
darkclad, line 92: **🟢 The stated consequence isn't observable.**

`wkflw_name` is `"workflow_"` — a name, not a path. The only consumer of `yaml_tag_paths['yaml']` is `input_output.py:268`:

```python
in_dict_in['root_workflow_yml_path'] = {'wic_inline_input': str(Path(yaml_tag_paths['yaml']).parent.absolute())}
```

and `Path('').parent.absolute()` and `Path('workflow_').parent.absolute()` are both the cwd. Verified.

The change is right on principle — pass what you parsed rather than re-deriving it — and I'd keep it. But this PR's argument rests on its claims being checkable, so the comment should say that, not that a wrong path was being compiled.

src/sophios/cwl_subinterpreter.py
```diff
@@ -95,9 +95,12 @@ def rerun_cwltool(homedir: str, _directory_realtime: Path, cachedir_path: Path,
         # Measure compile time
         time_initial = time.time()
 
-        # Setup dummy args
+        # This entry point has its own argument parser (`cli_watcher`), which
+        # shares no flags with the compiler's, so there is nothing of the
+        # user's to pass on: defaults are the honest answer here, and now the
+        # code says that rather than synthesising a command line to reach them.
         args = cli.get_args()
```
darkclad: **🟢 Two small things.**

"shares no flags with the compiler's" — `cli_watcher` declares `--homedir`, which the compiler's parser also declares, plus `--cachedir_path` and `--root_workflow_yml_path`. Between them that's all three keys of `YamlTagPaths`, so the values *are* in hand. The honest reason defaults are fine here is narrower: they'd only be read by a nested `cwl_subinterpreter` step, which doesn't occur.

And `args = cli.get_args()` now survives solely to reach `args.inputs_file` on line 118. Fishing one default out of a synthesised parse is the pattern this PR is retiring — `''` inline, or a named constant, would say what is meant and drop the last synthesised parse from this module.

tests/core/test_settings_boundary.py
```diff
@@ -0,0 +1,189 @@
...
+SRC: Final = REPO_ROOT / 'src' / 'sophios'
+
+#: The one module that may hold an `argparse.Namespace`: it makes them.
+PARSER_MODULE: Final = SRC / 'cli.py'
+
+#: The CLI entry point, which parses and immediately converts.
+CLI_ADAPTER: Final = SRC / 'main.py'
+
+
+def _python_files() -> list[Path]:
```
darkclad, line 45: **🟢 The scans can still go vacuous, through the file list rather than the matcher.**

`test_the_scan_can_actually_fail` proves the matcher works. Nothing proves it was pointed at anything. If `_python_files()` returns `[]` — tests run against an installed package with no `src/`, or a layout change — both parametrised scans become empty parameter sets, which pytest **skips** by default, and the two headline claims of this file pass without examining a line.

`test_zone_boundary.py` guards exactly this, twice: inline (`assert core, 'no core modules discovered; the zone scan is broken'`) and as its own test (`test_zone_scan_finds_both_zones`). One line matches the precedent:

```python
assert _python_files(), 'no package modules discovered; the scan is broken'
```

src/sophios/main.py
```diff
@@ -17,7 +17,7 @@
 from . import post_compile as pc
 from . import ast, cli, compiler, inference, inlineing, plugins, run_local, utils  # , utils_graphs
 from .schemas import wic_schema
-from .wic_types import GraphData, GraphReps, Json, RoseTree, StepId, Tools, Yaml, YamlTree
+from .wic_types import CompilerOptions, GraphSettings, YamlTagPaths, GraphData, GraphReps, Json, RoseTree, StepId, Tools, Yaml, YamlTree
```
darkclad: ⚪ Nit — 136 characters against `max-line-length=120`. Pre-existing violations elsewhere mean pylint's 9.25 floor absorbs it and autopep8 can't split it, so CI wouldn't have caught it either way — but `lint_and_test` didn't run on this head (only `branch_dispatch` and `package`), so nothing checked. Trivial wrap.

darkclad (commented):
All six addressed — checked against the new head rather than the commit message.

The `--yaml` move is placed correctly: the check in `_main` sits ahead of the `generate_config` exit and ahead of `generate_schemas`, so both generate modes still work without `--yaml` and a bare `sophios` still exits 2 with the identical message. I re-ran a *read* scan (not just the PR's write scan) over all 66 modules: the only surviving `sys.argv` references are a docstring, a debug print inside `cwl_subinterpreter.main()`, and `run_local.py` rebuilding the command — all call-time, all inside entry points. Both AST scans still clean.

The new CLI-driven property is a real upgrade on what I asked for, and deriving the parameter list from the three `TypedDict`s is worth more than the assertions it replaced — the failure mode is someone adding a flag and nobody noticing it isn't delivered, and a hand-maintained list has exactly that hole. I checked the three things that could have made the derivation quietly wrong, and none do: every bool setting really is `store_true`/`default=False` (a single `store_false` would have inverted `_non_default`), `wic_types.py` has no `from __future__ import annotations` so `__annotations__` yields real types rather than strings, and `cwl_adapters/touch.cwl` exists so the probe reaches the compiler. `_Delivered(BaseException)` does escape both the `except SophiosError` and `except Exception` arms.

One thing the rewrite dropped, inline below. Nothing blocking.

Two notes, neither worth a re-push on its own:

- `@pytest.mark.fast` is now a stretch — `test_every_setting_reaches_the_compiler` boots the full CLI eleven times (config load, `get_tools_cwl`, `get_validator`). CI went 3m02s → 3m16s, so the cost is small and worth paying, but these are integration tests wearing the unit-test marker.
- `lint_and_test` still hasn't run on either round — only `branch_dispatch` and `package`. I checked line lengths by hand and they're clean (`main.py:20` fixed, nothing over 120 in any changed file), but that's no substitute: the `Callable[..., object]` alias added to get a runtime `TypeError` past the type checker is exactly what mypy should be confirming.

Merge order unchanged and still strictly linear after all three force-pushes: `master → 395 (1133875) → 387 (375bdbc) → 388 (41cfe6a)`. Worth noting #388's rebase now inherits `test_every_setting_reaches_the_compiler`, which will cover `--lang_version` automatically the moment the field lands in `CompilerOptions`.

tests/core/test_settings_boundary.py
```diff
@@ -0,0 +1,298 @@
...
+@pytest.mark.fast
+def test_the_namespace_scan_can_actually_fail() -> None:
+    """The signature scan sees both spellings it claims to."""
+    assert _namespace_parameters(ast.parse('def f(args: argparse.Namespace) -> None: ...'))
+    assert _namespace_parameters(ast.parse('def f(*, args: Namespace | None = None) -> None: ...'))
+    assert not _namespace_parameters(ast.parse('def f(options: CompilerOptions) -> None: ...'))
+
+
+@pytest.mark.fast
+def test_the_converter_requires_arguments() -> None:
```
darkclad, line 186: **🟢 The rewrite deleted the only test of `default_compilation_settings()`.**

`test_defaults_are_available_without_a_command_line` used to sit here, and nothing replaced it:

```python
options, graph, tags = sophios.cli.default_compilation_settings()
assert options['allow_raw_cwl'] is False
assert graph['graph_dark_theme'] is False
assert tags['yaml'] == ''
```

Nothing now asserts that function returns the documented defaults — it's exercised only incidentally, via `tests/core/provocations.py:52`. That's the function whose entire reason for existing is that asking for defaults should be something you can *say*, and it's what the Python API, `wic_schema` and the subinterpreter all call — which is to say the path image-workflows and mm-workflows compile through.

The new CLI-driven property covers the other half of the split, and covers it better than what it replaced. It doesn't cover this half at all: `default_compilation_settings()` never appears in it. The deleted three lines still apply verbatim and cost nothing to restore.

darkclad (approved):

PR #397, opened 2026-09-01
style: bring every line within the configured width, and enforce it

vjaganat90:
`test_leak_boundary.py:51` was 126 characters: my wording fix in #387 replaced a sentence without re-wrapping, so it ran into the following clause.

Eight more violations predated it. Six wrap. The other three were bare links, and deleting them costs no understanding:

- `run_local.py` — a permalink into cwltool at a pinned SHA with no statement of what is there. The three lines under it name the actual constraint (`--leave-outputs` to disable `--outdir`, which has bugs) and carry their own issue link.
- `wic_schema.py` — a StackOverflow explainer of `$dynamicRef`/`$dynamicAnchor`, directly beneath the json-schema.org spec link for the same feature and a NOTE stating why the schema must be recursive.
- `wic_types.py` — a StackOverflow survey of struct-like types, beside the sentence that states the decision ("NamedTuple, to emphasize immutability") and the mypy docs link that supports it.

In each case the reason lives in the surrounding lines; the link was a second, weaker source for something already said.

The width is now a test in `test_zone_boundary.py`, with **no exemption** — not even for URLs, since all three that would have needed one were worth deleting. If a permalink ever genuinely needs the room, that is a decision to make in the test rather than a hole left open in advance.

Worth knowing why nothing caught this: autopep8 does not reflow comments, and pylint gates on a score that one long line never moves.

### DIFF
--- a/tests/core/test_zone_boundary.py
+++ b/tests/core/test_zone_boundary.py
@@ -23,11 +23,15 @@ result therefore means "no statically visible crossing", which is weaker than
 instrumentation; that is not what this test does.
 """
 import ast
+import subprocess
 from pathlib import Path
+from typing import Final
 
 import pytest
 
-SRC_ROOT = Path(__file__).resolve().parents[2] / 'src'
+REPO_ROOT: Final = Path(__file__).resolve().parents[2]
+
+SRC_ROOT = REPO_ROOT / 'src'
 PACKAGE = 'sophios'
 
 # The peripheral zone is a single subtree. Everything else under src/sophios
@@ -181,3 +185,59 @@ def test_contrib_may_import_core() -> None:
     assert any(
         any(not _is_contrib(t) for t in _reachable(m, graph)) for m in contrib
     ), 'expected contrib to depend on core; the scan may be resolving nothing'
+
+
+def _repo_python_files() -> list[Path]:
+    """Every tracked Python file. Via git, so build output is excluded.
+
+    Returns nothing outside a git checkout — an sdist ships `tests/` but no
+    `.git` — and the guard below turns that into a skip rather than a silent
+    pass. `-z` because git quotes unusual filenames otherwise, and an explicit
+    encoding because git emits UTF-8 while `text=True` decodes with the
+    locale's, which is cp1252 on Windows.
+    """
+    listing = subprocess.run(['git', 'ls-files', '-z', '*.py'], cwd=REPO_ROOT,
+                             capture_output=True, text=True, encoding='utf-8',
+                             check=False)
+    if listing.returncode != 0:
+        return []
+    return [REPO_ROOT / name for name in listing.stdout.split('\0') if name]
+
+# --------------------------------------------------------------------------
+# Lines stay within the configured width
+# --------------------------------------------------------------------------
+
+
+#: Matches pyproject's max-line-length.
+MAX_LINE_LENGTH: Final = 120
+
+
+def _over_long_lines(path: Path) -> list[tuple[int, int]]:
+    """Line numbers and lengths that exceed the limit. No exemptions."""
+    return [(number, len(line))
+            for number, line in enumerate(path.read_text(encoding='utf-8').splitlines(), start=1)
+            if len(line) > MAX_LINE_LENGTH]
+
+
+@pytest.mark.fast
+def test_the_width_scan_sees_the_repo() -> None:
+    """Zero parametrized cases is a green test that enforces nothing."""
+    if not (REPO_ROOT / '.git').exists():
+        pytest.skip('not a git checkout; nothing to enumerate')
+    assert _repo_python_files(), 'git ls-files resolved nothing; the width scan is vacuous'
+
+
+@pytest.mark.fast
+@pytest.mark.parametrize('path', _repo_python_files(),
+                         ids=lambda p: str(p.relative_to(REPO_ROOT)))
+def test_no_line_exceeds_the_configured_width(path: Path) -> None:
+    """No Python file carries a line longer than pyproject allows.
+
+    autopep8 does not reflow comments and pylint gates on a score, so nothing
+    else enforces the setting.
+    """
+    offenders = _over_long_lines(path)
+    assert not offenders, (
+        f'{path.relative_to(REPO_ROOT)} has lines over {MAX_LINE_LENGTH} columns: '
+        f'{[f"line {n} ({length})" for n, length in offenders]}'
+    )
--- a/conftest.py
+++ b/conftest.py
@@ -12,7 +12,8 @@ CONTRIB_DIR = Path(__file__).resolve().parent / 'tests' / 'contrib'
 
 
 def pytest_addoption(parser: pytest.Parser) -> None:
-    parser.addoption('--cwl_runner', type=str, required=False, default='cwltool', choices=['cwltool', 'toil-cwl-runner'],
+    parser.addoption('--cwl_runner', type=str, required=False, default='cwltool',
+                     choices=['cwltool', 'toil-cwl-runner'],
                      help='The CWL runner to use for running workflows locally.')
 
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -254,9 +254,6 @@ fail-under=9.25
 
 [tool.pylint.format]
 
-# Regexp for a line that is allowed to be longer than the limit.
-ignore-long-lines='^\s*(# )?(See )?(and )?<?https?://\S+>?$'
-
 # Maximum number of characters on a single line.
 max-line-length=120
 
--- a/src/sophios/cli.py
+++ b/src/sophios/cli.py
@@ -133,7 +133,8 @@ parser.add_argument('--custom_net', type=str, required=False,
                     help='Passes --custom-net flag to cwltool.')
 parser.add_argument('--passthrough_flags', type=str, default='no', required=False,
                     help='''Indicates that the user is passing flags to the cwl_runner backend.
-                    No checks are done on the flags or values passed. User must verify that they are sending correct flags.
+                    No checks are done on the flags or values passed.
+                    User must verify that they are sending correct flags.
                     only two valid values of this flag 'yes' or 'no'.
                     If set to 'no' (default) passthrough flags won't be sent to the cwl_runner backend.''')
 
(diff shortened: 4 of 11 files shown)


tests/core/test_zone_boundary.py
```diff
@@ -181,3 +185,41 @@ def test_contrib_may_import_core() -> None:
     assert any(
         any(not _is_contrib(t) for t in _reachable(m, graph)) for m in contrib
     ), 'expected contrib to depend on core; the scan may be resolving nothing'
+
+
+def _repo_python_files() -> list[Path]:
+    """Every tracked Python file. Via git, so build output is excluded."""
+    listing = subprocess.run(['git', 'ls-files', '*.py'], cwd=REPO_ROOT,
+                             capture_output=True, text=True, check=True)
+    return [REPO_ROOT / name for name in listing.stdout.split()]
```
darkclad: **🟠 `check=True` here takes down three unrelated tests in a non-git tree, and `.split()` mangles some filenames.**

This runs at *collection* time (it feeds `@pytest.mark.parametrize` on line 213), so both failure modes are import-time failures for the whole module, not just for the width test.

**1. `check=True` in a tree without `.git`.** `git ls-files` exits 128 outside a repository -- I verified this. The sdist ships `tests/` but not `.git`, so `pytest --pyargs` against an unpacked sdist raises `CalledProcessError` during collection and takes `test_core_does_not_import_contrib`, `test_no_contrib_reachable_from_core` and `test_contrib_may_import_core` down with it. Those three are pure AST walks over `src/` and need nothing from git; they should not be collateral.

**2. `listing.stdout.split()` splits on whitespace and ignores git's C-quoting.** Two cases I reproduced:

- `my mod.py` comes back as `'my'` and `'mod.py'` -- two paths, neither real.
- `naive_ü.py` comes back as `"naive_\303\274.py"`, quotes and octal escapes included, under the default `core.quotePath=true`.

Either way `read_text` raises `FileNotFoundError` and the real file is never scanned, so enforcement silently stops covering the file it was pointed at. No such names exist in the tree today, which is why this would land later as a confusing collection error rather than now as a caught bug.

`-z` fixes both: NUL-delimited output is never quoted and never ambiguous.

```suggestion
def _repo_python_files() -> list[Path]:
    """Every tracked Python file. Via git, so build output is excluded.

    Returns nothing outside a git checkout (an sdist ships tests/ but no
    .git); the guard below turns that into a skip rather than a silent pass.
    """
    listing = subprocess.run(['git', 'ls-files', '-z', '*.py'], cwd=REPO_ROOT,
                             capture_output=True, text=True, check=False)
    if listing.returncode != 0:
        return []
    return [REPO_ROOT / name for name in listing.stdout.split('\0') if name]
```

I ran that against this checkout: same 115 files as the current version.

tests/core/test_zone_boundary.py
```diff
@@ -181,3 +185,41 @@ def test_contrib_may_import_core() -> None:
...
+
+def _over_long_lines(path: Path) -> list[tuple[int, int]]:
+    """Line numbers and lengths that exceed the limit. No exemptions."""
+    return [(number, len(line))
+            for number, line in enumerate(path.read_text(encoding='utf-8').splitlines(), start=1)
+            if len(line) > MAX_LINE_LENGTH]
+
+
+@pytest.mark.fast
+@pytest.mark.parametrize('path', _repo_python_files(),
```
darkclad, line 231: **🟡 An empty listing makes this pass while enforcing nothing.**

`parametrize` over an empty sequence produces zero cases, and zero cases is a green test. Not hypothetical: I checked, and `git ls-files` exits **0 with empty stdout** when `cwd` is an untracked or ignored subdirectory of some *other* repository, so a checkout nested inside an unrelated worktree reports success and scans nothing. With the `check=False` suggested above it becomes the behavior in a non-git tree too, so this needs a guard either way.

The sibling test at line 187 already defends against exactly this failure mode (`'expected contrib to depend on core; the scan may be resolving nothing'`). Worth the same treatment:

```python
@pytest.mark.fast
def test_the_width_scan_sees_the_repo() -> None:
    """Guard: zero parametrized cases is a green test that checks nothing."""
    if not (REPO_ROOT / '.git').exists():
        pytest.skip('not a git checkout; nothing to enumerate')
    assert _repo_python_files(), 'git ls-files resolved nothing; the width scan is vacuous'
```

Skip when there is genuinely no repo, fail when there is one and the enumeration came back empty.

tests/core/test_zone_boundary.py
```diff
@@ -23,10 +23,14 @@
 instrumentation; that is not what this test does.
 """
 import ast
+import subprocess
 from pathlib import Path
+from typing import Final
 
 import pytest
 
+REPO_ROOT: Final = Path(__file__).resolve().parents[2]
```
darkclad, line 32: **🔵 Nit:** `SRC_ROOT` on line 34 re-walks the same `parents[2]` this line just computed. Now that `REPO_ROOT` exists, one source of truth reads better:

```python
SRC_ROOT = REPO_ROOT / 'src'
```

Behavior is identical; it just means a future change to how the root is located only has to happen in one place.

darkclad (commented):
The mechanical half of this is clean. I checked the parts that could silently change behavior:

- Both rewrapped `compile_workflow(...)` calls in `tests/core/test_examples.py` keep argument order and arity against the 14-parameter signature at `src/sophios/compiler.py:29`.
- The `--passthrough_flags` help-string split (`src/sophios/cli.py:135`) is behavior-neutral: the parser uses the default `HelpFormatter`, which collapses whitespace.
- Moving `# pylint: disable=too-many-return-statements` off the `def` line into the body of `input_lines` (`tests/core/test_lang_parser.py:105`) still suppresses R0911. I confirmed that empirically with pylint 4 / astroid 4 on a minimal repro using the same docstring-then-disable placement (10.00/10, message not emitted) rather than trusting the docs.
- The new test passes today (115 tracked `.py` files, 0 violations) and it does run in CI: `.github/workflows/lint_and_test.yml:232` invokes `tests/core/test_zone_boundary.py` directly.

The new enforcement test is the right idea, and the premise in its docstring holds: autopep8 does not reflow comments and pylint gates on a score, so nothing else actually holds the line. Three issues in `_repo_python_files()` and one config leftover, the first three inline below.

The config one has no line in this diff to hang off, so it goes here:

**`pyproject.toml:258` -- `ignore-long-lines` contradicts the new test.**

```toml
ignore-long-lines='^\s*(# )?(See )?(and )?<?https?://\S+>?$'
```

`_over_long_lines`'s docstring says "No exemptions", but a bare-URL comment line is still whitelisted by pylint. So a contributor adding a long permalink comment passes `pylint` and `autopep8` locally, then fails a test whose message cites "the configured width" -- pointing them at a setting that says the line is fine. Nothing in the tree needs the exemption right now (0 violations either way), so given this PR's stated position, deleting that key here is what makes the two agree.

tests/core/test_zone_boundary.py
```diff
@@ -181,3 +185,56 @@ def test_contrib_may_import_core() -> None:
...
+
+def _repo_python_files() -> list[Path]:
+    """Every tracked Python file. Via git, so build output is excluded.
+
+    Returns nothing outside a git checkout — an sdist ships `tests/` but no
+    `.git` — and the guard below turns that into a skip rather than a silent
+    pass. `-z` because git quotes and escapes unusual filenames otherwise.
+    """
+    listing = subprocess.run(['git', 'ls-files', '-z', '*.py'], cwd=REPO_ROOT,
+                             capture_output=True, text=True, check=False)
```
darkclad: **🟡 `-z` fixed the delimiter but not the decode — non-ASCII filenames still escape the scan on Windows.**

`text=True` decodes git's stdout with the **locale** encoding, which is `cp1252` on a stock Windows box, not UTF-8. Git emits path bytes as UTF-8, so the round-trip is lossy for anything outside ASCII.

Reproduced in a scratch repo containing `naive_ü.py` (bytes `\xc3\xbc` on disk), running this exact helper:

```
"my mod.py"        exists= True
"naive_\u00c3\u00bc.py"  exists= False     <- decoded via cp1252
"test_zone_boundary.py"  exists= True
```

`read_text` then raises `FileNotFoundError` and the file is never scanned — the same silent-coverage-gap failure mode `-z` was added to close, just narrower. And it is not Linux-only theory: the matrix at `.github/workflows/lint_and_test.yml:89` is `[ubuntu-latest, windows-latest]` and line 232 runs this file on both.

Pinning the encoding closes it:

```suggestion
    listing = subprocess.run(['git', 'ls-files', '-z', '*.py'], cwd=REPO_ROOT,
                             capture_output=True, text=True, encoding='utf-8',
                             check=False)
```

Verified against the same scratch repo: all three paths resolve. Latent today — every tracked filename is ASCII — same as the case before it.

darkclad (commented):
Re-reviewed at `1d46ef1`. Four of the five points from the last pass are addressed, and I re-ran rather than reading the diff:

- The full file at this head: **119 passed** — 115 width cases, the three architecture tests, and the new guard.
- Copied `src/` plus the test into a directory with no `.git`: **3 passed, 2 skipped**. The architecture tests survive now and the two width tests skip with honest reasons. The same tree at `f736c14` dies with `CalledProcessError` during collection and takes all three down, so the fix lands on the case it was aimed at.
- Scratch repo containing `my mod.py`: `-z` recovers it as a single path and the file resolves.

`SRC_ROOT = REPO_ROOT / 'src'` and `test_the_width_scan_sees_the_repo` both read well. The skip-vs-fail split in the guard is the right call: skip when there is genuinely no repo, fail when there is one and the enumeration came back empty, so CI (which always has `.git`) still fails loudly if the scan ever stops resolving.

Two things left. One is inline; the other has no line in this diff to hang off:

**`pyproject.toml:258` — `ignore-long-lines` still contradicts the new test.**

```toml
ignore-long-lines='^\s*(# )?(See )?(and )?<?https?://\S+>?$'
```

`_over_long_lines`'s docstring says "No exemptions", but pylint still whitelists a bare-URL comment line. A contributor adding a long permalink comment passes `pylint` and `autopep8` locally, then fails a test whose message cites "the configured width" — pointing them at a setting that says the line is fine. Nothing in the tree needs the exemption (0 violations either way), so deleting that key is what makes the config and the test agree.

darkclad (approved):

PR #398, opened 2026-09-01
fix(nextflow) : fix two missed defects and simplify code and tests alike

vjaganat90:
## What this is

A post-merge adversarial review of the merged Phase 1 baseline (`4b3fced`) confirmed the backend implements the locked design (`design_docs/sophios_nextflow_backend.md`) but found a small set of edge-scoped semantic gaps plus structural debt. This PR closes all of them, fail-closed, and consolidates the code and tests without changing the supported Phase 1 subset. The locked design document is untouched: every rejection added here narrows behavior the design never promised, so no `design:` commit is required.

Every behavioral fix lands together with the regression test that pins it, in the same commit.

### `92aaedc` fix(nextflow): fail closed on post-merge review gaps

1. **Connection order could select the staging policy for mixed `path_kind` sinks.** The workflow-input channel agreement check compared only qualifiers, so one parameter feeding a file-kind port and a directory-kind port was accepted and staged as whichever connection appeared first — a direct violation of "connection order never selects a qualifier or staging policy". `path_kind` now participates in the agreement check and mismatches reject at IR construction.
2. **The reader partially parsed a call with one opaque argument.** A misplaced `continue` in `_parse_workflow` recorded the line as opaque yet still emitted connections for the call's other arguments, so structural import could silently misrepresent topology (e.g. `TASK(foo { bar }, plain)` produced a connection for the second port and dropped the first). One unparsable argument now makes the whole call opaque and contributes no connections.
3. **The renderer was not total over the executable model.** `NfPort` admitted `tuple`/`env`/`stdin` qualifiers that no lowering produces and the renderer would emit as invalid DSL2. The executable qualifier set narrows to the lowered subset (`path`, `val`); construction and strict hydration both reject the rest.
4. **`$(inputs.x.path)` and `$(inputs.x.basename)` were silently conflated** into the same input reference. `.basename` now rejects with a Phase 2 deferral diagnostic instead of relying on undocumented staging coincidence.
5. **An absent optional workflow input feeding a required tool input escaped capability analysis** and flowed through as `params[x] = null` into channel construction. Wired inputs now reject when any boundary source resolves to null, regardless of the tool input's optionality.
6. **Missing local Nextflow silently skipped the runtime suites.** With `SOPHIOS_REQUIRE_NEXTFLOW=1` the harness fails instead of skipping; the required `nextflow-runtime` job sets it, so a PATH regression between steps can no longer skip 14 runtime tests and pass.

### `349830a` refactor(nextflow): share one topological sort and trim dead surface

- Three hand-rolled Kahn loops (IR cycle validation, renderer ordering, reader acyclicity) are replaced by one `graphlib.TopologicalSorter`-backed helper pair in `nf_types` (`topological_order`, `process_dependencies`). Renderer ordering keeps its exact deterministic batch order via the position key.
- Groovy literal/GString escaping uses precomputed `str.translate` tables instead of per-character loops.
- Scalar-default validation, previously duplicated verbatim for tool inputs and workflow inputs, is one helper.
- `write_nextflow_files` and `write_nextflow_json` are removed. Neither had a caller anywhere in the repository, and the design prescribes a single writer (`to_nextflow` / `write_nextflow_artifacts`).
- Renderer branches made unreachable by IR construction invariants are removed (output qualifier/glob rechecks, duplicate-source recheck, unconnected-input recheck), along with the reader's dead `// TODO: scatter` branch and an inert `start` parameter.

### `657ef84` test(nextflow): reorganize suites by architecture seam

`tests/core/test_nextflow_{api,properties,public_api}.py` (one 2,300-line file plus two satellites) become `tests/core/nextflow/`, one file per pipeline stage, so the layout matches the locked architecture:

| File | Question it answers |
|---|---|
| `test_capabilities.py` | what rejects before artifacts |
| `test_lowering.py` | what supported source lowers to |
| `test_executable_ir.py` | what the validated IR can and cannot represent |
| `test_renderer.py` | what the deterministic artifacts look like |
| `test_reader.py` | what structural import retains and promotes |
| `test_public_api.py` | what Python/CLI users are promised |
| `test_runtime.py` | what real pinned Nextflow proves (R1–R11) |
| `test_properties.py` | adversarial property suites |
| `testkit.py`, `conftest.py` | shared builders, run harness, compiled fixtures |

The run harness and rose/tool/workflow builders that were duplicated across files now live once in the testkit. Four tests that exactly duplicated a stronger sibling were merged into it (two compile-target duplicates, one CLI-flag subset, one redundant runtime execution); every other behavioral assertion is carried over verbatim. Net test SLOC decreases (2,923 → 2,850) while nine new defect regressions are added. CI paths in `nextflow.yml` follow the move.

### `0531de3` ci: pin nodejs below 26.8 so cwltool can parse its version

Cherry-pick of `d51bac1` from `master`: conda-forge published nodejs 26.8.0 with a pre-release version string that cwltool's sandboxjs cannot parse (`invalid literal for int(): '0-alpha'`), which broke every cwltool-invoking CI step repository-wide. Needed here so this branch's CI can run; content is identical to the master commit.

### `00f641f` test(nextflow): build fixtures through the public APIs, not raw CWL

The CLI runtime test drives the checked-in `cwl_adapters/touch.cwl` through a `write_wic()`-generated workflow instead of heredoc CWL/`.wic` text (container path, docker-marked, shared `require_docker` guard, asserts `docker.enabled` and the staged output). The adversarial-argument and supported-rose fixtures are built with the `tool_builder` API instead of raw CWL dictionaries. Capability suites intentionally keep synthetic raw documents: they must express unsupported shapes the builders refuse to produce.

### `e1e97da` test(nextflow): one tool-construction idiom per test

Each test builds its workflow either from checked-in `cwl_adapters` or with `tool_builder`, never both; the parametrized writer test is split in two.

### `693f0f0` test(nextflow): prove mixed adapter/builder workflows execute end to end

New runtime case: an adapter-imported echo step fans out to two `tool_builder` consumers (`cp` and `wc -c`) in one workflow, compiled through `to_nextflow` and executed under pinned Nextflow with content and byte-count assertions — the deliberate, documented exception to the one-idiom rule, proving both public construction surfaces compose in a single executed workflow. Pure addition; no existing test changes.

### `47100b5` fix(nextflow): CLI runtime fixture step name must resolve its adapter

`write_wic()` serves the step id to the CLI as the tool lookup key, so a step name differing from the adapter stem failed compilation. The fixture uses the default step name so `search_paths_cwl` resolves `touch.cwl`.

### `97a6c44` docs(nextflow): Args/Returns docstrings on every public callable

All public functions and public class methods across `nf_symbols`, `nf_types`, `nf_reader`, `utils_nf`, and `input_output_nf` now follow the Google-style `Args:`/`Raises:`/`Returns:` pattern the Sphinx napoleon dev docs are generated from; private (`_`-prefixed) helpers keep one-line summaries. `is_nextflow_identifier` now returns `TypeGuard[str]` and `normalize_nextflow_identifier` takes `str`; the boundary validators take `object` by design and produce the proven `str`.

### `94a7892` fix(nextflow): stop mislabeling absent required inputs as deferred-optional

Fix #5 above widened the wired-input check to also catch a required tool input fed by an absent workflow value, but the finding message unconditionally read "absent optional values are deferred to Phase 2" regardless of which case fired. The message now reflects which case triggered it; the regression test whose name already said `feeding_required_input` now asserts the corrected text instead of the stale optional-flavored one.

### `1506a06` fix(nextflow): fail Docker regressions in CI instead of skipping them

`require_docker()` and the pinned-image pull check always skip when Docker is unavailable, with no `SOPHIOS_REQUIRE_NEXTFLOW`-style escape hatch, so the only test that drives the full CLI subprocess end to end could pass in CI while exercising nothing. `SOPHIOS_REQUIRE_DOCKER` mirrors the existing Nextflow-binary gate and is set in the CI job that runs on a Docker-equipped runner.

### `62fc23e` fix(nextflow): validate before serializing in write_nextflow_artifacts

`write_nextflow_artifacts` called `workflow.to_json()` before `_require_executable`, so a non-executable value raised `AttributeError` instead of the `TypeError` its own docstring promises, the only one of the four Nextflow renderers that didn't validate first. Now matches the sibling renderers and the docstring; a regression test pins the contract.

### `7ca0bea`, `015fef7` fix(nextflow): repoint touch/cat adapters off the broken bash:4.4 image

`docker.io/bash:4.4` was rebuilt upstream without `/bin/bash` at that path (verified every current `bash:*` tag shares this, not just 4.4), so Nextflow's docker executor exits 127 on every process using it. `cwl_adapters/touch.cwl` and `cwl_adapters/cat.cwl` now pin `debian:bookworm-slim` — the smallest official image I could confirm has a real `/bin/bash` under this executor (ubuntu:24.04 worked too but is needless bulk for two tools that just need touch/cat).

### `f0c8c59` fix(nextflow): drop redundant docker/nextflow availability probing

With the image pin fixed, `SOPHIOS_REQUIRE_DOCKER`/`SOPHIOS_REQUIRE_NEXTFLOW` and the `require_docker()`/`nextflow_executable()` probing behind them were solving a problem `nextflow.yml`'s own marker-based job selection already solves on a runner that always has both tools. Removed both helpers, the "pull" skip heuristic, and the env vars; docker/nextflow-dependent tests now call the tool directly and fail like any other assertion if it's missing. Verified the full suite end to end with real Docker and Nextflow and no env vars set.

### `b80dc1c` ci: exclude nextflow/docker tests from the package-build pytest run

Removing the runtime skip broke `build_wheel.yml`/`test_and_publish_pypi.yml`, which run the full suite with no Nextflow installed and were relying on it. Both `nextflow` and `docker` markers are scoped only to `tests/core/nextflow`, so excluding them in those jobs' own `-m` expression is the direct fix and drops nothing else — the same job-declares-its-scope approach `nextflow.yml` already uses.

### `015fef7` fix(nextflow): pin touch/cat adapters to debian:bookworm-slim, not ubuntu

`ubuntu:24.04` fixed the broken `/bin/bash`, but is a full distro for two tools that only need `touch`/`cat` and a shell. Every current `bash:*` tag shares the same missing-`/bin/bash` defect, so there's nothing smaller left in that family; `debian:bookworm-slim` is the smallest official image confirmed to work under Nextflow's docker executor.

### `fc7ce5b` fix(nextflow): finish propagating the touch/cat image swap

Two spots still pinned the old images after the adapters moved: `examples/scripts/ground_truth_multistep1.json` inlines `touch.cwl`/`cat.cwl`'s `DockerRequirement` verbatim, so its `expected_workflow == compiled.cwl_workflow` assertion broke silently (nothing in CI runs that script); and `test_docker_container_behavior` built its own container directly rather than through an adapter, so the `-m docker` job was still pulling `ubuntu:24.04` alongside `debian:bookworm-slim`.

## Evidence

The runtime conformance cases run under the required pinned-Nextflow job (`nextflow-runtime`: Nextflow 26.04.2, Java 17, Python 3.12).

Scope is strictly the Nextflow backend: `nf_types`, `nf_reader`, `utils_nf`, `input_output_nf`, `api/python/nextflow`, `tests/core/nextflow/`, `.github/workflows/nextflow.yml`, plus the cherry-picked dependency pin.


darkclad (changes requested):
## Review

Reviewed `4b3fced...pr-398-head` (22 files, +2949/−2544).

Both headline defect fixes are correct and properly pinned, and the refactors are faithful. Three findings below.

---

### 1. The one true end-to-end test is now Docker-gated and self-skipping

`tests/core/nextflow/test_runtime.py:307`

The only test that drives the real `python -m sophios.main --target nextflow` subprocess end to end now sits behind `require_docker()` + `@pytest.mark.docker`, because the new fixture uses `cwl_adapters/touch.cwl` with `dockerPull: docker.io/bash:4.4`. On top of that, a failed run whose output contains `"pull"` is converted into a `pytest.skip`.

On the base commit this test ran unconditionally against a container-free adapter with a hard `returncode == 0` assert.

This is the same silent-skip failure mode that fix #6 (`SOPHIOS_REQUIRE_NEXTFLOW`) sets out to eliminate, but there is no `SOPHIOS_REQUIRE_DOCKER` counterpart — so in CI this can pass while testing nothing.

### 2. Docstring contract does not match behavior

`src/sophios/input_output_nf.py:315`

The new docstring promises `TypeError` for a non-executable argument, but `workflow.to_json()` runs before any validation, so `write_nextflow_artifacts({}, tmp)` raises `AttributeError` instead. The three sibling renderers all call `_require_executable` first; only `render_nextflow` has a test pinning the contract.

### 3. Stale message after widening the check

`src/sophios/utils_nf.py:969`

`_absent_optional_findings` was widened from optional-only to all tool inputs (fix #5) but kept the message `absent optional values are deferred to Phase 2`. A **required** tool input is now reported as an absent *optional* value — visible in this PR's own new regression test.

---

### Verified sound

- **`path_kind` in the channel-agreement check** (`nf_types.py` `_validate_graph`) genuinely closes the hole. `_parameter_expression` / `_workflow_input_port` pick the *first* connection's port, and `qualifier` + `path_kind` fully determine the rendered `Channel.fromPath(..., type: ...)` / `Channel.value(...)` — nothing else about the destination port leaks into the channel, so the fix is complete rather than partial.
- **The `break` + `for/else` in `_parse_workflow`** is right: an unparsable argument in any position discards the whole `call_connections` list and records the line as opaque. The zero-argument and `_split_arguments("")` edge cases both behave.
- `topological_order` via `graphlib.TopologicalSorter` reproduces the old batch-level Kahn semantics exactly, and `key=position.__getitem__` preserves deterministic batch ordering.
- The removed renderer guards really are unreachable: `NfProcess.__post_init__` enforces `path` qualifier + non-`None` glob on outputs, and `_validate_graph`'s `_record_incoming` + `expected_inputs - incoming` enforce exactly-one-source and full connectivity. The `assert port.glob is not None` in `_process_output` is backed by a real raise upstream, so it holds even under `-O`.
- `str.translate` escaping tables are character-for-character equivalent to the old loops (no key overlap between control escapes and `` \ ``/`'`/`"`/`$`).
- `_default_value_findings` is an exact merge of both duplicated blocks; the removed `default` local in `_workflow_capability_findings` is unused afterwards.
- `_input_binding_items`' dropped `start` param was dead (`del start`), and the sort keys `(pos, 0, int)` vs `(pos, 1, str)` never compare across types.
- The removed `// TODO: scatter` reader branch was byte-identical to its `else` and had no test coverage.
- `NfPort.ALLOWED_QUALIFIERS` narrowing is safe: the reader uses its own `NextflowPort`, and the only `NfPort(...)` construction site derives its qualifier from `cwl_type_to_nf_qualifier`, which returns only `path`/`val`.
- No stale references to the deleted `write_nextflow_files` / `write_nextflow_json` anywhere in the tree.

**On the test move:** I diffed every `pytest.raises(..., match=...)` string between the three old files and the new `tests/core/nextflow/` package — zero negative assertions were lost. The `schema_version` / `representation_kind` assertions dropped from `test_public_api.py` are covered at `test_executable_ir.py:69`. `REPO_ROOT = parents[3]` is correct for the new depth, and the package imports resolve (`tests/core/__init__.py` exists, `tests/__init__.py` does not, so relative imports work under pytest's prepend mode). No other workflow references the old test paths.

---

### Out of scope

`codespellwords.txt` gains `nd`, but the only `nd` token in the tree is the pre-existing Unicode category name at `src/sophios/nf_symbols.py:152`. That entry suppresses the real `nd` → `and` check repo-wide, and is unrelated to this PR's stated scope.

darkclad (approved):
All findings resolved across three rounds. Ground truth and the docker test now match the adapters, CI green on fc7ce5b. Agreed on leaving CONTRIBUTING and the example-executing CI step out of scope.

PR #399, opened 2026-09-01
refactor: one home for the wic: vocabulary, one home for the generators

vjaganat90:
Two structural gaps Spec 1 left behind, plus the test-surface tidy that falls out of the second.

**The `wic:` vocabulary had no home.** The set of keys a `wic:` block admits lived only in a dict literal inside the validating schema, unlinked to the language. That is why `lang_version` reached §7, the resolver and the compiler while the validator still rejected the document before compilation began. `Grammar.SIDECAR_KEYS` now names the vocabulary where the rest of the language is named, and a test requires the validator to admit exactly those keys — no more, no fewer. The shapes stay with the validator; only the vocabulary is shared. Removing `lang_version` from the validator now fails that test.

**The generators had no home either.** Three suites quantify over the same language, and the parser suite owned the strategies the other two imported from it. A second generator beside the first is how outputs and nested sidecars became invisible to every property fed by one of them, so this is the file most worth having one of. Moved to `tests/core/strategies.py`, extracted unchanged.

**Nine single-claim parser tests were one test written nine times.** Each parsed a source, checked the verdict, and inspected one thing. They are now two tables, `ACCEPTED` and `REPORTED`, each driving one parametrized test; 34 test functions become 28, and every row stays a named case in the report. Breaking the parser site behind a row fails that row. One row is new: `_step` documents and handles a step with no body at all (`some_step.wic:` with nothing under it), reachable from the CLI, and suppressing that branch failed no test.

**Eight parser branches had no test behind them.** Coverage over the language suites found them once the tables made them cheap to add: a step that is not a mapping, a repeated step-body key, a second `id:` inside a body, an `out:` entry that is neither a name nor a single-key mapping, a `wic:` block reached through its own alias, a repeated `wic:` step key, an uninterpreted step key, and the literal-text fallback for a scalar YAML cannot load. All eight are reachable from `parse()`. Each is now a row, and breaking the branch fails only that row. The alias case is not the one `test_alias_cycles_are_reported[sidecar]` covers — that source is caught earlier, in `_opaque`, and never reaches the guard in `_sidecar`, so the existing name promised more than it checked.

**The renderer never had to choose a spelling in a test.** A parsed literal remembers how it was written and is transcribed verbatim, so every round-trip property — including `@example('!ii .nan')`, which looks like it covers this — takes the transcription path. Documents built through the Python API have no source text, and that is the only way into `_spell_scalar`. The NaN arm was the branch this hid: NaN is the one value not equal to itself, so the equality check below it cannot accept one. Four synthesized values now pin the choice.

**`test_alias_cycles_are_reported[sidecar]` asserted a cycle and checked nothing of the kind.** Its source hit an unrelated malformed-key check before ever reaching `_sidecar`'s recursion guard, so it passed on `MALFORMED_WIC_STEP_KEY` while its name claimed a cycle report. Changed to the shape that actually recurses, and pinned `RECURSIVE_ALIAS` on it and its two siblings rather than asserting only failure.

**Dropped `WicSidecar.entry` and `Document.step`.** Neither has a single caller in `src`, `tests` or docs — they were the untested half of an accessor pair whose used half, `Step.input`, stays.

**Deleted:** a `scalars` strategy with no users (already dead on master), and `test_tags_map_to_their_forms`, whose four cases `test_surface_forms_are_equivalent` already asserts — breaking `!&` so it builds an `EdgeRef` fails that test. Its one unique case, a bare name resolving to `UnresolvedName`, survives as a row in `ACCEPTED`.


tests/core/test_lang_parser.py
```diff
@@ -369,26 +227,117 @@ def _payload(value: InputValue) -> Any:
...
+@pytest.mark.parametrize('case', REPORTED, ids=[case.claim for case in REPORTED])
+def test_reported_shapes(case: Reported) -> None:
+    """Each diagnosed shape reports its code, with a span, and recovers or not."""
+    result = parse(case.source, 'reported.wic')
+    matching = [d for d in result.diagnostics if d.code is case.code]
+    assert matching, f'{case.code} did not fire: {[str(d) for d in result.diagnostics]}'
+    assert all(d.span is not None for d in matching), 'diagnostic without a span'
+    if case.message_contains:
+        assert any(case.message_contains in d.message for d in matching)
+    assert (result.document is not None) is case.recovers
```
darkclad, line 345: **🟢 The fold dropped "exactly one diagnostic" and "on this line", and both fit back in one field.**

Two of the deleted tests asserted more than this one does:

```python
# test_repeated_input_is_reported_not_silently_resolved
assert [d.code for d in result.diagnostics] == [Code.DUPLICATE_KEY]   # nothing else fired
assert result.diagnostics[0].span.start_line == 5                     # here, not somewhere

# test_malformed_yaml_reports_a_located_diagnostic
assert len(result.diagnostics) == 1
```

`test_reported_shapes` asserts a matching diagnostic exists and its span is not `None`. `test_diagnostic_spans_index_real_source` bounds spans to the document but does not pin them — so a regression that reported `DUPLICATE_KEY` on line 1, or that fired a spurious second diagnostic alongside the right one, now passes every test in the suite. That is the failure mode the table is otherwise so good at catching.

It is cheap to restore, because the table already has the discipline for it. I ran all nine rows: **every one produces exactly one diagnostic, on a determinate line.**

| Row | Code | Line |
|---|---|---|
| collection step key | `EXPECTED_SCALAR` | 2 |
| input bound twice | `DUPLICATE_KEY` | 5 |
| step not a mapping | `EXPECTED_MAPPING` | 2 |
| step body key repeated | `DUPLICATE_KEY` | 4 |
| second `id:` in body | `DUPLICATE_KEY` | 3 |
| bad `out:` entry | `EXPECTED_SCALAR` | 3 |
| `wic:` through its own alias | `RECURSIVE_ALIAS` | 1 |
| repeated `wic:` step key | `DUPLICATE_KEY` | 4 |
| malformed YAML | `INVALID_YAML` | 3 |

So one integer per row gets both assertions back — for nine rows instead of the two that used to carry them. On `Reported`, between `code` and the defaulted fields:

```python
    code: Code
    #: The line the diagnostic must land on. Every row reports exactly one,
    #: so the position is determinate and worth pinning.
    line: int
```

and here:

```python
    result = parse(case.source, 'reported.wic')
    assert len(result.diagnostics) == 1, [str(d) for d in result.diagnostics]
    matching = [d for d in result.diagnostics if d.code is case.code]
    assert matching, f'{case.code} did not fire: {[str(d) for d in result.diagnostics]}'
    assert all(d.span is not None for d in matching), 'diagnostic without a span'
    assert [d.span.start_line for d in matching] == [case.line], 'reported on the wrong line'
    if case.message_contains:
        assert any(case.message_contains in d.message for d in matching)
    assert (result.document is not None) is case.recovers
```

(`--no-strict-optional` is on, so `d.span.start_line` past the `is not None` assert type-checks fine.) Not offered as a suggestion because the nine rows have to grow the field in the same commit.

src/sophios/lang/parser.py
```diff
@@ -49,6 +49,17 @@ class Grammar:  # pylint: disable=too-few-public-methods  # a namespace, not a t
     #: makes the leak boundary a specification rather than an accident.
     INTERPRETED_STEP_KEYS: Final = frozenset({'scatter', 'scatterMethod', 'when', 'run'})
 
+    #: Every key the `wic:` block admits. The block is Sophios-owned syntax
```
darkclad: **🟡 "Every key the `wic:` block admits" — `inference` is a `wic:` key nothing here declares.**

There is one more, and it has the shape of the bug this constant exists to prevent:

- `inference.py:381-382` — `if 'inference' in wic.get('wic', {}): rules = wic['wic']['inference']`
- `compiler.py:1259` — `inf_dict = {'wic': {'inference': inference_rules_dict}}`, installed under `yaml_tree_mod['wic']['steps'][keystr]`, so it lands inside a nested `wic:` block

Validated against the schema this PR now pins:

```
{'inference': {...}}                                  -> Additional properties are not allowed
{'steps': {'(1, x)': {'wic': {'inference': {...}}}}}  -> Additional properties are not allowed
{'lang_version': '0.0.1'}                             -> OK
```

I traced whether that is reachable from user source, and it is: `compiler.py:187` reads `wic_steps = wic['wic'].get('steps', {})` straight off the user's `yaml_tree`, `compiler.py:1004` hands it to `perform_edge_inference`, and `inference.py:227` calls `get_inference_rules` on it. So `wic: {steps: {(1, foo): {wic: {inference: {...}}}}}` is a per-step inference-rules override the compiler will honour and the validator refuses before compilation starts — `lang_version` again, one layer down.

**This predates the PR and is not a regression.** No `.wic`, `.md` or `.yml` in the repo writes it, and `docs/advanced.md:196-208` documents `inference_rules` only as a config-file key, so the path is dead today. The wider question deserves its own issue rather than this branch.

What belongs here is just the claim. Either add `inference` to both sides if the per-step override is meant to work, or say what the set actually is:

```suggestion
    #: Every key the `wic:` validator admits. The block is Sophios-owned syntax
```

tests/core/test_lang_schema.py
```diff
@@ -332,3 +332,28 @@ def test_an_empty_step_id_is_rejected_because_the_parser_reports_it() -> None:
...
+@pytest.mark.fast
+def test_the_validator_admits_exactly_the_declared_sidecar_keys() -> None:
+    """`Grammar.SIDECAR_KEYS` and the validating schema name the same keys.
+
+    They are separate places by necessity — the language layer owns the
+    vocabulary, the validator owns each key's shape — and nothing linked them,
+    so `lang_version` reached the reference, the resolver and the compiler
+    while the validator still rejected it before compilation ever started.
+    """
+    from sophios.schemas.wic_schema import _wic_tag_schema
```
darkclad: **⚪ The function-local private import wants its reason.**

`C0415: Import outside toplevel`, new in this PR, and it reaches for an underscore-private name across a package boundary. If the local placement is load-cost or cycle avoidance, this file's comment density says it should say so in a clause; if it is neither, it belongs at the top with the other imports.

tests/core/test_lang_schema.py
```diff
@@ -332,3 +332,28 @@ def test_an_empty_step_id_is_rejected_because_the_parser_reports_it() -> None:
...
+
+    validated = set(_wic_tag_schema(hypothesis=False)['properties'])
+    declared = set(Grammar.SIDECAR_KEYS)
+
+    assert validated == declared, (
+        f'declared but not validated: {sorted(declared - validated)}; '
+        f'validated but not declared: {sorted(validated - declared)}'
+    )
```
darkclad, line 358: **⚪ The `hypothesis=True` schema is not pinned, and that is the branch the property tests run against.**

`_wic_tag_schema` builds `schema_props` in two modes — `implementations` is added only `if not hypothesis` (`wic_schema.py:388-390`). This checks `hypothesis=False` only, so a key added inside that guard keeps the assertion green while the two modes diverge. Same drift, one branch over. One line pins it:

```python
    hyp = set(_wic_tag_schema(hypothesis=True)['properties'])
    assert declared - hyp == {'implementations'}, 'the hypothesis schema drifted from the vocabulary'
```

darkclad (commented):
Reviewed at `57b5aa6` against merge base `ccb8063`. I checked out the branch and ran the three touched suites, coverage, pylint and mypy against a `ccb8063` baseline, and probed the parser directly for the claims that turn on which branch actually fires.

**Every claim in the description holds, and the coverage one is larger than it sounds.** Measured with `--cov=sophios.lang --cov-branch` over `test_lang_parser` + `test_lang_render` + `test_lang_schema`:

| Module | `ccb8063` | `57b5aa6` |
|---|---|---|
| `parser.py` | 93% — 19 statements, 11 branches missed | **99%** — 2 statements, 2 branches |
| `render.py` | 96% — `212` missed | **97%** |
| `nodes.py` | 96% — `92, 237, 258` missed | **98%** — `92` |

`render.py:212` is the `_spell_scalar` NaN arm, so the fifth paragraph is exactly right about why the round-trip properties could not reach it. `nodes.py:237, 258` were the two dropped accessors, and I confirmed neither they nor the `scalars` strategy have a caller anywhere in `src`, `tests` or `docs`.

**The alias finding is right, and sharper than stated.** Running all four sources through `parse()`:

```
wic: &w / steps: *w                    -> MALFORMED_WIC_STEP_KEY   (the old [sidecar] case)
wic: &w / steps: / (1, a): / wic: *w   -> RECURSIVE_ALIAS          (the new row)
```

One correction: the old case is not "caught earlier, in `_opaque`". `key == 'steps'` routes it into the steps branch, where `_step_key('steps')` returns `None` and reports a malformed step key — `_opaque` is never called on it. The conclusion is unaffected.

**The fold pays a bill I didn't expect.** 860 → 738 lines clears the `C0302: Too many lines in module (860/750)` that `ccb8063` was tripping against `max-module-lines=750`, and the deleted duplicate-input test took three `E1101: Instance of 'Diagnostics' has no 'message'/'span' member` errors with it. Pylint over the six touched files: **9.77 → 9.92**. mypy unchanged.

**Nothing blocking.** One thing is worth a fix on its own — the fold dropped two assertions, inline below, in a PR whose thesis is that a test should fail when the thing it names breaks. Everything else is a batch worth folding into the same push.

Two of those have no line in this diff to hang off:

**⚪ `test_lang_parser.py:25` — `Diagnostic` is now an unused import.** `W0611`, new in this PR. Its only user was `first: Diagnostic = result.diagnostics[0]` in the deleted `test_malformed_yaml_reports_a_located_diagnostic`; the surviving mention at line 147 is inside a docstring. (`nodes.py:15 Unused Final` also fires but is pre-existing on `ccb8063` — not yours.)

**⚪ `test_lang_parser.py:495-506` — `test_alias_cycles_are_reported[sidecar]` is still the case you name in the description.** You identify it as over-promising and add the real coverage in `REPORTED`, but leave the misleading row in place: `'wic: &w\n  steps: *w\n'` still asserts only `not result.ok`, and still passes on `MALFORMED_WIC_STEP_KEY`. Now that the genuine `_sidecar` cycle has a row, this one either wants its source changed to the real shape or an `assert any(d.code is Code.RECURSIVE_ALIAS ...)`, so the name and the check agree.

tests/core/test_lang_parser.py
vjaganat90: Verified all nine rows produce exactly one diagnostic on your predicted line before changing anything. Added `line: int` and the count+line assertions; revert-checked both against a wrong-span break and a spurious-second-diagnostic break, and both fail now.

src/sophios/lang/parser.py
vjaganat90: Traced it the same way and got the same answer: reachable, pre-existing, not a regression. Took the minimal fix — "validator admits" instead of "block admits" — since that's what this set actually is. Left the inference gap alone; it's a real finding but not this branch's to close.

tests/core/test_lang_schema.py
vjaganat90: Lifted it to the top import block.

tests/core/test_lang_schema.py
vjaganat90: Good catch — pinned. hypothesis=True is exactly the branch the property suites run against, so checking only hypothesis=False was the same gap the whole test exists to prevent, one guard over.

darkclad (approved):

PR #400, opened 2026-09-02
feat(nextflow): lower boolean inputBinding flags

vjaganat90:
Stacked on #398 (`nfplan_1followon`).

## What this adds

Boolean `inputBinding` flags lower to real command-line flags instead of Phase 1's blanket rejection.

- A boolean input with an `inputBinding` prefix and no `valueFrom` lowers to a typed `NfFlag` token: `true` renders the shell-quoted prefix as one argv word, `false` renders nothing.
- No prefix means no contribution either way, matching CWL.
- The prefix goes through the generated shell-quote helper, so a hostile prefix can't inject shell syntax.

## Boundaries

- `NfFlag` is only valid in command token position — unrepresentable in globs or stream targets by type.
- Flags validate against `val` ports at construction, not render time.
- `valueFrom` on a boolean binding keeps its existing template treatment; absent optional booleans still reject.

## Schema

`flag` registers as the token kind introduced in `schema_version: 3`, the current version; hydration still accepts the strict-subset version 2. Evolution policy is in the design doc.

## Design authority

`design:` commit `09b5ef2` approves this per §10 of `design_docs/sophios_nextflow_backend.md`.

## Review fixes

- `valueFrom` on a boolean binding emitted an inverted flag (fell through the "is this a flag" check, so `false` still rendered the prefix). Now rejects in capability analysis instead.
- Nothing required the flag's source to actually be boolean — a `File` output pinned the flag on via Groovy truthiness. Capability analysis now resolves every flag source's type and requires boolean.
- `separate` without a `prefix` now matches cwltool exactly (including `separate: null`); a command whose argv is only conditional flags keeps its fallback program instead of rendering empty.
- A whitespace-only prefix now hits the named capability diagnostic instead of an unrelated `NfFlag` construction error.
- Design doc corrections: dropped a `valueFrom` rejection claim that never existed, fixed "non-optional" wording (here and in `docs/nextflow.md`), and made the schema-subset invariant actually enforced per-kind instead of just asserted.
- Closed coverage gaps the story declared but hadn't tested: flag ordering among other bindings, byte-stability with a flag present, and flag-bearing artifacts round-tripping through parse/promote.
- `NfFlag`/`NfCommandToken` weren't actually exported from `sophios.api.python.nextflow` despite an earlier version of this description claiming they were. Fixed and pinned with a test.
- Round 2: the flag-source guard's exemption was narrower than it needed to be and cost three mapping shapes their aggregating diagnostic — widened to any `Mapping`. None of the three `separate`/prefix/source-shape fixes above had a test that actually distinguished old from new behavior; added the three missing cases.

### DIFF
--- a/docs/nextflow.md
+++ b/docs/nextflow.md
@@ -76,11 +76,19 @@ written from validated executable IR. Parsed source alone is never promoted by
 guessing shell or Groovy semantics; changed source, changed parameters, unknown
 content, and invalid or stale IR fail closed.
 
-## Phase 1 limits
+## Boolean flags
+
+A non-optional `boolean` input with an `inputBinding` prefix and no
+`valueFrom` becomes a real command-line flag: `true` contributes the prefix as
+exactly one argument and `false` contributes nothing. A boolean binding
+without a prefix contributes nothing for either value, matching CWL. Absent
+optional booleans still reject until option lowering is approved.
+
+## Current limits
 
 - Workflows must be flat and use `CommandLineTool`-equivalent processes.
-- Boolean `inputBinding` flag semantics and fractional CPU requirements reject
-  before lowering; they are not silently stringified or rounded.
+- Fractional CPU requirements reject before lowering; they are not silently
+  rounded.
 - Scatter is retained as opaque structure; executable scatter is deferred.
 - Nested workflows, arbitrary Groovy, channel operators, `when`, and `exec`
   blocks are not interpreted.
--- a/docs/nextflow.md
+++ b/docs/nextflow.md
@@ -78,7 +78,7 @@ content, and invalid or stale IR fail closed.
 
 ## Boolean flags
 
-A non-optional `boolean` input with an `inputBinding` prefix and no
+A `boolean` input with an `inputBinding` prefix and no
 `valueFrom` becomes a real command-line flag: `true` contributes the prefix as
 exactly one argument and `false` contributes nothing. A boolean binding
 without a prefix contributes nothing for either value, matching CWL. Absent
--- a/src/sophios/utils_nf.py
+++ b/src/sophios/utils_nf.py
@@ -12,7 +12,9 @@ from .nf_types import (
     ExecutableNextflowWorkflow,
     NF_INTERNAL_IDENTIFIERS,
     NfCommand,
+    NfCommandToken,
     NfConnection,
+    NfFlag,
     NfInputReference,
     NfLiteral,
     NfPort,
@@ -253,8 +255,8 @@ def _argument_items(arguments: list[Any]) -> list[tuple[tuple[int, int, int], tu
 
 def _input_binding_items(
     inputs: Mapping[str, Any],
-) -> list[tuple[tuple[int, int, str], tuple[NfTemplate, ...]]]:
-    items: list[tuple[tuple[int, int, str], tuple[NfTemplate, ...]]] = []
+) -> list[tuple[tuple[int, int, str], tuple[NfCommandToken, ...]]]:
+    items: list[tuple[tuple[int, int, str], tuple[NfCommandToken, ...]]] = []
     for raw_name, definition in inputs.items():
         input_definition = _as_mapping(definition, error=f"CWL input {raw_name!r} must be a mapping")
         match input_definition.get("inputBinding"):
@@ -266,16 +268,28 @@ def _input_binding_items(
                 raise ValueError(f"CWL inputBinding for {raw_name!r} must be a mapping")
         name = _identifier(raw_name, context="input binding name")
         value_from = binding.get("valueFrom")
+        position = _position(binding.get("position"), default=0)
+        if value_from is None and _required_type(input_definition.get("type")) == "boolean":
+            # CWL boolean bindings contribute their prefix, or nothing at all
+            # when the flag is false or no prefix is declared.
+            match binding.get("prefix"):
+                case None:
+                    continue
+                case str() as prefix if prefix:
+                    items.append(((position, 1, str(raw_name)), (NfFlag(name, prefix),)))
+                    continue
+                case _:
+                    raise ValueError(f"CWL command prefix for {raw_name!r} must be a non-empty string")
         value = (
             _template(value_from, context=f"CWL input {raw_name!r} valueFrom")
             if value_from is not None else NfTemplate((NfInputReference(name),))
         )
         tokens = _binding_tokens(binding.get("prefix"), value, separate=binding.get("separate", True))
-        items.append(((_position(binding.get("position"), default=0), 1, str(raw_name)), tokens))
+        items.append(((position, 1, str(raw_name)), tokens))
     return items
 
 
-def _command_items(tool: Mapping[str, Any]) -> tuple[NfTemplate, ...]:
+def _command_items(tool: Mapping[str, Any]) -> tuple[NfCommandToken, ...]:
     arguments = _as_list(tool.get("arguments", []), error="CommandLineTool arguments must be a list")
     inputs = _as_mapping(tool.get("inputs", {}), error="CommandLineTool inputs must be a mapping")
     ordered = sorted([*_argument_items(arguments), *_input_binding_items(inputs)])
@@ -284,7 +298,7 @@ def _command_items(tool: Mapping[str, Any]) -> tuple[NfTemplate, ...]:
 
 def _command(tool: Mapping[str, Any]) -> NfCommand:
     base_command = tool.get("baseCommand")
-    tokens: list[NfTemplate] = []
+    tokens: list[NfCommandToken] = []
     match base_command:
         case str() as command if command:
             tokens.append(_template(command, context="CWL baseCommand"))
@@ -715,14 +729,6 @@ def _tool_capability_findings(
                 if binding.get("shellQuote") is False:
                     findings.append(
                         f"{input_binding_path}.shellQuote: shellQuote false is deferred to Phase 2"
-                    )
-                if (
-                    _required_type(raw_definition.get("type")) == "boolean"
-                    and binding.get("valueFrom") is None
-                ):
-                    findings.append(
-                        f"{input_binding_path}: boolean inputBinding flag semantics are "
-                        "deferred to Phase 2"
                     )
         case _:
             pass
--- a/src/sophios/utils_nf.py
+++ b/src/sophios/utils_nf.py
@@ -226,6 +226,9 @@ def _position(value: Any, *, default: int) -> int:
 def _binding_tokens(prefix: Any, value: NfTemplate, *, separate: Any = True) -> tuple[NfTemplate, ...]:
     match prefix:
         case None:
+            if separate is not None and separate is not True:
+                # cwltool raises for separate without prefix, type-independently.
+                raise ValueError("CWL separate cannot be specified without a prefix")
             return (value,)
         case str() as text:
             prefix_template = _template(text, context="CWL command prefix")
@@ -274,6 +277,8 @@ def _input_binding_items(
             # when the flag is false or no prefix is declared.
             match binding.get("prefix"):
                 case None:
+                    if binding.get("separate") is not None:
+                        raise ValueError("CWL separate cannot be specified without a prefix")
                     continue
                 case str() as prefix if prefix:
                     items.append(((position, 1, str(raw_name)), (NfFlag(name, prefix),)))
@@ -315,8 +320,9 @@ def _command(tool: Mapping[str, Any]) -> NfCommand:
             raise ValueError("CommandLineTool baseCommand must be a string or list of strings")
 
     tokens.extend(_command_items(tool))
-    if not tokens:
-        tokens.append(_template("true", context="empty CWL command"))
+    if not any(isinstance(token, NfTemplate) for token in tokens):
+        # A command made only of conditional flags has nothing to execute.
+        tokens.insert(0, _template("true", context="empty CWL command"))
 
     def stream(name: str) -> NfTemplate | None:
         return None if tool.get(name) is None else _template(tool[name], context=f"CWL {name}")
--- a/src/sophios/utils_nf.py
+++ b/src/sophios/utils_nf.py
@@ -730,6 +730,15 @@ def _tool_capability_findings(
                     findings.append(
                         f"{input_binding_path}.shellQuote: shellQuote false is deferred to Phase 2"
                     )
+                if (
+                    _required_type(raw_definition.get("type")) == "boolean"
+                    and binding.get("valueFrom") is not None
+                ):
+                    findings.append(
+                        f"{input_binding_path}.valueFrom: valueFrom on a boolean inputBinding is "
+                        "deferred to Phase 2; CWL applies boolean flag semantics to the "
+                        "valueFrom result, which has no approved lowering"
+                    )
         case _:
             pass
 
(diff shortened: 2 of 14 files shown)


src/sophios/utils_nf.py
```diff
@@ -253,25 +271,39 @@ def _input_binding_items(
...
         value_from = binding.get("valueFrom")
+        position = _position(binding.get("position"), default=0)
+        if value_from is None and _required_type(input_definition.get("type")) == "boolean":
+            # CWL boolean bindings contribute their prefix, or nothing at all
+            # when the flag is false or no prefix is declared.
+            match binding.get("prefix"):
+                case None:
+                    if binding.get("separate") is not None:
+                        raise ValueError("CWL separate cannot be specified without a prefix")
+                    continue
```
darkclad, line 282: **🟡 `separate: true` without a prefix rejects here and is accepted one function away — cwltool accepts both.**

This uses `binding.get("separate") is not None`, so *any* explicit `separate` on a prefix-less boolean rejects. `_binding_tokens` at line 227 uses `separate is not True` against a `binding.get("separate", True)` default, which collapses "absent" with "explicitly true" — so the string path accepts what this path rejects. Measured on this head:

```
bool   no-prefix, separate absent : ACCEPTED
bool   no-prefix, separate: true  : REJECTED   <- cwltool accepts
bool   no-prefix, separate: false : REJECTED
string no-prefix, separate: true  : ACCEPTED
string no-prefix, separate: false : REJECTED
```

cwltool's rule is only about falsiness — `builder.py:618-626`:

```python
sep = binding.get("separate", True)
if prefix is None and not sep:
    raise WorkflowException("'separate' option can not be specified without prefix")
```

So the comment "cwltool raises for separate without prefix, type-independently" is true of neither the reference nor this file. Both sites also accept `separate: null`, where cwltool raises. One expression fixes all of it:

```python
# line 227
if not separate:
# line 280
if not binding.get("separate", True):
```

I applied exactly that and re-ran: the table becomes cwltool's, and 147 passed — `test_rejects_separate_without_a_prefix` pins `separate: False`, so it still bites.

src/sophios/utils_nf.py
```diff
@@ -253,25 +271,39 @@ def _input_binding_items(
...
+                case str() as prefix if prefix:
```
darkclad: **🟡 This guard is looser than `NfFlag`'s invariant, so a blank prefix escapes as an unmapped IR error.**

`nf_types.py:233` rejects `not self.prefix.strip()`; this admits any non-empty string. A whitespace-only prefix slips through here and dies on the invariant:

```
string, prefix "   " : ACCEPTED, lowers to NfLiteral('   ')
bool,   prefix "   " : ValueError: flag prefix must be a non-empty string
bool,   prefix ""    : ValueError: CWL command prefix for 'verbose' must be a non-empty string
```

Three things go wrong at once. The message names no input and no step, so it reads as an internal error rather than a source diagnostic. It fires during process construction, after `_raise_capability_findings`, so it does not aggregate — which is what the design document promises ("A rejected flag construct names the source path", "independent findings aggregate like every other capability diagnostic"). And it is a type-dependent divergence: cwltool emits `'   '` as the flag for a boolean exactly as it emits it as a literal for a string.

Pick one invariant and hold it at one place. Either relax `NfFlag` to `if not self.prefix`, or tighten this to `case str() as prefix if prefix.strip():` so the blank case falls into the `case _` arm below, which already produces the named, aggregating diagnostic.

docs/nextflow.md
```diff
@@ -76,11 +76,19 @@ written from validated executable IR. Parsed source alone is never promoted by
 guessing shell or Groovy semantics; changed source, changed parameters, unknown
 content, and invalid or stale IR fail closed.
 
-## Phase 1 limits
+## Boolean flags
+
+A non-optional `boolean` input with an `inputBinding` prefix and no
```
darkclad: **⚪ "non-optional" is the exact wording the design doc was corrected for in this PR.**

The description records the correction: *"the document ... said 'non-optional' where the code accepts optional booleans with values."* `design_docs/sophios_nextflow_backend.md` now reads "An optional boolean lowers the same way when a value is present." This file was rewritten in the same PR and still opens "A non-optional `boolean` input...".

Confirmed against the code at this head: `type: ["null", "boolean"]` with a value lowers to `NfFlag('verbose', '--v')`, and `test_rejects_absent_optional_boolean_flag` shows only the *absent* case rejects. The sentence four lines down ("Absent optional booleans still reject") is already correct — it is only the opening qualifier that is wrong.

src/sophios/utils_nf.py
```diff
@@ -966,6 +995,98 @@ def _absent_optional_findings(
...
+            raw_source = step_inputs.get(raw_name)
+            if raw_source is None:
+                continue
+            try:
+                sources = _source_values(
+                    raw_source,
+                    context=f"step input {step_index}.{raw_name}",
+                )
+            except ValueError:
+                continue
```
darkclad: **⚪ This fail-open sits inside a fail-closed guard.**

Skipping a step input whose shape `_source_values` cannot read is harmless today — I checked the two shapes that reach it, `{"default": ...}` and `{"source": ..., <extra>}`, and both are rejected elsewhere with their own diagnostics:

```
steps[0].in.verbose.default: default is not consumed by Nextflow Phase 1 lowering
```

But the coupling is invisible. The moment a new step-input form becomes supported without `_source_values` learning it, the boolean-source requirement silently stops applying to it — which is the precise defect this guard was added to fix. Letting the error propagate, or asserting the shape is reported elsewhere, keeps the guard's closure a property of the guard.

darkclad (commented):
Reviewed at `1d92af6` against merge base `97a6c44` (the tip of #398). The headline stats are the whole stack; the increment this PR adds on top of #398 is **13 files, +730/−32**, which is what I reviewed. I checked the branch out, built an environment, ran the suite, and probed the lowering directly rather than reading for the claims.

**`tests/core/nextflow`: 147 passed, 17 deselected.** mypy clean on `nf_types.py`, `utils_nf.py`, `input_output_nf.py`; pylint **9.89/10** across those three plus the whole nextflow test package, with no finding attributable to this branch. CI on `1d92af6`: `package` OK, `nextflow-runtime` OK — so R2.1 really did run `sort` under pinned Nextflow 26.04.2 for both flag values.

**The security claim holds under adversarial prefixes.** I rendered flags carrying `'; rm -rf /; '`, `$(id)`, `${x}`, `-x"y"`, a raw newline and a backslash. Every one lands inside a Groovy single-quoted string — where `$` is not interpolated — and then through the quoting helper:

```
'$(id)'  ->  ${reverse ? __sophios_shell_quote_9f72e('$(id)') : ''}
```

One argv word, no syntax escape, in either theme of hostility.

**The type boundaries are real, not documentary.** `NfTemplate` rejects an `NfFlag` segment, `NfCommand.__post_init__` rejects one in `stdin`/`stdout`/`stderr`, and `NfPort` rejects one in `glob` (`port glob must be an NfTemplate or None`). The `val`-port check bites at construction.

**Promotion still fails closed with the new token.** I wrote the flag artifacts, edited only `workflow.nf` to widen the prefix from `'-r'` to `'-r -o /tmp/x'`, and re-promoted: `generated Nextflow source does not match its executable IR artifact`. And the v2→v3 subset claim is free rather than asserted — `_render_command_token` delegates verbatim to `_render_template` for every `NfTemplate`, so v2-era artifacts render byte-identically.

**The `raw_source is None` skip in the flag-source guard is safe — I checked why.** An unwired flag input rejects at connection time (`process input FLAGS.verbose is not connected`). One carrying a tool-level `default` bypasses `_flag_source_findings` entirely, but the default is type-checked against the declared `boolean` independently: `default: "yes"`, `"false"`, `0`, `1`, `""` all reject. The gap is covered — though nothing in the tests says so.

Both adversarial-review fixes are genuinely closed. Four findings inline — **the first two are worth a commit before merge**, both cwltool divergences in a PR whose case rests on matching cwltool, and the first has a verified one-line fix that keeps the suite green. Three more that have no line in the diff to hang on:

- **`NfFlag` and `NfCommandToken` are not exported from `sophios.api.python.nextflow`.** The description's last line says they are: *"The new token types are also exported from `sophios.api.python.nextflow` beside the segment types they sit next to."* `git diff 97a6c44...1d92af6 -- src/sophios/api/` is empty, and `__all__` still lists `NfLiteral`, `NfTemplate`, `NfCommand` with no `NfFlag` — so a caller on the public surface can hold an `NfCommand` whose tokens it cannot name or construct. Nothing catches it: `test_concrete_public_module_exports_importer` spot-checks four reader names, and unlike `tests/core/test_tool_builder.py:42` there is no `__all__` assertion for this module.
- **The description says `schema_version: 4`; the code writes `3`.** *"bumps the executable IR to `schema_version: 4`... accepts earlier versions (2 and 3)."* It is `SCHEMA_VERSION = 3`, `SUPPORTED_SCHEMA_VERSIONS = {2, 3}`, and it was never 4 in any commit on the branch. The code is right and the tests pin it; only the prose is off by one — but it is the paragraph a reviewer reads to reason about compatibility.
- **The flag-only `true` fallback, which I am not filing.** `_command` now inserts `true` when the argv is all flags, so a flag-only tool renders `true --verbose` and exits zero where cwltool would try to exec `--verbose` and fail. It is the same divergence the pre-existing empty-command fallback already had, and the empty script it replaced exits zero too. Rejecting a program-less command outright would beat both, and that is a separate decision from this story.

src/sophios/utils_nf.py
vjaganat90: Fixed both sites to `if not separate:` / `if not binding.get("separate", True):` — matches cwltool's `not sep` exactly, including `separate: null`.

src/sophios/utils_nf.py
vjaganat90: Tightened to `prefix.strip()` so a blank/whitespace prefix falls into the named, aggregating diagnostic instead of NfFlag's raw invariant. Left NfFlag itself alone — it's pinned by `test_flag_requires_a_non_empty_prefix`.

docs/nextflow.md
vjaganat90: Dropped the "non-optional" qualifier — fixed.

src/sophios/utils_nf.py
vjaganat90: Now only exempts the two shapes you verified (`default` present, or `source` present with extra fields); anything else re-raises instead of silently continuing.

src/sophios/utils_nf.py
```diff
@@ -977,6 +998,106 @@ def _absent_optional_findings(
...
+            except ValueError:
+                if isinstance(raw_source, Mapping) and ("default" in raw_source or "source" in raw_source):
+                    # {"default": ...} is rejected by the absent-optional
+                    # findings pass; {"source": ..., <extra>} is rejected by
+                    # _source_values itself during lowering. Neither needs a
+                    # boolean-source finding on top. Any other shape that
+                    # reaches here is unrecognized and must not be silently
+                    # exempted from the boolean-source requirement.
+                    continue
+                raise
```
darkclad: **🟡 This exemption is narrower than it needs to be, and the `raise` below trades a named, aggregating diagnostic for a bare one.**

Requiring `"default"` or `"source"` in the mapping leaves out every *other* mapping shape — `{"valueFrom": ...}` and `{"linkMerge": ...}` with no `source` — which then hits the `raise`. Those shapes were already rejected, and rejected *better*, before this commit:

```
in = {"valueFrom": "x"}
  before: Nextflow Phase 1 capability analysis failed:
          - steps[0].in.verbose.valueFrom: valueFrom is not consumed by Nextflow Phase 1 lowering
  now:    step input 0.verbose must define source

in = {"linkMerge": "m"}
  before: - steps[0].in.verbose.linkMerge: linkMerge is not consumed by Nextflow Phase 1 lowering
  now:    step input 0.verbose must define source
```

Three things got worse, and they are the same three the blank-prefix thread was about. The message drops the `steps[0].in.verbose.<field>` path for a bare step *index*, which is not the vocabulary any other diagnostic here uses. It no longer aggregates. And an independent finding is now lost outright:

```
in = {"valueFrom": "x"} + ShellCommandRequirement
  before: - steps[0].run.requirements.ShellCommandRequirement: shell-mode command lowering is deferred to Phase 2
          - steps[0].in.verbose.valueFrom: valueFrom is not consumed by Nextflow Phase 1 lowering
  now:    step input 0.verbose must define source        <- the requirement finding is gone
```

Exempting any `Mapping` fixes it — mappings are exactly the shapes the unconsumed-fields pass already reports by path, and non-mappings are the genuinely unrecognized ones this guard was added to catch:

```python
if isinstance(raw_source, Mapping):
    continue
raise
```

I ran that: all four mapping diagnostics come back verbatim (aggregation included), `in = []` and `in = 42` still fail closed, and 148 pass. The comment then needs one less clause too.

src/sophios/utils_nf.py
```diff
@@ -224,6 +226,9 @@ def _position(value: Any, *, default: int) -> int:
 def _binding_tokens(prefix: Any, value: NfTemplate, *, separate: Any = True) -> tuple[NfTemplate, ...]:
     match prefix:
         case None:
+            if not separate:
```
darkclad, line 229: **🟡 Nothing pins this fix — or either of the other two in `97d88dc`.**

I reverted each of the three changes in this commit separately at `6d8578b` and ran the suite:

```
separate     revert (this line + line 280) -> 148 passed
prefix       revert (line 283)             -> 148 passed
sourceraise  revert (line 1081)            -> 148 passed
```

All three regress in silence. `test_rejects_separate_without_a_prefix` only pins `separate: False`, which both the old and new expressions reject — the cases that separate them are `separate: true` (cwltool accepts; the old boolean path rejected) and `separate: null` (cwltool rejects; the old `_binding_tokens` accepted), and neither is exercised.

This is the round that dropped the discipline the PR set for itself — *"Sequence in this PR: design → tests → implementation"*, and the previous round landed Red-only commits that failed with `DID NOT RAISE` first. Three behavior changes went in with no test at all, while the fourth fix (`6d8578b`, the export) got one that does bite — I checked by stripping the four lines back out, and it is the only failure.

The three missing cases, each of which fails before its fix and passes after:

- `separate: true` with no prefix on a `boolean` is **accepted** and contributes no token; `separate: null` is **rejected** on both the boolean and the string path.
- a whitespace-only prefix on a boolean binding reports `CWL command prefix for 'verbose' must be a non-empty string`, not `flag prefix must be a non-empty string`. Note `test_flag_requires_a_non_empty_prefix` cannot cover this — it exercises `NfFlag`'s invariant directly, which is the path this fix routes *around*.
- a non-mapping step input (`[]`, `42`) still raises rather than silently skipping the boolean-source requirement.

tests/core/nextflow/test_executable_ir.py
```diff
@@ -149,6 +151,93 @@ def test_rejects_qualifiers_without_an_approved_lowering(qualifier: str) -> None
...
+@pytest.mark.parametrize("prefix", ["", "   "])
+def test_flag_requires_a_non_empty_prefix(prefix: str) -> None:
+    with pytest.raises(ValueError, match="prefix"):
+        NfFlag("reverse", prefix)
+
+
+@pytest.mark.fast
+def test_public_module_exports_flag_and_command_token() -> None:
+    from sophios.api.python import nextflow
+    from sophios.nf_types import NfCommandToken
```
darkclad: **⚪ Two new `C0415`s, and the local import is avoidable here.**

`NfFlag` is already imported at the top of this file (line 15), so only `NfCommandToken` needs adding to that block — both lines can move up. The base file has zero `C0415`; these are the first two.

I realize this mirrors `test_reader.py:367`, as the commit message says. But that one is testing *importability* of the reader surface from a module the test file otherwise does not import, which is a reason for the local import. Here the module under test is already a top-level import away, so the pattern is copied without the reason. PR #399 treated `C0415` in tests as worth lifting; same call applies.

tests/core/nextflow/test_executable_ir.py
```diff
@@ -149,6 +151,93 @@ def test_rejects_qualifiers_without_an_approved_lowering(qualifier: str) -> None
...
+
+    assert nextflow.NfFlag is NfFlag
+    assert nextflow.NfCommandToken is NfCommandToken
+    assert "NfFlag" in nextflow.__all__
+    assert "NfCommandToken" in nextflow.__all__
+
+
+@pytest.mark.fast
+def test_flag_must_reference_a_declared_value_input() -> None:
+    command = NfCommand((NfTemplate((NfLiteral("sort"),)), NfFlag("reverse", "-r")))
```
darkclad: **⚪ `W0621`: this `command` shadows the `command` helper imported from `testkit` at line 25.**

Not from this round — it came in with the flag tests earlier in the PR, and I missed it in round 1. Zero `W0621` in this file on `46166dc`. `flag_command` or `tokens` reads fine and keeps the helper reachable in the rest of the function.

darkclad (commented):
Round 2 — reviewed at `6d8578b` against merge base `46166dc`.

**The rebase is clean and every round 1 finding is closed.** 8 own commits replayed onto the rebased base, the 9 duplicated #398 commits dropped, 0 base commits missing — `CLEAN` / `MERGEABLE`, CI green. I verified the fixes by running, not by reading the replies:

| # | Round 1 finding | Status |
|---|---|---|
| 🟡 | `separate: true` rejects on bool, accepted on string | **Fixed** — `97d88dc` |
| 🟡 | Blank prefix escapes as an unmapped IR error | **Fixed** — `97d88dc` |
| ⚪ | `_flag_source_findings` fail-open | **Fixed, with a regression** — see below |
| ⚪ | `docs/nextflow.md` "non-optional" | **Fixed** — `3e5a98a` |
| ⚪ | `NfFlag` not exported from the public API | **Fixed** — `6d8578b`, pinned |
| ⚪ | Description said `schema_version: 4` | **Fixed** — now reads `3` |

The `separate` table now matches cwltool exactly on both paths — `true` accepted, `false` and `null` rejected, for `boolean` and `string` alike. A whitespace prefix now yields the named `CWL command prefix for 'verbose' must be a non-empty string`. mypy clean; **148 passed**, up from 147.

One correction to a reply: the blank-prefix fix routes to the *named* diagnostic, but not an aggregating one — a lone blank prefix still raises bare, outside the `Nextflow Phase 1 capability analysis failed:` envelope, because `_input_binding_items` runs after `_raise_capability_findings`. The substantive complaint is fixed; only the "aggregating" half of the reply overstates it. That same distinction is what the first finding below turns on.

Two findings this round, both inline: **the `_flag_source_findings` fix costs three mapping shapes their named, aggregating diagnostic** (one-word fix, verified), and **none of the three behavior fixes in `97d88dc` is pinned by a test** — I reverted each one separately and the suite stayed green at 148. Plus two lint nits, one of which is mine from round 1.

Everything else looks merge-ready.

src/sophios/utils_nf.py
vjaganat90: Widened to any `Mapping` exactly as you verified — all four mapping diagnostics come back with aggregation, non-mappings still fail closed. One less clause in the comment too.

src/sophios/utils_nf.py
vjaganat90: Fair — added the three missing cases (separate: true accepted on boolean, separate: null rejected on both boolean and string paths, whitespace-prefix message, and the non-mapping re-raise). Revert-checked one myself to confirm they actually catch it.

tests/core/nextflow/test_executable_ir.py
vjaganat90: Moved `NfCommandToken` to the top-level import; left `nextflow` local per your own distinction — it's checking the module object's `__all__`, same reason as `test_reader.py`.

tests/core/nextflow/test_executable_ir.py
vjaganat90: Renamed to `flag_command`. Fixed.

darkclad (approved):

PR #401, opened 2026-09-02
feat(nextflow): lower $(inputs.x.basename) references

vjaganat90:
Stacked on #400 (`nf201_boolean_flags`). Review after that branch; its commits appear here until it merges.

## What this adds

`$(inputs.<name>.basename)` is lowered instead of rejected, so a tool can name its outputs after a staged input.

- The expression becomes an `NfBasenameReference` segment, valid in command tokens, stream targets, and output globs.
- Rendering uses the staged path's name property: `name.name.toString()` inside quoted command expressions, `${name.name}` inside glob GStrings.
- Phase 1 conflated `$(inputs.x.path)` and `$(inputs.x.basename)` into one reference before rejecting the latter outright; the two are now distinct typed segments.

## Boundaries

- A basename segment requires a `path` port and is rejected against `val` ports at model construction — the lowering can never claim a basename for a value channel.
- Everything else in the expression grammar is unchanged: any residual `$` still fails as an unsupported CWL expression.

## Schema

Adds `schema_version: 4` for the new segment kind, with hydration accepting 2 and 3 under the same evolution policy as #400.

## Design authority

`design:` commit `d798a22` approves this, including why it's sound: Nextflow stages an input under its original file name, so the staged name is exactly the CWL `basename`. That rationale was undocumented in Phase 1 and is now part of the locked design.

## Review fixes

- Basename artifacts now round-trip through parse and promote — a basename glob renders as a GString, and promotion is byte-identity bound, so this was an untested gap the story had declared as acceptance.
- The `val`-port rejection is now checked in every position the design claims (command, stream target, output glob), not just command position.
- `basename` registers as the kind introduced in schema version 4, so a payload declaring an older version while carrying it is rejected — previously the version integer gated nothing.
- `NfBasenameReference` joins `NfFlag` and `NfCommandToken` in the public API.

### DIFF
--- a/docs/nextflow.md
+++ b/docs/nextflow.md
@@ -84,6 +84,23 @@ exactly one argument and `false` contributes nothing. A boolean binding
 without a prefix contributes nothing for either value, matching CWL. Absent
 optional booleans still reject until option lowering is approved.
 
+## Basename references
+
+`$(inputs.<name>.basename)` is supported in command arguments, stream targets,
+and output globs, so a tool can name its outputs after a staged input:
+
+```yaml
+outputs:
+  result:
+    type: File
+    outputBinding:
+      glob: $(inputs.source.basename).copy
+```
+
+The reference must target a `File` or `Directory` input. Nextflow stages an
+input under its original file name, so the staged name is exactly the CWL
+`basename`.
+
 ## Current limits
 
 - Workflows must be flat and use `CommandLineTool`-equivalent processes.
--- a/docs/nextflow.md
+++ b/docs/nextflow.md
@@ -101,6 +101,17 @@ The reference must target a `File` or `Directory` input. Nextflow stages an
 input under its original file name, so the staged name is exactly the CWL
 `basename`.
 
+The `.copy` suffix above is load-bearing: a basename glob must derive a **new**
+name. A Nextflow output declaration does not capture staged inputs, so a glob
+naming only the input itself compiles cleanly and then fails at run time with
+`Missing output file(s)`.
+
+Two further limits apply to a rendered glob. It stays a Nextflow glob pattern,
+so a staged name containing `*`, `?`, `[`, `{` or `}` is interpreted as a
+pattern rather than matched literally, and the process fails to find the file
+it wrote. Rendering such a name literally is a separate change, because it
+alters emitted bytes and the generated-subset reader with them.
+
 ## Current limits
 
 - Workflows must be flat and use `CommandLineTool`-equivalent processes.
--- a/src/sophios/input_output_nf.py
+++ b/src/sophios/input_output_nf.py
@@ -9,6 +9,7 @@ from typing import Any
 from .nf_types import (
     ExecutableNextflowWorkflow,
     NF_SHELL_QUOTE_HELPER,
+    NfBasenameReference,
     NfConnection,
     NfFlag,
     NfInputReference,
@@ -58,13 +59,18 @@ def _groovy_gstring_fragment(value: str) -> str:
     return value.translate(_GSTRING_FRAGMENT_TABLE)
 
 
+def _segment_expression(segment: Any) -> str:
+    match segment:
+        case NfInputReference():
+            return f"{segment.name}.toString()"
+        case NfBasenameReference():
+            return f"{segment.name}.name.toString()"
+        case _:
+            return _groovy_literal(segment.value)
+
+
 def _template_expression(template: Any) -> str:
-    return " + ".join(
-        f"{segment.name}.toString()"
-        if isinstance(segment, NfInputReference)
-        else _groovy_literal(segment.value)
-        for segment in template.segments
-    )
+    return " + ".join(_segment_expression(segment) for segment in template.segments)
 
 
 def _shell_quote(value: str) -> str:
@@ -90,10 +96,13 @@ def _render_glob(template: Any) -> str:
         return _groovy_literal("".join(segment.value for segment in template.segments))
     rendered: list[str] = []
     for segment in template.segments:
-        if isinstance(segment, NfInputReference):
-            rendered.append(f"${{{segment.name}}}")
-        else:
-            rendered.append(_groovy_gstring_fragment(segment.value))
+        match segment:
+            case NfInputReference():
+                rendered.append(f"${{{segment.name}}}")
+            case NfBasenameReference():
+                rendered.append(f"${{{segment.name}.name}}")
+            case _:
+                rendered.append(_groovy_gstring_fragment(segment.value))
     return f'"{"".join(rendered)}"'
 
 
--- a/src/sophios/nf_types.py
+++ b/src/sophios/nf_types.py
@@ -152,7 +152,30 @@ class NfInputReference:
         return {"kind": "input", "name": self.name}
 
 
-NfTemplateSegment = NfLiteral | NfInputReference
+@dataclass(frozen=True, slots=True)
+class NfBasenameReference:
+    """Reference to the staged file name of one path input.
+
+    Nextflow stages an input under its original file name, so the staged
+    path's name is exactly the CWL ``basename``.
+    """
+
+    name: str
+
+    def __post_init__(self) -> None:
+        _validate_ir_identifier(self.name, field_name="template basename reference")
+
+    def to_dict(self) -> dict[str, str]:
+        """Return a JSON-compatible representation.
+
+        Returns:
+            dict[str, str]: The segment as ``{"kind": "basename", "name": ...}``.
+        """
+        return {"kind": "basename", "name": self.name}
+
+
+NfTemplateSegment = NfLiteral | NfInputReference | NfBasenameReference
+NfReferenceSegment = NfInputReference | NfBasenameReference
 
 
 def _segment_from_dict(value: Mapping[str, Any]) -> NfTemplateSegment:
@@ -164,6 +187,9 @@ def _segment_from_dict(value: Mapping[str, Any]) -> NfTemplateSegment:
         case "input":
             _check_fields(item, type_name="NfInputReference", required={"kind", "name"})
             return NfInputReference(item["name"])
+        case "basename":
+            _check_fields(item, type_name="NfBasenameReference", required={"kind", "name"})
+            return NfBasenameReference(item["name"])
         case kind:
             raise ValueError(f"unsupported template segment kind {kind!r}")
 
@@ -177,7 +203,7 @@ class NfTemplate:
     def __post_init__(self) -> None:
         canonical: list[NfTemplateSegment] = []
         for segment in self.segments:
-            if not isinstance(segment, (NfLiteral, NfInputReference)):
+            if not isinstance(segment, (NfLiteral, NfInputReference, NfBasenameReference)):
                 raise TypeError("template segments must be typed literal or input references")
             if isinstance(segment, NfLiteral) and canonical and isinstance(canonical[-1], NfLiteral):
                 canonical[-1] = NfLiteral(canonical[-1].value + segment.value)
@@ -487,20 +513,26 @@ class NfProcess:
             *(template for template in (self.command.stdin, self.command.stdout, self.command.stderr) if template),
             *(port.glob for port in outputs if port.glob),
         ]
-        references = flag_names | {
-            segment.name
-            for template in templates
-            for segment in template.segments
-            if isinstance(segment, NfInputReference)
+        segments = [segment for template in templates for segment in template.segments]
+        basename_names = {
+            segment.name for segment in segments if isinstance(segment, NfBasenameReference)
+        }
+        references = flag_names | basename_names | {
+            segment.name for segment in segments if isinstance(segment, NfInputReference)
         }
         if unknown := references - input_names:
             raise ValueError(
                 f"process {self.name!r} templates reference unknown inputs: {', '.join(sorted(unknown))}"
             )
-        value_names = {port.name for port in inputs if port.qualifier == "val"}
-        if invalid := flag_names - value_names:
+        qualifiers = {port.name: port.qualifier for port in inputs}
+        if invalid := {name for name in flag_names if qualifiers[name] != "val"}:
             raise ValueError(
                 f"process {self.name!r} flag tokens must reference val inputs: "
+                f"{', '.join(sorted(invalid))}"
+            )
+        if invalid := {name for name in basename_names if qualifiers[name] != "path"}:
+            raise ValueError(
+                f"process {self.name!r} basename references must target path inputs: "
                 f"{', '.join(sorted(invalid))}"
             )
         match self.container:
@@ -743,13 +775,15 @@ def _connection_from_dict(value: Mapping[str, Any]) -> NfConnection:
 class ExecutableNextflowWorkflow:
     """Closed, immutable, versioned executable representation of a DSL2 workflow."""
 
-    SCHEMA_VERSION: ClassVar[int] = 3
+    SCHEMA_VERSION: ClassVar[int] = 4
     # Earlier versions whose value space is a strict subset of the current
     # model hydrate unchanged; serialization always writes SCHEMA_VERSION.
-    SUPPORTED_SCHEMA_VERSIONS: ClassVar[frozenset[int]] = frozenset({2, 3})
+    SUPPORTED_SCHEMA_VERSIONS: ClassVar[frozenset[int]] = frozenset({2, 3, 4})
     # Each additive token or segment kind declares the version that
     # introduced it, so the subset property is enforced rather than assumed.
-    KIND_SCHEMA_VERSIONS: ClassVar[Mapping[str, int]] = MappingProxyType({"flag": 3})
+    KIND_SCHEMA_VERSIONS: ClassVar[Mapping[str, int]] = MappingProxyType(
+        {"flag": 3, "basename": 4}
+    )
     REPRESENTATION_KIND: ClassVar[str] = "executable"
 
     name: str
--- a/src/sophios/nf_types.py
+++ b/src/sophios/nf_types.py
@@ -6,7 +6,7 @@ from graphlib import CycleError, TopologicalSorter
 import json
 import math
 from types import MappingProxyType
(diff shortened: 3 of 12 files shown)


darkclad (commented):
The stack is clean: #400 has merged, so `git merge-base pr-401 origin/feat/nf-support` is `ee8af93` and this diff is NF-202 alone — 12 files, +284/−45. The description's "commits appear here until it merges" caveat is stale in the reader's favour.

I could not run the full suite here (no `cwl_utils`/`cwltool`, so the nextflow `conftest.py` fails to import), so rather than read for the claims I drove `nf_types`, `input_output_nf`, `utils_nf` and `nf_reader` directly, and ran `test_executable_ir.py` (31/31) and `test_renderer.py` (11 passed, 1 error on the fixture needing `cwl_utils`) in isolation.

**The lowering is sound and the rationale is the right one.** The renderer emits `path source`, Nextflow stages under the original file name, so `source.name` is exactly the CWL `basename`. What comes out:

```groovy
    output:
    path "${source.name}.copy", emit: result

    script:
    """
    ${q('cp')} ${q(source.toString())} ${q(source.name.toString() + '.copy')} > ${q(source.name.toString() + '.log')}
    """
```

**All three review fixes hold, and I checked each rather than taking them.**

- *Provenance.* `write_nextflow_artifacts` → `parse_nf_file` → `promote_nextflow_document`, on a workflow carrying a basename glob **and** a basename `stdout`, returns the identical object with `opaque_regions == ()`. That was the untested acceptance gap and it is now closed.
- *Every position.* The `val` rejection fires in command, `stdout`, **`stdin`** and glob position. The design claims three; the implementation covers four.
- *Schema 4 gates something.* A payload carrying `basename` while declaring version 2 or 3 is rejected by name; `flag` still gates at 3; a plain payload still hydrates from 2, 3 and 4; 5 rejects. The version integer is load-bearing now.

**The grammar did not widen.** `.nameroot`, `.dirname`, `.size`, `.contents`, `.BASENAME`, `.basename.txt`, an unclosed `$(inputs.x.basename`, and `$(runtime.outdir)/…` all still fail as unsupported expressions. Whitespace forms and mixed templates lower correctly.

**The `qualifiers[name]` refactor is safe.** Swapping set arithmetic for a dict lookup would `KeyError` on an unknown name, but `references - input_names` runs first and `references` is a superset of both `flag_names` and `basename_names`, so every key is present. It is a real ordering dependency and nothing in the code says so.

Nothing blocking. One substantive item — the `val`-port rejection is the single plain CWL source error that reaches executable-model construction instead of capability analysis, contrary to §3 — plus one latent runtime hazard this feature is the first to make reachable, and three one-liners. All inline.

tests/core/nextflow/test_capabilities.py
```diff
@@ -512,27 +512,27 @@ def test_rejects_absent_optional_workflow_input_feeding_required_input() -> None
...
+            [step("BASENAME", **{"in": {"label": "label"}, "out": ["result"]})],
+            inputs={"label": {"type": "string"}},
             outputs={"result": {"type": "File", "outputSource": "BASENAME/result"}},
         ),
         [basename],
-        workflow_inputs={"source": {"class": "File", "path": "input.txt"}},
+        workflow_inputs={"label": "input"},
     )
-    with pytest.raises(ValueError, match="basename"):
+    with pytest.raises(ValueError, match="basename.*path"):
```
darkclad: **⚪ This asserts on an IR raise, not a capability finding.**

The message it matches comes from `NfProcess.__post_init__`, not from `_raise_capability_findings`. Sitting in `test_capabilities.py` is what makes the layering look covered when it isn't — nothing here would notice if the capability pass never grew a basename check.

Once that pass exists (see my note on `nf_types.py`), this should match the aggregated finding text *and* its source path. The IR-level raise is already pinned by `test_basename_segment_must_reference_a_path_input` in `test_executable_ir.py`, so nothing is lost by moving this one up a layer.

src/sophios/input_output_nf.py
```diff
@@ -90,10 +96,13 @@ def _render_glob(template: Any) -> str:
...
     for segment in template.segments:
-        if isinstance(segment, NfInputReference):
-            rendered.append(f"${{{segment.name}}}")
-        else:
-            rendered.append(_groovy_gstring_fragment(segment.value))
+        match segment:
+            case NfInputReference():
+                rendered.append(f"${{{segment.name}}}")
+            case NfBasenameReference():
+                rendered.append(f"${{{segment.name}.name}}")
```
darkclad: **🟡 The rendered glob is still glob-interpreted at runtime, and this is the change that makes that reachable.**

This emits `path "${source.name}.copy", emit: result`. Nextflow treats an output `path` name as a **glob pattern** by default, so the pattern's meaning depends on runtime data. A staged file named `a[1].txt` gives the pattern `a[1].txt.copy`, whose `[1]` is a character class: it does not match the file the process actually wrote, and the process fails with `Missing output file(s)`. `{`, `}`, `*` and `?` behave the same way.

The mechanism predates this PR — `NfInputReference` renders `${source}` on the line above and did so on `ee8af93` — but nothing before NF-202 made a reference-in-glob genuinely useful. Naming an output after an input's basename is the whole point of this story, and it is precisely the case where the compiler *knows* the name is meant literally.

The codebase already handles the same hazard on the other side of the boundary: line 243 passes `glob: false` to `Channel.fromPath`. The output declaration takes the same option.

A rule that stays decidable at compile time, so the renderer stays total: if the template has at least one reference segment **and** no `NfLiteral` part contains a glob metacharacter, the name is meant literally — render `glob: false`. If a literal part does carry one, the user asked for globbing; leave it alone.

I would not block NF-202 on this. It changes rendered bytes, so `test_flag_workflow_artifacts_are_byte_stable` and the reader's `_PORT` regex (`nf_reader.py:31`, which stops at the first `,`) have to move with it — that is its own review surface. But if it ships as a follow-up, §"Outputs" should stop claiming every output has a concrete capture mechanism without noting the limit.

src/sophios/nf_types.py
```diff
@@ -152,7 +152,30 @@ def to_dict(self) -> dict[str, str]:
...
+        """Return a JSON-compatible representation.
+
+        Returns:
+            dict[str, str]: The segment as ``{"kind": "basename", "name": ...}``.
+        """
+        return {"kind": "basename", "name": self.name}
+
+
+NfTemplateSegment = NfLiteral | NfInputReference | NfBasenameReference
+NfReferenceSegment = NfInputReference | NfBasenameReference
```
darkclad: **⚪ `NfReferenceSegment` is introduced dead.**

No reference in `src/`, `tests/`, `docs/` or `design_docs/`. Either use it — `_segment_expression` and `_render_glob` both switch on exactly this union, and the `basename_names`/`references` collection in `NfProcess.__post_init__` gathers exactly these two kinds — or drop it.

#400 closed with "drop an avoidable local import and a shadowed name"; a new unused alias goes the other way.

src/sophios/nf_types.py
```diff
@@ -177,7 +203,7 @@ class NfTemplate:
     def __post_init__(self) -> None:
         canonical: list[NfTemplateSegment] = []
         for segment in self.segments:
-            if not isinstance(segment, (NfLiteral, NfInputReference)):
+            if not isinstance(segment, (NfLiteral, NfInputReference, NfBasenameReference)):
```
darkclad: **⚪ The `TypeError` on the next line no longer names what it accepts.**

"template segments must be typed literal or input references" — a basename reference is now valid too, so someone who passes an `NfFlag` here gets a message that omits a third of the answer.

The tuple also restates `NfTemplateSegment` (line 177) a second time, so the next segment kind has to land in both places. `isinstance(segment, get_args(NfTemplateSegment))` keeps them in sync and lets the message be built from the union rather than hand-maintained beside it.

docs/nextflow.md
```diff
@@ -84,6 +84,23 @@ exactly one argument and `false` contributes nothing. A boolean binding
...
+outputs:
+  result:
+    type: File
+    outputBinding:
+      glob: $(inputs.source.basename).copy
+```
+
+The reference must target a `File` or `Directory` input. Nextflow stages an
+input under its original file name, so the staged name is exactly the CWL
+`basename`.
```
darkclad: **⚪ Two things this paragraph should say.**

**The `.copy` suffix in the example above is load-bearing.** Nextflow's output `path` defaults to `includeInputs: false`, so a declaration matching *only* a staged input captures nothing — a bare `glob: $(inputs.source.basename)` compiles cleanly and then fails at runtime with `Missing output file(s)`. This is mostly defused already, since `InitialWorkDirRequirement` is deferred with a clean diagnostic (`utils_nf.py:460`) and that is the CWL idiom which produces the shape — but the design paragraph says basename references are "valid in every template position: command tokens, stream targets, and output globs" with no qualification. One sentence here would close it: a basename glob must derive a **new** name.

**`Directory` is claimed but unpinned.** It does work — `NfPort("s", "path", path_kind="directory")` accepts a basename reference, I checked — but no test covers it, so nothing holds the `path_kind` half of this sentence in place.

darkclad (commented):
1

src/sophios/nf_types.py
```diff
@@ -487,22 +513,28 @@ def __post_init__(self) -> None:
...
+        if invalid := {name for name in flag_names if qualifiers[name] != "val"}:
             raise ValueError(
                 f"process {self.name!r} flag tokens must reference val inputs: "
                 f"{', '.join(sorted(invalid))}"
             )
+        if invalid := {name for name in basename_names if qualifiers[name] != "path"}:
+            raise ValueError(
+                f"process {self.name!r} basename references must target path inputs: "
+                f"{', '.join(sorted(invalid))}"
+            )
```
darkclad, line 537: **🟡 This is the one plain CWL source error that escapes capability analysis.**

`cwl_rosetree_to_nextflow` says it runs "closed-world capability analysis first; every unsupported source semantic is aggregated and rejected before any lowering happens", and §3 says a rejected capability "produces a structured diagnostic **before executable-model construction**", aggregating "so users do not have to fix unsupported constructs one at a time".

`$(inputs.label.basename)` against `label: string` arrives here instead:

```
process 'BASENAME' basename references must target path inputs: label
```

Normalized process and port, no source path, no CWL field — and it short-circuits, so a document with three of these reports one.

The asymmetry with #400 is sharper than it looks. The flag/`val` rule four lines up can never fire from CWL lowering, because `cwl_type_to_nf_qualifier` maps every boolean to `val`; it guards hand-built IR only, which is exactly why #400 *also* added `_flag_source_findings` (`utils_nf.py:1046`) to catch the source-level shape by path. This rule **does** fire from CWL lowering — `test_rejects_basename_against_a_value_input` is that path — and it has no capability-pass counterpart.

Phase 1's diagnostic for the same document was better on this one axis, because it at least named the field:

```
CWL output glob for 'result' uses $(inputs.<name>.basename); basename lowering is deferred to Phase 2
```

Keep this guard — it is the right backstop for hand-built IR. But mirror `_flag_source_findings` in `_tool_capability_findings`: for each input reached by a basename segment in any template position, require `cwl_type_to_nf_qualifier(...) == "path"` and emit a finding tagged with the real path (`steps[0].run.outputs.result.outputBinding.glob`).

darkclad (commented):
Six inline findings at 03eebb9. Two are worth acting on: the val-port rejection escaping capability analysis (nf_types.py), and the runtime-glob hazard this story is the first to make reachable (input_output_nf.py). The other four are one-liners.

tests/core/nextflow/test_capabilities.py
darkclad: **⚪ This asserts on an IR raise, not a capability finding.**

The message it matches comes from `NfProcess.__post_init__`, not from `_raise_capability_findings`. Sitting in `test_capabilities.py` is what makes the layering look covered when it isn't — nothing here would notice if the capability pass never grew a basename check.

Once that pass exists (see my note on `nf_types.py`), this should match the aggregated finding text *and* its source path. The IR-level raise is already pinned by `test_basename_segment_must_reference_a_path_input` in `test_executable_ir.py`, so nothing is lost by moving this one up a layer.

src/sophios/input_output_nf.py
darkclad, line 103: **🟡 The rendered glob is still glob-interpreted at runtime, and this is the change that makes that reachable.**

This emits `path "${source.name}.copy", emit: result`. Nextflow treats an output `path` name as a **glob pattern** by default, so the pattern's meaning depends on runtime data. A staged file named `a[1].txt` gives the pattern `a[1].txt.copy`, whose `[1]` is a character class: it does not match the file the process actually wrote, and the process fails with `Missing output file(s)`. `{`, `}`, `*` and `?` behave the same way.

The mechanism predates this PR — `NfInputReference` renders `${source}` on the line above and did so on `ee8af93` — but nothing before NF-202 made a reference-in-glob genuinely useful. Naming an output after an input's basename is the whole point of this story, and it is precisely the case where the compiler *knows* the name is meant literally.

The codebase already handles the same hazard on the other side of the boundary: line 243 passes `glob: false` to `Channel.fromPath`. The output declaration takes the same option.

A rule that stays decidable at compile time, so the renderer stays total: if the template has at least one reference segment **and** no `NfLiteral` part contains a glob metacharacter, the name is meant literally — render `glob: false`. If a literal part does carry one, the user asked for globbing; leave it alone.

I would not block NF-202 on this. It changes rendered bytes, so `test_flag_workflow_artifacts_are_byte_stable` and the reader's `_PORT` regex (`nf_reader.py:31`, which stops at the first `,`) have to move with it — that is its own review surface. But if it ships as a follow-up, §"Outputs" should stop claiming every output has a concrete capture mechanism without noting the limit.

src/sophios/nf_types.py
darkclad: **⚪ `NfReferenceSegment` is introduced dead.**

No reference in `src/`, `tests/`, `docs/` or `design_docs/`. Either use it — `_segment_expression` and `_render_glob` both switch on exactly this union, and the `basename_names`/`references` collection in `NfProcess.__post_init__` gathers exactly these two kinds — or drop it.

#400 closed with "drop an avoidable local import and a shadowed name"; a new unused alias goes the other way.

src/sophios/nf_types.py
darkclad: **⚪ The `TypeError` on the next line no longer names what it accepts.**

"template segments must be typed literal or input references" — a basename reference is now valid too, so someone who passes an `NfFlag` here gets a message that omits a third of the answer.

The tuple also restates `NfTemplateSegment` (line 177) a second time, so the next segment kind has to land in both places. `isinstance(segment, get_args(NfTemplateSegment))` keeps them in sync and lets the message be built from the union rather than hand-maintained beside it.

docs/nextflow.md
darkclad, line 102: **⚪ Two things this paragraph should say.**

**The `.copy` suffix in the example above is load-bearing.** Nextflow's output `path` defaults to `includeInputs: false`, so a declaration matching *only* a staged input captures nothing — a bare `glob: $(inputs.source.basename)` compiles cleanly and then fails at runtime with `Missing output file(s)`. This is mostly defused already, since `InitialWorkDirRequirement` is deferred with a clean diagnostic (`utils_nf.py:460`) and that is the CWL idiom which produces the shape — but the design paragraph says basename references are "valid in every template position: command tokens, stream targets, and output globs" with no qualification. One sentence here would close it: a basename glob must derive a **new** name.

**`Directory` is claimed but unpinned.** It does work — `NfPort("s", "path", path_kind="directory")` accepts a basename reference, I checked — but no test covers it, so nothing holds the `path_kind` half of this sentence in place.

src/sophios/nf_types.py
vjaganat90: Fixed. `_basename_source_findings` in the capability pass now requires a path qualifier for every input a basename segment reaches — across `baseCommand`, `arguments`, input bindings, the three streams and output globs — each reported at its CWL path. Three offenders now report three findings instead of one. Kept the `NfProcess` raise as the hand-built-IR backstop, as you suggested.

tests/core/nextflow/test_capabilities.py
vjaganat90: Moved up a layer. It now asserts the aggregated finding text *and* its path, via a small `_findings` helper that splits the diagnostic into its individual findings — so the assertion pins count, text and path rather than a substring. Added a second case with two offenders in different positions to pin the aggregation itself. The IR raise stays pinned by `test_basename_segment_must_reference_a_path_input`.

src/sophios/input_output_nf.py
vjaganat90: Taking your read that this shouldn't block NF-202. The `glob: false` change moves rendered bytes, the reader's `_PORT` regex and the byte-stability goldens together, so it wants its own diff rather than riding along here.

I did take the condition you attached to it: `docs/nextflow.md` and §Outputs now state the limit instead of claiming capture unqualified — a rendered glob stays pattern-interpreted, so an assembled name carrying a metacharacter misses the file the process wrote.

Say the word if you'd rather the renderer change land before this merges and I'll do it next.

src/sophios/nf_types.py
vjaganat90: Dropped it. Both candidate sites `match` on the concrete types individually and render differently per branch, so there was no natural reader for the union.

src/sophios/nf_types.py
vjaganat90: Now `isinstance(segment, get_args(NfTemplateSegment))`, with the message built from the union — so the next segment kind lands in one place and the message names every kind it accepts. Its test asserts the full message rather than a fragment.

docs/nextflow.md
vjaganat90: Both added. The `.copy` point is explicit now: a basename output glob must derive a **new** name, since a declaration matching only a staged input captures nothing and fails at run time. And `Directory` is pinned — `test_basename_segment_accepts_both_path_kinds` covers both path kinds rather than leaving half the sentence unheld.

darkclad (approved):

PR #403, opened 2026-09-08
Spec 2: hermetic generators and the equivalence relation

vjaganat90:
Rebased onto `master` now that #402 has merged; this branch is the bottom of the Spec 2 stack.

Spec 2 builds the oracle the design requires before the Spec 3 IR migration: the compiler's invariants checked against generated workflows rather than twenty-two hand-written ones. This is the first three of nine tasks — the generator half, plus the artifact the rest is built on.

**A tool registry the suite owns** (`synthetic_tools.py`). `get_tools_cwl` globs directories named in a user config, so a property fed by it is a claim about the compiler *and* about which plugin repos the machine has checked out. Eight stubs with known signatures replace it, chosen to reach branches single-typed tools never do: a union input, a terminal tool, two output formats, one optional argument.

**A compile path with no environment** (`hermetic.py`), with hermeticity checked two ways (`test_hermeticity.py`): a static import-closure scan, and a subprocess run with plugin discovery poisoned and `HOME` redirected — `get_config` provisions `~/wic` when it is missing, so a suite that reads the config doesn't merely depend on the environment, it creates one.

**Generators derived from the AST** (`ast_strategies.py`). Documents are built as `sophios.lang` nodes and rendered to source rather than assembled as dicts, so the space the properties quantify over is the space the types admit, and a counterexample arrives as a `.wic` file. Adequacy is itself a property: fifteen construct kinds, derived from the reference's tables rather than from the strategy.

**The equivalence relation** (`equivalence.py`) — the artifact. Four properties here and all six of Spec 3's differential ones are `compile(f(W)) ≡ compile(W)`, differing only in `f` and in what `≡` may ignore. The guide calls Spec 3's comparison "semantic, not byte-level" and never says what semantic means; the two regression tests already in the tree answer it differently and nothing reconciles them. Three strengths as a lattice, each naming what it forgives and why — `run:` paths encode embedding, namespaces encode nesting depth — returning *where* two compilations diverge rather than that they do, since Spec 3 runs this over thousands of inputs.

**Worth a look:**

- Discrimination is tested before agreement. A relation returning "equivalent" unconditionally satisfies reflexivity and the lattice law, and makes every property in both Specs vacuous at once; five pairs no strength may accept lead that file.
- `UP_TO_RENAMING` compares the tool multiset, not just the shape. A bare `DiGraphMatcher` — what `test_inline_subworkflows` uses — accepts a graph whose every step was swapped for a different tool of the same arity, which is the kind of thing an IR migration gets wrong.
- Two deliberate re-implementations: `required_inputs_of` mirrors `_arg_has_default_or_is_optional` rather than importing it, and type compatibility is written from the CWL semantics rather than calling `types_match`. An oracle that asks the compiler what it requires and then checks the compiler honoured it is one implementation grading itself.
- The relation reads `outputSource` as an edge, not as a dropped name. Workflow outputs are nodes labelled with their declared shape, so a lowering that rewired every output to the wrong producer is a divergence — a sorted multiset of shapes cannot see a permutation of itself.
- It also refuses the surface forms the compiler does not emit — a mapping-form `steps:`, a step with no `id`, a non-mapping `in:` — rather than reading them as an empty document. The one verdict a comparison oracle must not invent is "no difference".
- `NOT_YET_COMPILABLE` is empty. It held one entry, and #402 removed that construct from the language; the table stays because `!cwl` is the next one.

**Findings, none fixed here** — Spec 2 touches the compiler only where a property cannot otherwise be stated, and both of these are scheduled with a `Code` each in a later task:

| | |
|---|---|
| **CE-12** | `_arg_has_default_or_is_optional` tests `get('default')` for *truthiness*, so `default: 0`, `False` or `''` reads as no default and the argument is classed required. Two `mm-workflows` adapters declare `convert_Kd_dG` as `boolean` with `default: False`. |
| **CE-14** | `populate_scalar_val` coerces with a naked `int()`/`float()`, so `in: {n: !ii _}` against a tool declaring `n: int` raises a bare `ValueError` — no code, no span, no file. |

Both are the class §4.2 of the design names: a key present with a falsy or ill-typed value reaching code that assumes otherwise.

These modules are validated by mutation, and each mutation that survived is named in the docstring of the test written in answer to it. The survivals are the useful half — a green suite says nothing until something is known about what turns it red.

**Not in this PR.** Tasks 4–9: the transformations and the one equivalence property, canonical emission, the single-compilation predicates, the canonical path, and the weekly benchmarks.


tests/core/equivalence.py
```diff
@@ -0,0 +1,446 @@
...
+    """The multiset of workflow-level port shapes under `key`, names dropped.
+
+    Verified against a hermetic compilation rather than assumed: the compiler
+    emits workflow-level `inputs`/`outputs` keys as `{step_id}___{port}` —
+    `oracle__step__1__mk_file___name` — so those keys are namespaced and are
+    *not* stable under renaming. This is the correction to the brief's
+    expectation that they would be. What renaming cannot change is how many
+    ports there are (a renaming is a bijection on names) and what type each
+    one has, so that is what is compared. `outputSource` is excluded for the
+    same reason the key is: it is a name.
```
darkclad: **🔴 `outputSource` is excluded by the argument this PR already refuted for `in[].source`.**

The reason given is "it is a name." But `_dataflow`'s own docstring (lines 310-320) is the refutation of exactly that reasoning on the input side:

> Dropping it made a whole class of rewiring invisible: [...] a document feeding `sink.n` from the `int` input and `mk_file.name` from the `string` one was UP_TO_RENAMING-equal to the document that fed each from the other.

`outputSource` is the mirror image, and it is still dropped. Take two documents identical except that workflow output `a` is fed by `step_1/f` and `b` by `step_2/n`, versus the swap:

* `_port_shapes` reduces both to a **sorted multiset** of shapes. A permutation of the pairing cannot change a sorted multiset, so this passes - and it passes *even when the two outputs have different types*, because `sorted(['{type: File}', '{type: int}'])` is the same list either way. The type-safety argument that makes this reduction defensible for `inputs` does not carry to `outputs`, because for outputs the name is the only thing tying a shape to a producer.
* `_requirement_names` is unaffected.
* `_same_dag` never sees it: `_dataflow` builds edges only from `_bindings(step)` (line 332), which reads `in:` and nothing else. Workflow outputs are not nodes.

Verdict: UP_TO_RENAMING-equivalent. Which means the relation cannot see which step feeds which workflow output - a compiler change that rewired every output to the wrong producer would be reported as equivalent, and P28/partition-independence would confirm it.

The fix is the one already written twenty lines up. Give each workflow-level output a node in `_dataflow`, labelled with its declared shape and never its name - the exact treatment `_INPUT_NODE` gets at lines 338-340 - and an edge from the producing step carrying the `(port, '')` label. A swap then changes the labelled graph and the isomorphism fails, while a genuine renaming still passes because only shapes are compared. `_port_shapes` can go on dropping the key; what is missing is the *edge*, not the name.

Worth adding the swap to `MUST_DIFFER` in the same commit - it is the fixture that would have caught this, and its absence is why the input-side fix did not generalise.

tests/core/ast_strategies.py
```diff
@@ -0,0 +1,459 @@
...
+    used: set[str] = set()
+    for _ in range(count):
+        step = draw(_step(defined_edges, referenced_inputs))
+        if as_mapping:
+            # A repeated key is `wic010`, not a document. Skip rather than
+            # filter: filtering a composite this deep hits filter_too_much.
+            if step.id in used:
+                continue
+            used.add(step.id)
+        steps.append(step)
```
darkclad: **🔴 A discarded step leaves its `!&` names in `defined_edges`, so a later step can reference an edge nothing defines.**

`_step` is drawn at line 253, *before* the duplicate-id check at 257. By then `_fresh_edge` (line 158) has already appended every `!&` name that step defines to `defined_edges`, which is document-scoped by design. The `continue` at 258 drops the step; it does not un-append the names.

Mapping form draws stems with replacement (`st.sampled_from(STEMS)` at line 186, `count` up to 4), so collisions are common, not exotic. After one, `defined_edges` contains names no surviving step defines, and the next `_step` can draw an `EdgeRef` from that list (line 200). The rendered document then carries `!*name` with no `!&name` anywhere in it.

**The failure is silent, not loud.** `compile_hermetic` hardcodes `testing=True` (tests/core/hermetic.py:48), and compiler.py:784 reads:

```python
if is_root and not testing:
    raise ValueError(f"Error! No definition found for !&{arg_val}!")
```

so under this suite it never raises - it falls through and adds a CWL input "for testing only." The generator believes it built a document with N steps and an internal edge; the compiler is handed a document with N-1 steps and a *workflow input* where that edge was. That document is what `compilable_documents()` yields to every Task 3-7 property, including the equivalence relation, whose whole subject is which port is fed from where.

It also lands in the one place that would have surfaced it: `test_the_workflows_strategy_produces_documents_the_compiler_accepts` catches bare `ValueError` at test_generators.py:151 and attributes it to "the declared `!ii` residual." Any non-`testing` caller's dangling-edge error would be swallowed there under the wrong name, and the measured "about one document in ten" residual is a mix of two causes.

Cheapest correct fix is to stop discarding steps at all - draw the mapping-form stems up front:

```python
stems = draw(st.lists(st.sampled_from(STEMS), unique=True, min_size=1, max_size=4))
```

That also removes the silent shrinking that makes `count` not mean what it says: today a mapping-form document asking for 4 steps routinely gets 2, so the distribution the properties quantify over is narrower than it reads. Truncating `defined_edges` back to a snapshot length on the `continue` would close the correctness half alone, but leaves the distribution problem.

tests/core/equivalence.py
```diff
@@ -0,0 +1,446 @@
...
+    the document declares for that input, never its name, which is what keeps
+    the forgiveness honest in both directions: differently-typed inputs cannot
+    be swapped silently, and swapping two identically-shaped ones stays
+    equivalent, because that swap is exactly what re-rooting a namespace does.
+    """
+    graph = nx.DiGraph()
+    shapes = _declared_shapes(document, 'inputs')
+    steps = document.get('steps')
+    steps = steps if isinstance(steps, list) else []
+    declared = [s for s in steps if isinstance(s, dict) and 'id' in s]
```
darkclad: **🟡 Array-form `steps:` and `in:` are compared nowhere, and this module reads both forms everywhere else.**

`_port_shapes` (363), `_declared_shapes` (383-387) and `_requirement_names` (400-403) each read the mapping form *and* the array form, with an argument for why: the compiler emits one, CWL admits the other. `steps` and `in` are the two places that argument is not applied, and they are the two that carry the dataflow.

* Line 325: a `steps:` that is not a list becomes `[]`. CWL admits the mapping form (`steps: {step_1: {...}}`), and two documents differing *only* in their steps then produce two empty graphs and compare as equivalent - every check that would distinguish them operates on `declared`, which is empty.
* `in:` is worse because it is dropped twice. `_FORGIVEN_STEP_KEYS` (260) forgives it unconditionally, and `_step_body` re-adds it only `if isinstance(bindings, dict)` (273). `_bindings` likewise returns early on a non-dict (227). So an array-form `in: [{id: n, source: s}]` is forgiven by the exclusion list and never restored by either reader - the binding's `default`, `valueFrom` and `linkMerge` go uncompared along with the source.

Both silently return "equivalent," which is the failure mode this file's own thesis names: *"a wrongly-compared key is a loud false divergence, a wrongly-forgiven one is silent."*

If the position is that the relation only ever sees compiler output, and compiler output is always list-form `steps` with mapping-form `in`, then say so and **assert it** - `raise TypeError` on the unexpected shape rather than coercing to `[]`. A coercion that turns an unhandled shape into "no difference" is the one behaviour a comparison oracle must not have. Otherwise read both forms here for the same reason the three functions above do.

tests/core/test_generators.py
```diff
@@ -0,0 +1,186 @@
...
+    """
+    compiled: Counter[str] = Counter()
+
+    @settings(max_examples=50, suppress_health_check=list(HealthCheck), deadline=None)
+    @given(strat.compilable_documents())
+    def _collect(document: Document) -> None:
+        form = 'mapping' if document.steps_as_mapping else 'sequence'
+        compiled['drawn'] += 1
+        try:
+            compile_hermetic_cwl(strat.to_yml(document), 'oracle')
```
darkclad: **🟡 The test named for `workflows()` never calls `workflows()`, and nothing else does either.**

I grepped the whole tree: `workflows()` (ast_strategies.py:406) has **zero call sites**. The only mentions outside its own definition are this test's docstring and its two assertion messages.

This test rebuilds the strategy by hand - `@given(strat.compilable_documents())` at 145, `strat.to_yml(...)` at 150 - which is `compilable_documents().map(to_yml)`, i.e. `workflows()`'s body (line 413) retyped. So the docstring's premise, *"`workflows()` is what Tasks 3-7 quantify over, and nothing exercised it,"* is still true after the test. Replacing the body of `workflows()` with `return st.none()` leaves the suite green.

The assertion messages make it harder to notice, not easier: lines 159 and 162 both say "from workflows()", so the one artifact a reader would consult to decide whether the strategy is covered asserts that it is.

The obstacle to the one-line fix is real - line 147 needs `document.steps_as_mapping`, and `workflows()` yields only the `Yaml`. Two ways out, either fine:

* Have `workflows()` yield the pair (`compilable_documents().map(lambda d: (d, to_yml(d)))`), or add a `workflows_with_documents()` beside it and define `workflows()` in terms of that, so the composition under test is the one shipped.
* Keep this test as the compile-coverage test it actually is, rename it accordingly, and add a separate two-line test that draws from `workflows()` and asserts the result equals `to_yml` of the corresponding document.

The first is better: it removes the duplicate composition rather than testing it twice.

tests/core/test_hermeticity.py
```diff
@@ -0,0 +1,296 @@
...
+
+    Writes a module that imports the environment, points the graph at it, and
+    asserts the crossing is reported. Reverting the FORBIDDEN tuple to empty
+    fails this test.
+    """
+    breach = tmp_path / 'core' / 'breach.py'
+    breach.parent.mkdir()
+    breach.write_text('import sophios.plugins\n', encoding='utf-8')
+    graph = {'core.breach': _imports_of_test_module(breach)}
+    assert any(t in FORBIDDEN for t in graph['core.breach'])
```
darkclad: **🟡 This asserts on `FORBIDDEN` and `_imports_of_test_module`, not on the scan.**

The docstring promises three things - "points the graph at it", "asserts the crossing is reported", "reverting the FORBIDDEN tuple to empty fails this test." Only the third is true.

Line 132 is `any(t in FORBIDDEN for t in graph['core.breach'])`: it reads the module's own import set and intersects it with a constant. `_reachable` is never called. The crossings comprehension from `test_no_oracle_module_reaches_plugin_discovery` is never run. `graph` is built and then used only as a dictionary lookup - the "points the graph at it" step happens syntactically but has no effect on what is asserted.

So the guard this test exists to prove *can fire* stays green through mutations that break it outright: make `_reachable` return `set()`, invert the crossings filter, or drop `ORACLE_MODULES` to empty - line 132 touches none of them. Given that this module found its sibling defect by mutation (lines 250-252, "adding `import sophios.plugins` to `test_generators.py` left `test_no_oracle_module_reaches_plugin_discovery` green"), the same treatment applied here would have caught it.

Fix: extract the reachability + crossings computation out of `test_no_oracle_module_reaches_plugin_discovery` into a helper both tests call, seed the graph so `core.breach` is reachable from an `ORACLE_MODULES` entry, and assert the crossing appears in the helper's *result*. Then the mutation the docstring describes is one of several that turn it red, instead of the only one.

tests/core/test_hermeticity.py
```diff
@@ -0,0 +1,296 @@
...
+    `os.environ`. `-p core._poison_plugins` is resolved at pytest's
+    `consider_preparse`, before collection would otherwise put `tests/` on
+    `sys.path` — so without `tests/` on `PYTHONPATH` up front, the plugin
+    itself fails to import (`ImportError: No module named 'core'`) and every
+    caller of this function fails for that reason instead of the one it
+    claims to test. Built explicitly, not merged with an inherited value: a
+    `PYTHONPATH` that happens to already include `tests/` in one shell is not
+    evidence this works in general.
+    """
+    python_path = f'{REPO_ROOT / "src"}:{REPO_ROOT / "tests"}'
```
darkclad: **⚪ Hardcoded `:` should be `os.pathsep`.**

`REPO_ROOT` is a `Path`, so on Windows this renders `D:\...\src:D:\...\tests` and `PYTHONPATH` splits on `;` - one unparseable entry, `tests/` never on `sys.path`, and `-p core._poison_plugins` fails with exactly the `ImportError: No module named 'core'` the docstring above says this explicit construction exists to prevent.

Both callers then fail for the wrong reason: `test_the_oracle_suite_passes_with_plugin_discovery_disabled` on `returncode == 0`, and `test_the_poison_fires_on_a_suite_that_needs_discovery` on `POISON_MESSAGE in output` - the second is caught only because lines 287-291 insisted on matching the message rather than a bare nonzero exit, which is the reasoning in that docstring paying off immediately.

Not a CI break today: `lint_and_test.yml`'s `windows-latest` leg never runs this file (its Spec 0 step names `test_zone_boundary.py` and `test_entry_points.py` only), and the whole-tree run that does reach it is `build_wheel.yml:34` on `ubuntu-latest`. So this is a Windows-dev-machine failure - until the Spec 2 CI step in my top-level note exists, at which point it becomes a matrix break. Worth fixing now while the reason is in view.

```python
python_path = os.pathsep.join((str(REPO_ROOT / 'src'), str(REPO_ROOT / 'tests')))
```

REVIEW-403.md
```diff
@@ -0,0 +1,154 @@
+# Adversarial review — PR #403 (`semrefac_7.1..semrefac_8`)
```
darkclad: **⚪ Per-PR review scratch in the repository root.**

Branch range, mutation table and reviewer notes for #403 specifically - this is PR-description or `design_docs/` material, not a root-level file the repo carries forever. `git log` and the PR already hold the range.

The mutation table genuinely is worth keeping - it is the evidence behind several "found by mutation" comments in the test files, and those comments are a large part of why those tests are trustworthy. Fold it into `design_docs/core-refactor-design.md` section 6 (or a `design_docs/spec-2-mutations.md`) and drop the rest.

darkclad (commented):
The stack: `semrefac_8` on `origin/master` (`cea6e31`), 10 commits, 12 files, +2468/-20 - Spec 2's test infrastructure. A hermetic synthetic tool registry, an AST-derived Hypothesis generator, a three-strength compilation-equivalence relation, a plugin-discovery poison, and three test files.

I could not run the suite here (no sophios env on this machine), so rather than read for the claims I re-implemented the hermeticity scan in stdlib `ast` and ran it against the real tree, hand-traced every `test_equivalence.py` fixture through `_dataflow`/`_same_dag`/`_first_difference`, and checked the compiler paths the generator depends on.

**The load-bearing claims hold.** `ORACLE_MODULES` all resolve and there are zero forbidden crossings, so P25's static half genuinely passes rather than passing vacuously. `recursively_delete_dict_key` is non-mutating, so the lattice tests are not order-dependent. `compile_hermetic`'s fourteen-argument call matches `compile_workflow`'s signature. Every equivalence fixture I traced reaches the verdict it asserts.

**The self-critical machinery is the best thing here.** `NOT_YET_COMPILABLE` being empty *and* keeping its non-vacuity companion, `_FORGIVEN_STEP_KEYS` written as an exclusion so an unknown key is compared rather than forgiven, the mutation notes recording which green tests were found to be worthless - that discipline is why the two blocking items below are findable at all.

**Both blockers are the same defect class this PR already fixed once.** `356df27` ("the relation could not see which workflow input feeds which port") corrected `_dataflow` for the input side after finding that dropping a name made a class of rewiring invisible. The identical argument is still standing, uncorrected, on the output side - `_port_shapes` excludes `outputSource` "for the same reason the key is: it is a name." The second is the generator leaking edge names from steps it discards, which `testing=True` then converts into a phantom workflow input rather than an error, so the corrupted document reaches the equivalence properties silently.

**One thing not in the diff: none of these files run in `lint_and_test.yml`.** Spec 0 got its own named step ("PyTest Spec 0 Invariants") explicitly so a red build is attributable to a zone without bisecting. Spec 2 gets nothing - `test_hermeticity.py`, `test_generators.py` and `test_equivalence.py` are reached only by `build_wheel.yml:34`'s undifferentiated whole-tree `pytest -m "not skip_pypi_ci" -vv`, on `ubuntu-latest` alone. That is also what makes the `os.pathsep` item below a dev-machine problem rather than a CI break today, and it is worth a "PyTest Spec 2 Oracle" step on the same principle Spec 0 was given one.

Two blocking, three substantive, two minor. All inline.

tests/core/test_hermeticity.py
vjaganat90: Fixed — `os.pathsep.join(...)`.

tests/core/equivalence.py
vjaganat90: Fixed, the way you describe. Each workflow output is a node labelled with its declared shape, with an edge from the producer carrying `(port, "")`, and `_port_shapes` goes on dropping the key. The swap is in `MUST_DIFFER`, and there is a test beside it pinning both directions — differently-shaped producers diverge, identically-shaped ones stay equivalent, since that permutation is what a renaming is. An output that lost its `outputSource` entirely now shows up too; the shape multiset could not see that either.

The reference and the binding are now read by one function, so the two positions cannot drift apart again.

tests/core/ast_strategies.py
vjaganat90: Fixed with the up-front unique draw, so nothing is discarded and `count` means what it says.

Two tests, because one is not enough: the invariant (`!*` only names an edge the document defines) and a detector that records what the generator drew and compares it against what the document kept. The invariant alone would have been a poor regression test — against the discarding generator a leak that a later step then referenced landed in 4 documents in 2000, and a gap in the edge numbering in 0 — so it states the claim and the other one catches the regression.

tests/core/equivalence.py
vjaganat90: Took the second option: `steps:` that is not a list, a step with no `id`, and a non-mapping `in:` all raise `TypeError` now, with the module docstring carrying a WILL NOT READ section saying which forms the relation reads and why the ports and requirements are the opposite case. Also closes the "a step with no `id` is invisible" item I had declared as not-fixed — same defect, same argument.

tests/core/test_generators.py
vjaganat90: Fixed the first way. `workflows_with_documents()` yields the pair, `workflows()` is its projection, and the compile test draws the pair — so the composition under test is the shipped one and there is only one of it. A projection is still one `map` away from covered, which is exactly where `st.none()` fits, so there is a short test that draws `workflows()` itself.

tests/core/test_hermeticity.py
vjaganat90: Fixed. `_crossings` is extracted and the firing test runs it, with the breach one hop past a real oracle module rather than read straight off the import set — so `_reachable` returning `set()`, an inverted filter and an empty tuple all turn it red now.

REVIEW-403.md
vjaganat90: Deleted. `design_docs/` carries the one locked architecture document and nothing else, so the mutation table went to the working folder with the trackers rather than into the repo — a record of how the suite was validated is not architecture. What the repo keeps is each mutation's lesson in the docstring of the test written in answer to it, which is where it is load-bearing.

darkclad (approved):

PR #404, opened 2026-09-08
Spec 2: meaning-preserving rewrites and the equivalence property

vjaganat90:
**Stacked on #403** — review that first; this branch contains its commits.

Task 3 defined `≡`. This adds the space of `f`, and the single parameterised property that partition independence (P28) and its nested case (P29) are named instances of.

Five rewrites that must not change meaning, each carrying the strength it preserves and the reason it does:

| rewrite | preserves | why it is meaning-preserving |
|---|---|---|
| `identity` | `IDENTICAL` | compiling one input twice must agree — not trivial, since the compiler mutates a module global, the tool registry, and four structures threaded through the recursion |
| `text_roundtrip` | `IDENTICAL` | serialising and reloading is what every user does between writing a file and compiling it |
| `split` | `UP_TO_RENAMING` | a subworkflow is the steps written in one file; grouping them differently renames every port and moves no edge |
| `inline_all` | `UP_TO_RENAMING` | `split` inverted, via `sophios.inlineing` — a split/inline pair agreeing only with each other would be an inverse-pair blind spot |
| `rename_workflow` | `UP_TO_RENAMING` | the stem is the first segment of every namespace, and must move no edge |

The rationale is a field rather than a comment because it is the load-bearing claim: a rewrite on this list asserts that Sophios's semantics are invariant under it, and that assertion is exactly what the property checks.

**Reordering independent steps is deliberately not on the list.** Inference scans backwards over previous steps and takes the most recent match (`inference.py:216`), so step order is part of what a workflow means in this language. A reorder rewrite would be a property asserting something false.

**Worth a look:**

- Three guards keep the green honest — no rewrite may be a no-op, `split` must really rename, and a deliberately dishonest rewrite (one dropping a step while claiming `UP_TO_RENAMING`) must be caught. Without the last, a mis-wired property comparing a document with itself satisfies the other two.
- `_flatten` inlines subworkflow steps so a split and an unsplit compilation are comparable. `inlineing.inline_subworkflow_cwl` does that job for a dict-keyed `steps:`; `compile_workflow` emits a list, so it could not be reused directly.

**Finding, not fixed here** — `src/` is out of scope for this spec:

**CE-15.** Reference §3.1 documents three step surface forms and says all three produce the same result. `sophios.lang` parses the third — a sequence of single-key mappings — into a `Step` with no diagnostic; `canonicalize_steps_list`'s list branch calls `require_step_id` on each element and raises. Its dict branch synthesises `id` from the key; the list branch never learned to. No file reachable through `search_paths_wic` uses the form, so nothing catches it.

That is why `--insert_steps_automatically` cannot work: `insert_step_into_workflow` writes exactly that form, after canonicalisation has run, so every insertion yields a step whose id reads as empty. It also means `list(set(insertions))` — the hazard the design names as determinism's motivating example — is unreachable, which is independent support for restating that property as canonical emission over the two default-path sites.

**Not in this PR.** Tasks 5–7: canonical emission, the single-compilation predicates (namespace injectivity, edge soundness, termination), and the canonical path.

### DIFF
--- a/tests/core/test_equivalences.py
+++ b/tests/core/test_equivalences.py
@@ -1,0 +1,324 @@
+"""P24c: every rewrite `transformations` names preserves the meaning it claims.
+
+One property, parameterised over `Transformation`, rather than a bespoke test
+per rewrite. P28 (partition independence) and P29 (its nested case) are named
+instances of it — `split` at one level and `split` composed with itself — not
+separate tests with their own idea of "the same workflow", which is how the
+two regression tests already in this tree ended up with two.
+"""
+import copy
+from typing import Any, Final
+
+import pytest
+import yaml
+from hypothesis import given
+from hypothesis import strategies as st
+
+from sophios.wic_types import NodeData, RoseTree, Yaml
+
+from . import ast_strategies as strat
+from .equivalence import Strength, equivalent
+from .hermetic import PARTITION, compile_hermetic, subworkflow_step
+from .synthetic_tools import STEMS, inputs_of
+from .transformations import TRANSFORMATIONS, Transformation, split, transformations
+
+#: The name both sides compile under. Identical on purpose: `identity` and
+#: `text_roundtrip` claim IDENTICAL, and `equivalent(..., Strength.IDENTICAL)`
+#: compares `steps[].id` literally (see `equivalence.py`'s own docstring —
+#: IDENTICAL forgives nothing) — so compiling "before" and "after" under
+#: different names would make every namespaced id disagree regardless of the
+#: rewrite, and IDENTICAL would never hold for anything. UP_TO_RENAMING
+#: ignores names either way, so the same choice costs those rewrites nothing.
+_COMPILE_NAME: Final = 'oracle'
+
+
+def _source_of(value: Any) -> str | None:
+    """The `source` one `in:` binding names, in whichever of the two surface
+    forms the compiler emitted (`{name: source}` or `{name: {source: ...}}`)."""
+    source = value.get('source') if isinstance(value, dict) else value
+    return source if isinstance(source, str) else None
+
+
+def _set_source(bindings: Yaml, name: str, source: str) -> None:
+    """Rewrites one binding's source in place, keeping its own surface form."""
+    value = bindings[name]
+    if isinstance(value, dict):
+        value['source'] = source
+    else:
+        bindings[name] = source
+
+
+def _flatten(rose: RoseTree) -> Yaml:
+    """The compiled document with every subworkflow step inlined, so a split
+    and an unsplit compilation of the same steps can be compared directly.
+
+    `sophios.inlineing.inline_subworkflow_cwl` already does this job — for a
+    `steps:` represented as a dict keyed by step id. `compile_workflow` emits
+    `steps:` as a *list* (`type(compiled_cwl['steps']) is list`, checked
+    directly against this oracle's own output), so that function's
+    `list(steps.keys())` raises `AttributeError` on exactly the shape this
+    task's compiler produces, and nothing in the tree calls it. That is a
+    finding about `src/sophios/inlineing.py`, not something this task's own
+    files can work around by patching — `src/` is out of scope here — so this
+    is a from-scratch replacement, scoped to what `equivalent()` at
+    UP_TO_RENAMING actually forgives: it needs the DAG's topology and every
+    step's body right, and it does not need `run:` paths, `steps[].id`, or
+    workflow-level port *names* right at all, because those are exactly what
+    that strength already ignores (`equivalence.py`).
+
+    Recognising a subworkflow step by its own sub-tree's `class` (`Workflow`,
+    not `CommandLineTool`) rather than by an `id` suffix: `rose.sub_trees` has
+    one entry per step regardless of kind (verified directly — a flat,
+    two-`CommandLineTool` compile still has two leaf sub-trees, one per step),
+    so the id is not what marks a step as needing to be inlined here; its own
+    compiled class is.
+    """
+    node_data: NodeData = rose.data
+    document: Yaml = copy.deepcopy(node_data.compiled_cwl)
+    steps: list[Yaml] = document.get('steps', [])
+    sub_trees: list[RoseTree] = rose.sub_trees
+    flat_steps: list[Yaml] = []
+    #: `{wrapper_id}/{wrapper_out_port} -> real producer's own source string`,
+    #: for every subworkflow step this pass inlines away.
+    redirects: dict[str, str] = {}
+    #: Requirements the inlined subworkflows themselves declared (their own
+    #: `SubworkflowFeatureRequirement` already stripped, recursively, by the
+    #: nested `_flatten` call below) — e.g. `ScatterFeatureRequirement`, for a
+    #: `scatter:` that lived inside the subworkflow and is now a step directly
+    #: in *this* document. A CWL document declares only the features its own
+    #: `steps:` uses, so once a step moves in here, whatever requirement it
+    #: needed moves with it.
+    inherited_requirements: dict[str, Yaml] = {}
+    inlined_any = False
+
+    for step, sub in zip(steps, sub_trees):
+        sub_data: NodeData = sub.data
+        if sub_data.compiled_cwl.get('class') != 'Workflow':
+            flat_steps.append(step)
+            continue
+        inlined_any = True
+        inner = _flatten(sub)
+        inner_requirements = inner.get('requirements')
+        if isinstance(inner_requirements, dict):
+            inherited_requirements.update(inner_requirements)
+        # The wrapper's own `in:` maps the subworkflow's formal parameter
+        # names (its inner workflow-level `inputs:` keys) to the sources that
+        # actually feed them at *this* level.
+        formal_to_actual = {name: _source_of(value) for name, value in step.get('in', {}).items()}
+        for inner_step in inner.get('steps', []):
+            for name, value in inner_step.get('in', {}).items():
+                source = _source_of(value)
+                if source is not None and source in formal_to_actual:
+                    actual = formal_to_actual[source]
+                    if actual is not None:
+                        _set_source(inner_step['in'], name, actual)
+            flat_steps.append(inner_step)
+        wrapper_id = step['id']
+        # The subworkflow's own `outputs:` map its output port names to
+        # `outputSource: '<inner step>/<port>'`; anything outside referred to
+        # `{wrapper_id}/{that port name}`, so redirect that reference straight
+        # to the real producer, which is now itself one of `flat_steps`.
+        for out_name, out_value in inner.get('outputs', {}).items():
+            redirects[f'{wrapper_id}/{out_name}'] = out_value['outputSource']
+
+    for flat_step in flat_steps:
+        for name, value in flat_step.get('in', {}).items():
+            source = _source_of(value)
+            if source is not None and source in redirects:
+                _set_source(flat_step['in'], name, redirects[source])
+    document['steps'] = flat_steps
+
+    for out_value in document.get('outputs', {}).values():
+        out_source = out_value.get('outputSource')
+        if isinstance(out_source, str) and out_source in redirects:
+            out_value['outputSource'] = redirects[out_source]
+
+    if inlined_any:
+        # A flattened document contains no subworkflow step, so the
+        # requirement that declares one no longer applies — left in place, it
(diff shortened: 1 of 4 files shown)


tests/core/test_equivalences.py
```diff
@@ -0,0 +1,324 @@
...
+                continue
+            coerce = int if arg_type == 'int' else float
+            try:
+                coerce(literal)
+            except (TypeError, ValueError):
+                return True
+    return False
+
+
+@pytest.mark.skip_pypi_ci
```
darkclad: **🟡 All four tests in this file are excluded from every CI job.**

`skip_pypi_ci` is on all four (`:198`, `:251`, `:283`, `:308`), and every path that could run them filters it out or does not name the file:

| where | invocation | runs this file? |
|---|---|---|
| `build_wheel.yml:34` | `pytest -m "not skip_pypi_ci" -vv` | no — filtered |
| `test_and_publish_pypi.yml:43` | `pytest -m "not skip_pypi_ci" -vv` | no — filtered |
| `lint_and_test.yml:244` (PyTest Spec 2 Oracle) | `pytest ... test_equivalence.py -m "not skip_pypi_ci"` | no — names `test_equivalence.py`, singular |
| `test_hermeticity.py:288` (poisoned run) | `_run_poisoned(ORACLE_FILES)` | no — see below |

Those two whole-repo runs are the only ones in the repository, so the marker alone is enough to make the file dead. The Spec 2 step added in #403 would be the natural home, but it lists `test_equivalence.py` and this PR does not touch that workflow — the two filenames differing by one character is easy to read past.

The poisoned run is the second half of the same gap. `ORACLE_FILES` (`test_hermeticity.py:225`) still carries

```python
    # 'tests/core/test_equivalences.py',
```

commented out, so `test_the_oracle_suite_passes_with_plugin_discovery_disabled` does not execute this file either. Worth flagging that uncommenting that line alone will turn `test_the_scan_discovers_the_modules_it_claims_to_cover` red: this PR added `'core.transformations'` to `ORACLE_MODULES` but not `'core.test_equivalences'`, and that test asserts `ORACLE_FILES ⊆ ORACLE_MODULES`. Both lists need the entry, and the module's own note at `test_hermeticity.py:41` already says so — it lists `'core.transformations', 'core.test_equivalences'` together as what Task 7 restores.

On the marker itself: the siblings use it far more narrowly. `test_hermeticity.py` marks exactly one test with it — `test_every_stub_is_valid_cwl` (`:156`), which shells out to `cwltool --validate` — and `test_equivalence.py` and `test_generators.py` use it nowhere. Here it is blanket-applied, including to three `@pytest.mark.fast` tests (`:251`, `:283`, `:308`) that compile hermetically and have no external dependency at all. `test_the_property_fails_for_a_rewrite_that_changes_meaning` is the meta-test that proves the property is not mis-wired; it costs nothing to run and there is no reason it cannot run on PyPI.

Suggested: drop `skip_pypi_ci` from the three `fast` tests, keep it on the `slow` property if the runtime is the concern, add `tests/core/test_equivalences.py` to `lint_and_test.yml:244`, and uncomment `ORACLE_FILES` line 225 together with adding `'core.test_equivalences'` to `ORACLE_MODULES`.

tests/core/test_equivalences.py
```diff
@@ -0,0 +1,324 @@
...
+    could never tell them apart. What a no-op could fake there is skipping the
+    round trip entirely and handing back the very object it was given, so
+    that is what is checked: a fresh, value-equal object, not a different one.
+    """
+    yml = copy.deepcopy(_NOOP_GUARD_FIXTURE)
+    if rewrite.name == 'identity':
+        pytest.skip('identity changes the input by definition; see the module docstring')
+    result = rewrite.apply(copy.deepcopy(yml))
+    if rewrite.preserves is Strength.IDENTICAL:
+        assert result is not yml, f'{rewrite.name} returned its input unchanged, doing no work'
```
darkclad, line 342: **🟡 This assertion cannot fail, and the fake its docstring names passes the test.**

Line 275 hands `apply` a fresh deep copy:

```python
result = rewrite.apply(copy.deepcopy(yml))
```

so `result is not yml` holds for every possible implementation of `apply`, including `lambda w: w`. The object identity being checked is not the one the docstring is reasoning about.

That matters here more than it usually would, because lines 268-270 state precisely what this guard exists to catch:

> What a no-op could fake there is skipping the round trip entirely and handing back the very object it was given, so that is what is checked: a fresh, value-equal object, not a different one.

Demonstrated at this head — I replaced `_text_roundtrip`'s body with exactly that fake:

```python
def _text_roundtrip(document: Yaml) -> Yaml:
    return document
```

and the file stays green: `6 passed, 1 skipped`. `test_every_transformation_actually_transforms[text_roundtrip]` passes, because `result is not yml` is satisfied by the deepcopy and `result == yml` is satisfied by returning the input. The IDENTICAL branch is asserting nothing about `text_roundtrip` at all.

The fix is to give `apply` the object whose identity is being tested, and compare against a pristine copy instead:

```python
yml = copy.deepcopy(_NOOP_GUARD_FIXTURE)
if rewrite.name == 'identity':
    pytest.skip('identity changes the input by definition; see the module docstring')
expected = copy.deepcopy(yml)
result = rewrite.apply(yml)
if rewrite.preserves is Strength.IDENTICAL:
    assert result is not yml, f'{rewrite.name} returned its input unchanged, doing no work'
    assert result == expected, f'{rewrite.name} claims IDENTICAL but changed the value'
else:
    assert result != expected, f'{rewrite.name} is a no-op'
```

I checked this both ways: it stays green on the real implementations, and it turns red on the `return document` fake above. Comparing against `expected` rather than `yml` also keeps the `else` branch honest if a rewrite ever mutates in place — the current `result != yml` would compare a mutated input against itself.

Worth noting the `else` branch has no such problem today: `apply` genuinely returns a new object there, so `result != yml` is a real comparison. This is specific to the IDENTICAL branch.

tests/core/transformations.py
```diff
@@ -0,0 +1,289 @@
...
+#: actually given, so this bound only needs to be generous, not exact.
+_SPLIT_STEP_BOUND: Final = 5
+
+
+def transformations() -> SearchStrategy[Transformation]:
+    """Every transformation this module knows: the constant four, plus a
+    freshly drawn `split`."""
+    # pylint cannot see through @st.composite and reads partitionings() as
+    # returning its element type rather than a SearchStrategy.
+    return st.one_of(st.sampled_from(TRANSFORMATIONS),
```
darkclad: **⚪ Nothing pins what this strategy draws, and roughly half the budget goes to `identity`.**

`st.one_of` weights its branches by Hypothesis's own heuristics, not evenly, and `identity` is the first element of `TRANSFORMATIONS`. Measured by drawing exactly as `test_a_meaning_preserving_rewrite_preserves_meaning` does — `workflows()` filtered the same way, then `transformations()` — at `PARTITION`'s shipped `max_examples=50`, twelve independent runs:

| run | identity | text_roundtrip | inline_all | rename_workflow | split |
|---|---|---|---|---|---|
| 0 | 25 | 3 | 1 | 5 | 16 |
| 1 | 19 | 5 | 4 | 8 | 14 |
| 4 | 16 | 2 | 2 | 3 | 27 |
| 6 | 24 | 1 | 11 | 4 | 10 |
| 9 | 22 | 1 | 4 | 1 | 22 |
| 11 | 23 | 9 | 4 | 4 | 10 |

(six of twelve shown; the rest sit in the same band.) `identity` averages ~22 of 50 — **44% of the budget** — and it is the one rewrite whose own rationale says it tests the compiler's idempotence rather than any rewrite. Meanwhile the thinnest real rewrite drops to **1 example in 50** in several runs. To be clear about what I did *not* find: every transformation was drawn at least once in all twelve runs, so this is not "never exercised" — it is that the property is much thinner for some rewrites than the budget suggests, and nothing would say so if it got thinner still.

The sibling module has exactly this guard, for exactly this reason — `test_equivalence.py:235`:

```python
def test_every_rewrite_is_actually_drawn() -> None:
    """...both are *weaker* in proportion to how many rewrites the strategy
    stops producing — a `sampled_from` that only ever yielded `identity` would
    leave them asserting reflexivity and nothing else, silently."""
```

`test_every_transformation_actually_transforms` does not cover this: it is `parametrize`d over `TRANSFORMATIONS` directly and never calls `transformations()`, so it says nothing about the draw. And `split` is not in `TRANSFORMATIONS` at all, so the only rewrite built by a strategy is the one that guard cannot reach.

**The same silence covers P29.** The module docstring of `test_equivalences.py` calls P29 "`split` composed with itself", but this strategy yields one `Transformation` per example and the property applies it once — nothing is ever composed. The nested case is reached only incidentally, when `workflows()` happens to draw a subworkflow step and `split` wraps it. I measured that and it is currently fine — nesting depth ≥ 2 in **7 of 14** split draws at budget 50, and 84 of 130 at 500 — so P29 is genuinely exercised today. But that is a property of `ast_strategies.documents()`, not of anything in this file: if the generator's subworkflow-step branch narrowed, P29 coverage would go to zero and every test here would stay green.

Both are the same small fix — a `test_every_transformation_is_actually_drawn` mirroring `test_equivalence.py:235`, extended with the one extra assertion that a drawn `split` reaches a nested document. That turns two facts that are currently true by luck into two facts that are checked.

darkclad (commented):
Re-review at `510df4c`. The branch was rebased since my last read, and that resolved everything structural I had flagged: the stale duplicate of #402 (which was reverting `cea6e31`) is gone, #403 has landed, and the PR is now exactly three files on top of current `master` — `tests/core/test_equivalences.py`, `tests/core/transformations.py`, and one line in `test_hermeticity.py`. 614 insertions, nothing else. That is the shape I hoped for and it made this pass much easier.

**Ran it at this head: `6 passed, 1 skipped`** (`pytest tests/core/test_equivalences.py`, Windows, Python 3.13, branch materialised into a scratch tree — the repo working tree was not touched).

**The property is load-bearing, not decorative.** I checked the part I most expected to be untested: deleting the `parentargs['in']` wiring in `_wrap_steps` (`transformations.py:187`) — the line whose docstring argues at length that leaving a wrapper argument unwired lets the compiler backward-infer it onto a preceding sibling group's output — turns `test_a_meaning_preserving_rewrite_preserves_meaning` red at the shipped 50-example budget. The docstring's claim is real and the property catches its violation.

Three findings inline, none of them blocking on their own. The first is the one I would actually fix before merge: **as it stands, none of these four tests run in any CI job**, so the work above is currently unenforced.

Nothing else to raise — `_flatten`'s from-scratch replacement for `inline_subworkflow_cwl`, the `_hits_the_scalar_coercion_gap` filter, and the dishonest-transformation test all hold up, and the reasoning recorded for each is accurate to what the code does.

tests/core/test_equivalences.py
vjaganat90: Fixed, all four parts. The marker is gone from every test in the file — its meaning here is "needs something outside the repo", and its only other use in the oracle is the one that shells out to `cwltool`; these compile hermetically. The file joins the Spec 2 step, and both oracle lists together, since the scan test requires exactly that.

Thank you for tracing the four invocations. I had added the CI step in #403 and then written a file it does not name, one character apart, which is the kind of thing I would have read past for a while.

tests/core/test_equivalences.py
vjaganat90: Right, and demonstrated exactly as you did — `_text_roundtrip` returning its argument kept the file green.

`apply` now gets `yml` itself and the comparison is against a pristine `expected`, so the identity being asserted is the one `apply` received. The `else` branch compares against `expected` too, for the reason you give: a rewrite that mutated in place would otherwise be compared against itself. Your fake turns it red.

tests/core/transformations.py
vjaganat90: Your measurement changed the fix. I first added the guard you suggested and it failed about one run in five — on `rename_workflow` drawing zero, exactly the thinness you measured. So the guard was right and the distribution was the thing to fix.

It cannot be fixed by weighting: `st.sampled_from` favours its first element because the engine prefers simpler choices, and reordering only moves which rewrite is starved. So the property parametrises over the five kinds instead and each gets the whole budget, with `split`'s partitioning drawn inside it. `transformations()` is gone — with the property no longer drawing it, it would have been a strategy whose only caller was a test of itself, which is the defect #403 had.

That leaves your second point as the only thing still down to chance, and it now has its own test: emptying `documents()`'s subworkflow branch turns `test_split_reaches_a_document_that_is_already_nested` red, where before P29 would have gone to zero in silence.

darkclad (approved):

PR #405, opened 2026-09-08
Spec 2: weekly compile-cost benchmarks

vjaganat90:
Follows merged #404; this is the first still-open PR in the remaining Spec 2 chain.

Weekly compile-cost trend. Explicitly not a gate: no thresholds, no assertions, `main()` returns 0 whatever the numbers say, and a case that raises is reported in the table and skipped. A benchmark that can fail a build is a gate, and the design is explicit that the guard for correctness is the property suite, which runs on every change.

**Five cases**, each carrying its rationale as a field rather than a comment:

- the corpus, for the trend that actually matters
- `speculative_insertion`, because that is where the documented `O(2^n)` lives and no corpus workflow reaches it
- `deep_nesting`, since namespacing cost grows with depth and nothing else on the list has depth
- `wide_inference`, the quadratic one — `perform_edge_inference` scans backwards over previous steps
- `parse_only`, so a slow week has one candidate cause instead of two

**No `pytest-benchmark`.** Timing is `time.perf_counter` and the output is a Markdown table written to the step summary. A dependency for a job that gates nothing is not worth the install surface, and week-over-week on one runner needs no statistics library.

**Friday cron**, so it does not contend with `fuzzy_compile_weekly.yml` (Wednesday) for the self-hosted runner. A `workflow_dispatch` input runs a deliberately failing case, so the non-gating claim can be demonstrated rather than asserted.

`benchmarks.py` imports `wic_corpus` on purpose — the corpus half is the point — and is therefore excluded from the oracle suite's hermeticity scan. Its docstring says so, so nobody "fixes" it later.

**Two findings from the first run**, neither fixed here:

**CE-15**, described in #404: the `speculative_insertion` case raises, because `--insert_steps_automatically` writes a step form the compiler refuses. Reported in the table and skipped, which is the harness behaving as designed.

A whitelist converter's input `format` must be a single-element **list**, never a bare string. The insertion search flattens it with `utils.flatten`, so a bare string is iterated character by character and the search silently never matches. Recorded in the case's own docstring; documented nowhere else.

Both findings are fixed in the next PR, #408; they remain findings rather than fixes in this branch's own diff.

**The two-consecutive-run comparison** that CR-203's demo calls for needs two dispatches a week apart, so it is yours to trigger rather than something I can show here.


tests/core/benchmarks.py
```diff
@@ -0,0 +1,315 @@
...
+
+
+@pytest.mark.fast
+def test_every_case_has_a_rationale() -> None:
+    """A case nobody can explain is a number nobody will act on."""
+    assert all(c.rationale.strip() for c in CASES)
+
+
+@pytest.mark.fast
+def test_the_harness_reports_a_failing_case_without_failing_itself(
```
darkclad, line 297: **🔴 Neither test in this file is ever collected, including the one whose whole job is to prove the non-gating claim.**

pytest's default `python_files` is `test_*.py *_test.py`. `pyproject.toml`'s `[tool.pytest.ini_options]` sets `markers` and `addopts = ""` and nothing else — no `python_files` override, and there is no `pytest.ini`, `setup.cfg`, or `tox.ini` in the repo. `benchmarks.py` matches neither pattern.

That only stops mattering if some CI step names the file on the command line, and none does:

- `lint_and_test.yml:244` runs the oracle suite by explicit filename — `test_hermeticity.py test_generators.py test_equivalence.py test_equivalences.py` — and `benchmarks.py` is not among them.
- `build_wheel.yml:34` runs bare `pytest -m "not skip_pypi_ci"`, which is default collection.
- `benchmarks_weekly.yml` runs `python tests/core/benchmarks.py`, which executes `main()` and never touches the two `test_` functions.

Reproduced with the repo's exact config (`markers` + `addopts = ""`, `tests/core/__init__.py` present, a `benchmarks.py` and a `test_generators.py` side by side): `1 test collected`, and it is the one in `test_generators.py`.

The irony is load-bearing. `test_the_harness_reports_a_failing_case_without_failing_itself` argues for its own existence: *"without it, 'benchmarks never fail a build' is an assertion about code nobody has seen fail."* As written it is an assertion about a test nobody has seen run — the `monkeypatch.setattr('core.benchmarks.CASES', ...)` target, the `capsys` assertion, and the `main() == 0` check have never executed in this repository, and a typo in any of them would be invisible.

Two ways out. Either move the two tests to `tests/core/test_benchmarks.py` importing `core.benchmarks` (keeps default collection working and keeps the runnable script a script), or add `benchmarks.py` to the explicit file list at `lint_and_test.yml:244`. The first is less coupling; the second is one line. Either way it is worth a line in `lint_and_test.yml` so the fast tests actually gate on this file.

.github/workflows/benchmarks_weekly.yml
```diff
@@ -0,0 +1,127 @@
...
+      - name: Install Molecular Modeling Workflows
+        if: always()
+        # Also run mm-workflows command to generate
+        # mm-workflows/autogenerated/schemas/config_schemas.json
+        # NOTE: Use ".[test]" instead of ".[all_except_runner_src]"
+        # We do not want or need to install the workflow_deps extra.
+        # (Many of the packages conflict with pypy.)
+        run: cd mm-workflows/ && pip install ".[test]" && mm-workflows --generate_schemas
+
+      - name: Generate Sophios Validation Jsonschema
```
darkclad, line 101: **🟠 The corpus this job benchmarks is a strict subset of the corpus every other job walks — the Python-API generation step is missing.**

`fuzzy_compile_weekly.yml`, whose shape this file follows, has one more step right here:

```yaml
      - name: Generate Sophios Python API Workflows (*.py -> *.wic)
        if: always()
        run: cd sophios/ && pytest tests/core/test_python_api.py -k test_compile_python_workflows
```

`lint_and_test.yml:196` and `run_workflows.yml:162` run it too, and in `lint_and_test.yml` it runs *before* the oracle suite at line 244 for exactly this reason. It is the only step in the diff against `fuzzy_compile_weekly.yml` that is dropped rather than renamed.

It is not cosmetic. `test_compile_python_workflows` does `retval.write_wic(path.parent, inline_subworkflows=False)` (`test_python_api.py:1035`), writing generated `.wic` files next to the `.py` sources that `_iter_python_workflow_paths` found via `get_py_paths`. Those directories are under `search_paths_wic` — `update_sophios_config.py:20-23` points it at `image-workflows/workflows` among others — so `wic_corpus._discover()`, which is `get_yml_paths(config)`, picks those files up *only if the generation step has already run*.

Without it, `CORPUS` here silently omits every Python-API-generated workflow. That contradicts `wic_corpus`'s own docstring — *"CI reaches everything and is the arbiter of full coverage"* — for this job specifically, and because `mm-workflows` and `image-workflows` are deliberately unpinned, the gap widens on its own as those repos add Python workflows. Nothing surfaces it: a smaller corpus is a shorter table, not an error, and the job cannot go red.

Adding the step back costs one `pytest -k` invocation on a job that already installs the full toolchain.

tests/core/benchmarks.py
```diff
@@ -0,0 +1,315 @@
...
+        _corpus_env = _CorpusEnv(tools, yml_paths, validator)
+    return _corpus_env
+
+
+def _compile_corpus_file(path: Path) -> None:
+    """Compile one corpus workflow for real, from disk, against the real
+    plugin registry. Mirrors `tests/core/test_examples.py::run_workflows`'s
+    compile-only half — the only pattern in this repository for compiling a
+    `.wic` file from disk rather than an in-memory dict."""
+    env = _get_corpus_env()
```
darkclad, line 117: **🟠 The first corpus row's number is plugin discovery, not compile cost.**

`_get_corpus_env()` is lazy, and the docstring gives a good reason — *"importing this module to run only `parse_only` or the synthetic cases should not pay for plugin discovery."* But `main()` puts the corpus cases first and times each one with `perf_counter` around `case.run()`, so the cost the laziness deferred lands *inside the first timed case*: `get_args()`, `io.get_config`, and `get_tools_cwl`, which is a filesystem walk over every entry of `search_paths_cwl` (`image-workflows/cwl_adapters`, `biobb_adapters/biobb_adapters`, `mm-workflows/cwl_adapters`, `sophios/cwl_adapters`).

Two consequences, both against the stated purpose:

- Whichever corpus file sorts first is the one row not comparable to the others, so "compare two weeks by eye" has one row that is a different measurement wearing the same units.
- Its week-over-week movement tracks the size of the plugin registry — which changes whenever the unpinned adapter repos change — rather than compile cost. A week where `mm-workflows` adds fifty adapters shows up as "the first corpus workflow got slower."

The fix is a line in `main()`: call `_get_corpus_env()` once before the loop, outside the timer. If the setup cost is itself interesting — and it plausibly is, since it is the thing that grows with the corpus repos — give it its own `Case` with its own row, which also keeps the laziness honest for anyone importing the module for `parse_only` alone.

.github/workflows/benchmarks_weekly.yml
```diff
@@ -0,0 +1,127 @@
...
+
+      - name: Prove the job stays green (dispatch-only)
+        if: inputs.prove_non_gating == 'true'
+        # Not a real case failing — a whole step failing, deliberately, to
+        # demonstrate the claim on demand rather than merely asserting it: a
+        # guard nobody has seen fire is a guard whose green means nothing.
+        continue-on-error: true
+        run: exit 1
+
+      - name: Run benchmarks
```
darkclad, line 113: **🟠 No `timeout-minutes`, on the one job that deliberately runs a case chosen for exponential blowup.**

`continue-on-error` and `|| true` make the job immune to *exit codes*. They do nothing about wall clock, and wall clock is the failure mode this case list goes out of its way to invite: `speculative_insertion` is on the list precisely because `design_docs/core-refactor-design.md:52` documents speculative compilation as `O(2^n)`.

Today it raises — CE-15, disclosed in the PR body — so it returns in microseconds. The PR body also says #408 fixes CE-15. The moment it does, this becomes the first job in the repo to run an unbounded exponential search with nothing bounding it, on the single shared self-hosted runner, weekly. A hang there holds the runner until GitHub's 360-minute job limit and then ends the job red — the exact outcome the non-gating design goes to lengths to prevent, arriving by the one door `continue-on-error` does not cover. `concurrency.cancel-in-progress` does not help either: it only fires when a *second* run of this same workflow starts, and `fuzzy_compile_weekly` is a different workflow, so it queues behind.

The repo already reaches for this tool where it matters — `lint_and_test.yml:219` and `:225` both carry `timeout-minutes: 20` as backup timeouts. A `timeout-minutes` on this step (generous, since the corpus half is legitimately slow) plus `continue-on-error: true`, which is already there, keeps a hung case from becoming a held runner. A per-case timeout inside `main()` would be better still and would put `timeout` in the table next to `error:`, which is the same reporting contract the harness already honours.

tests/core/benchmarks.py
```diff
@@ -0,0 +1,315 @@
...
+    for case in CASES:
+        start = time.perf_counter()
+        try:
+            case.run()
+            elapsed = f'{time.perf_counter() - start:.3f}'
+        except Exception as exc:  # pylint: disable=broad-except
+            # A case that fails is reported and skipped. This job does not
+            # gate, so a broken case must not stop the other numbers arriving.
+            elapsed = f'error: {type(exc).__name__}'
+        print(f'| {case.name} | {elapsed} |')
```
darkclad, line 286: **🟡 One row per corpus file against one aggregate row for `parse_only`, and no total — which makes `parse_only`'s stated job impossible to do by eye.**

`parse_only`'s rationale is the sharpest one on the list: *"Separates a parser regression from a compiler regression; without it a slow week has two candidate causes and no way to choose."* Choosing means comparing parse cost against compile cost over the same input set.

But the two are emitted at different granularities. `_parse_corpus` loops over all of `CORPUS` inside one case and prints one number. The corpus compile is a generator expression over `CORPUS` producing one `Case` — and therefore one row — per file. There is no total row; `main()` prints a header, a row per case, and returns.

So the comparison `parse_only` exists to enable requires hand-summing every `corpus/...` row first. The in-repo floor is 14 files, and CI's config adds `mm-workflows/examples` and `image-workflows/workflows` on top, so this is not a short column. It also defeats the step summary's purpose as described — *"SHA and runner go above the table in the step summary, where two weeks can be compared by eye"* — since a 10% corpus-wide regression spread across a hundred millisecond-scale rows is invisible per row and never totalled.

A `| corpus (total) | ... |` row, and ideally a `| all cases (total) | ... |`, costs an accumulator in the loop and makes both the week-over-week read and the parser-vs-compiler split a two-number comparison. Worth keeping the per-file rows too — they are how you find *which* workflow regressed once the total says one did.

darkclad (commented):
Review at `c230c23`. Both files are new, so the whole diff is in scope (`835b941..c230c23`, 442 additions, 0 deletions).

The design is right and the restraint is the good kind: no thresholds, no `pytest-benchmark`, rationale as a field rather than a comment, and a dispatch input that demonstrates the non-gating claim instead of asserting it. Nine findings, all inline; the first four are worth fixing before merge, the rest are smaller.

**Checked and sound:** every API call matches its signature at this head — `compile_hermetic`'s keyword-only `insert_steps_automatically`, `subworkflow_step`'s `.wic` suffix assert, `get_tools_cwl(config, quiet=True)`, `read_ast_from_disk`'s six-arg shape, `compile_workflow`'s positional tail, `parse(text, filename=...)`. `SYNTHETIC_NS` is `'global'`, which is the namespace `_speculative_insertion_registry` registers under, so the speculative case reaches its tools rather than dying on "Neither mk_1 nor found!". `_compile_corpus_file` mirrors `test_examples.py::run_workflows` faithfully, including the shared registry. `wide_inference`'s premise holds: `inference.py:216` really is `for j in range(0, i)[::-1]`, and `mk_file.name` is a `string` against `File` outputs, so `types_match` never short-circuits the scan. Unsatisfied required inputs are promoted to workflow inputs rather than raising, so that case compiles rather than erroring. `core.benchmarks` is genuinely outside P25's reach — `ORACLE_MODULES` is an allowlist, not a glob, so importing `core.wic_corpus` costs nothing there; the docstring's claim is accurate. And `get_args()` builds its namespace from `_argv`, not `sys.argv`, so the bare call in `_get_corpus_env` is safe under pytest. I also chased `except Exception` missing `SystemExit` and dropped it: library `sys.exit` is gone on this branch (`diagnostics.py:135`), and the surviving one in `cwl_subinterpreter.py:281` is in `main()`, off the compile path.

tests/core/benchmarks.py
```diff
@@ -0,0 +1,315 @@
...
+# Case 4: wide inference.
+
+def _wide_inference_yml() -> Yaml:
+    """Twenty `mk_file` steps, each with its one `name` input left
+    unspecified. `name` is a string and every prior output is a `File`, so
+    `types_match` never accepts — `perform_edge_inference`'s backward scan
+    always runs to completion instead of returning at the first match, which
+    is what makes this case pay for the full quadratic scan rather than the
+    linear common case."""
+    return {'steps': [{'id': 'mk_file'} for _ in range(20)]}
```
darkclad, line 215: **🟡 Twenty steps is too few for the quadratic to be the thing this row measures.**

The premise is verified — `inference.py:216` is `for j in range(0, i)[::-1]`, and with `mk_file.name` a `string` against `File` outputs the scan genuinely never returns early. But at n=20 the scan does 190 iterations total, against twenty full step compilations: namespacing, `desugar_into_canonical_normal_form`, rose-tree construction, CWL emission. The quadratic term is a rounding error inside the linear one.

Which means this row's week-over-week movement is dominated by per-step fixed cost, and a regression that doubled the inference scan specifically — the thing the rationale says this case is here to catch — would move the number by less than runner variance. §6.3's own accepted limit, *"runner variance swamps small changes,"* bites hardest exactly here.

n=100 makes the scan 5,050 iterations against 100 compiles — a 25x growth in the quadratic term against 5x in the linear one, so the ratio moves by 5x and the row starts reporting what it claims to. Alternatively keep n=20 and add a second row at a larger n: two points on the curve turn "this got slower" into "this got slower superlinearly," which is the actual question. Either is cheap on a weekly job.

.github/workflows/benchmarks_weekly.yml
```diff
@@ -0,0 +1,127 @@
...
+        # design_docs/core-refactor-design.md §6.3.
+        continue-on-error: true
+        run: |
+          cd sophios/
+          {
+            echo "SHA: $(git rev-parse --short HEAD)"
+            echo "Runner: ${{ runner.name }} (${{ runner.os }})"
+            echo
+          } >> "$GITHUB_STEP_SUMMARY"
+          python tests/core/benchmarks.py | tee -a "$GITHUB_STEP_SUMMARY" || true
```
darkclad, line 127: **🔵 The comment's account of `|| true` is wrong — it cannot do what it says, because `tee` already did it.**

The comment at line 115 reads: *"`continue-on-error` and `|| true` together: the first keeps a crashed step from failing the job, the second keeps a non-zero exit from marking the step red at all."*

`defaults.run.shell` is `bash -l {0}`, which overrides GitHub's default `bash --noprofile --norc -eo pipefail {0}` and drops `pipefail` with it. Without `pipefail`, a pipeline's exit status is the *last* command's — `tee`'s — and `tee` exits 0 whether or not `python` did. So `python ... | tee ... || true` masks the Python exit code at the pipe, and the `|| true` never has a non-zero status to swallow. It is not wrong, just inert, and the comment explains a mechanism that is not the one operating.

The part worth keeping in mind is what the masking actually buys: a crash *before* the first `print` — an import error, a failure inside `_speculative_insertion_registry()` at module scope, a `get_args()` blowup — produces a step summary containing the SHA and runner header and no table at all, a green step, and a traceback only in the raw log. Since the job is non-gating by design that is the intended trade, but it is worth saying so here rather than describing `|| true` as the thing doing the work.

Either drop `|| true` and say `tee` is what makes the step green, or keep it and add `set -o pipefail` above, in which case `|| true` becomes the mechanism the comment already claims it is.

tests/core/benchmarks.py
```diff
@@ -0,0 +1,315 @@
...
+         "store from scratch costs minutes, not milliseconds.",
+         run=lambda p=path: _compile_corpus_file(p))  # type: ignore[misc]
+    for path in CORPUS
+) + (
+    Case('speculative_insertion',
+         "A workflow with two branches, each needing one whitelisted-format-"
+         "converter insertion — the design's own phrase, 'two or more "
+         "whitelisted converter tools', for what puts a compilation on this "
+         "path, and design_docs/core-refactor-design.md §6.3 names it as "
+         "where the documented O(2^n) lives. The corpus never reaches it. On "
```
darkclad, line 248: **🔵 §6.3 does not document the `O(2^n)`; line 52 does.**

The rationale says *"design_docs/core-refactor-design.md §6.3 names it as where the documented O(2^n) lives."* §6.3 runs from line 488 to 501 and says only *"Coverage spans the corpus plus the speculative-insertion path"* — it names the path as in scope, which is a different claim. The `O(2^n)` is at line 52, under "The algorithms carry known hazards": *"Speculative compilation is `O(2^n)`."* (Line 563 refers back to it as "the documented `O(2^n)`", confirming line 52 is the documenting site.)

Two citations doing two jobs, currently collapsed into one that supports neither cleanly. Worth splitting: §6.3 for why the case is on the list, §1 for the complexity claim.

The neighbouring citation in `_speculative_insertion_registry`'s docstring is correct — "two or more whitelisted converter tools" is at line 470, inside §6.2 as claimed.

tests/core/benchmarks.py
```diff
@@ -0,0 +1,315 @@
...
+CASES: Final[tuple[Case, ...]] = tuple(
+    Case(f'corpus/{corpus_id(path)}',
+         "One of wic_corpus.CORPUS's reachable workflows, compiled for real "
+         "against this environment's plugin registry — the trend everyone "
+         "actually cares about. Needs a jsonschema validator only because "
+         "read_ast_from_disk's signature requires one; it is never called "
+         "(ignore_validation_errors=True), which is what keeps this case from "
+         "paying for autogenerated/schemas/schema_store.json — building that "
+         "store from scratch costs minutes, not milliseconds.",
+         run=lambda p=path: _compile_corpus_file(p))  # type: ignore[misc]
```
darkclad, line 240: **🔵 `functools.partial` here drops the default-argument trick and the `type: ignore` with it.**

```python
run=lambda p=path: _compile_corpus_file(p))  # type: ignore[misc]
```

The `p=path` default exists only to bind the loop variable, and the `type: ignore[misc]` exists only because mypy cannot reconcile that default against `Callable[[], object]`. Both disappear with:

```python
run=functools.partial(_compile_corpus_file, path))
```

which binds eagerly by construction and types cleanly as `Callable[[], None]`. One import, two fewer things a reader has to decode, and one fewer suppression in a file whose whole argument is that its numbers should be easy to trust.

Same neighbourhood, much smaller: `print(f'| case | seconds |')` and `print(f'|---|---:|')` carry `# pylint: disable=f-string-without-interpolation` for f-strings that interpolate nothing. Dropping the `f` prefixes removes both suppressions.

tests/core/benchmarks.py
vjaganat90: The harness contracts now live in `tests/core/test_benchmarks.py`, which default pytest discovery collects and the explicit Spec 2 Oracle step names directly.

.github/workflows/benchmarks_weekly.yml
vjaganat90: Python API workflow generation now runs before the benchmark script, so generated `.wic` files exist before `CORPUS` is discovered.

tests/core/benchmarks.py
vjaganat90: `main()` now initializes the shared corpus environment before starting any corpus timer, while retaining lazy initialization when no corpus case is selected.

.github/workflows/benchmarks_weekly.yml
vjaganat90: The benchmark step now has a 60-minute ceiling. Its existing `continue-on-error` policy keeps a timeout non-gating.

tests/core/benchmarks.py
vjaganat90: The table now retains every per-file row and adds `corpus (total)` and `all cases (total)`, making aggregate parse and compile costs directly comparable.

tests/core/benchmarks.py
vjaganat90: `wide_inference` now uses 100 unmatched steps, increasing the quadratic backward-scan component relative to per-step fixed compilation cost.

.github/workflows/benchmarks_weekly.yml
vjaganat90: The benchmark pipeline now enables `pipefail`, so the Python status crosses `tee` and the explicit `|| true` performs the intended non-gating conversion. `continue-on-error` covers failures before the pipeline.

tests/core/benchmarks.py
vjaganat90: The rationale now assigns each citation one claim: §6.2 defines the converter shape, §6.3 places speculative insertion in benchmark coverage, and §1 documents its `O(2^n)` complexity.

tests/core/benchmarks.py
vjaganat90: Corpus cases now bind paths with `functools.partial`; the static table headers are plain strings. The default-argument workaround and both associated suppressions are gone.

darkclad (commented):
Re-review at `6d854f4`. The base is unchanged (`835b941`), so `c230c23..6d854f4` is exactly the round-2 change: 4 files, +109/−40. I verified by reading, not by running. There's no sophios environment on this box, so the new tests are checked by hand-tracing (the tick arithmetic, the event order) rather than executed.

**All nine fixes hold.** Verdicts are on each original thread. Four of those threads now show as outdated because the lines moved, but every one got a reply.

| # | thread | verdict |
|---|---|---|
| 1 | tests never collected | Fixed (placement follow-up below) |
| 2 | Python-API generation missing | Fixed (failure-mode follow-up below) |
| 3 | discovery charged to first row | Fixed |
| 4 | no timeout | Fixed |
| 5 | no totals | Fixed |
| 6 | n=20 too small | Fixed |
| 7 | inert `\|\| true` | Fixed |
| 8 | §6.3 citation | Fixed |
| 9 | `partial` / suppressions | Fixed |

**Two new findings, both inline, both about where the fixes landed rather than what they do.** Neither is large. The first one, I'd fix before merge, because I caused it: the step I asked for can now turn this job red.

.github/workflows/benchmarks_weekly.yml
```diff
@@ -0,0 +1,132 @@
...
+        # NOTE: Use ".[test]" instead of ".[all_except_runner_src]"
+        # We do not want or need to install the workflow_deps extra.
+        # (Many of the packages conflict with pypy.)
+        run: cd mm-workflows/ && pip install ".[test]" && mm-workflows --generate_schemas
+
+      - name: Generate Sophios Validation Jsonschema
+        if: always()
+        run: cd sophios/ && sophios --generate_schemas
+
+      - name: Generate Sophios Python API Workflows (*.py -> *.wic)
```
darkclad, line 105: **🟠 This step can turn the non-gating job red. It needs `continue-on-error: true`, and that's on me for asking for it without saying so.**

`test_compile_python_workflows` isn't a generator that happens to live in pytest. It's a test. It collects every import or compile failure across the Python workflow scripts `get_py_paths` finds and ends with `pytest.fail("Python workflow imports failed: ...")` (`test_python_api.py:1055-1057`). Those scripts live in `image-workflows`, which this job deliberately leaves unpinned. So the week `image-workflows` main lands a broken script, this step exits non-zero. It has no `continue-on-error`, so the **job concludes failure**. The `if: always()` on later steps means the benchmarks still run, but the job goes red anyway.

That is this job gating on a Python-API correctness regression in an external repo, which is the property `lint_and_test.yml:194` already owns on every change. It's also the exact thing the job's design exists to prevent.

The fix costs nothing in coverage. The test writes each workflow's `.wic` inside the loop (`retval.write_wic(path.parent, ...)`, line 1035) and only fails after the loop. So every script that *did* generate has already put its files on disk, and the corpus is as complete as it can be that week. Only `_write_manifest` gets skipped, and this job never reads the manifest.

```yaml
      - name: Generate Sophios Python API Workflows (*.py -> *.wic)
        if: always()
        continue-on-error: true  # corpus preparation, not a check; lint_and_test owns this failure
        run: cd sophios/ && pytest tests/core/test_python_api.py -k test_compile_python_workflows
```

For completeness: the install and schema-generation steps above have no `continue-on-error` either, and I didn't raise that last round. I'd leave those alone. A failed `pip install` means there's nothing to benchmark, and a red job is the honest signal for that. This step is different in kind, because it's a correctness assertion over someone else's code.

.github/workflows/lint_and_test.yml
```diff
@@ -241,7 +241,7 @@ jobs:
       # reason the contrib step gives below.
       - name: PyTest Spec 2 Oracle
         if: always()
-        run: cd sophios/ && pytest tests/core/test_hermeticity.py tests/core/test_generators.py tests/core/test_equivalence.py tests/core/test_equivalences.py -m "not skip_pypi_ci"
+        run: cd sophios/ && pytest tests/core/test_hermeticity.py tests/core/test_generators.py tests/core/test_equivalence.py tests/core/test_equivalences.py tests/core/test_benchmarks.py -m "not skip_pypi_ci"
```
darkclad: **🟡 `test_benchmarks.py` is in the one step whose comment promises "no corpus", and it imports the corpus.**

The comment directly above this step says: *"It needs nothing this matrix lacks, which is the point of P25 — no plugins, no containers, no corpus."* It also says the step exists so that *"a red build is attributable to a zone without bisecting."*

`from . import benchmarks` executes `benchmarks.py` at collection. Line 45 is `from core.wic_corpus import CORPUS, corpus_id`, and `CORPUS` is `_discover()` evaluated at import. In this matrix `update_sophios_config.py` has written a config, so that means reading `~/wic/global_config.json` and `plugins.get_yml_paths(config)` walking every `search_paths_wic` entry. It also means `_speculative_insertion_registry()` runs at module scope. So a corpus or config problem, like a malformed `.wic` path or a missing search directory, now reds the *Oracle* step. That's the misattribution the step was split out to avoid.

Nothing actually breaks today. The poisoned run is a subprocess over the explicit `ORACLE_FILES` tuple, and the static scan is seeded from `ORACLE_MODULES`, and neither names this file. So P25 stays green. The problem is that the step's contract no longer describes its contents, and the next person to add a poison fixture at session scope, which is the natural hardening, will find this file in the way.

The fast contracts don't need the Oracle step. Default collection already reaches them via `build_wheel.yml:34`. If you want them gated on every change in `lint_and_test`, which is reasonable, give them a one-line step of their own, something like `PyTest Benchmark Harness`. Don't fold them into "Spec 0 Invariants" at line 230 instead: that's a zone step too, with the same attributability argument. A dedicated step keeps "Spec 2 Oracle" meaning what its comment says, and it makes a red benchmark-harness contract read as exactly that.

tests/core/benchmarks.py
darkclad: **Fixed.** Verified at `6d854f4`: the contracts moved to `tests/core/test_benchmarks.py`, which matches `test_*.py`, so default collection (`build_wheel.yml:34`) now picks it up, and `from . import benchmarks` resolves to `core.benchmarks` under pytest's rootdir layout. The new `test_corpus_setup_precedes_the_first_timer` and totals tests are a welcome bonus. One follow-up that is not this thread: *where* it is named explicitly. The Oracle step's own comment says it needs "no plugins, no containers, no corpus", and this file imports `wic_corpus`. Commented on `lint_and_test.yml:244`.

.github/workflows/benchmarks_weekly.yml
darkclad: **Fixed.** The generation step now runs before `Run benchmarks`, so the generated `.wic` files exist when `CORPUS` is discovered at import. It comes after `--generate_schemas` rather than before as in `fuzzy_compile_weekly`, but that order doesn't matter here because this job never builds the real validator. One follow-up that I introduced by suggesting the step: it has no `continue-on-error`, so it can now turn the job red. Commented on line 105.

tests/core/benchmarks.py
darkclad: **Fixed.** `main()` warms `_get_corpus_env()` before the loop, gated on a `corpus/` case being present, so laziness still holds for synthetic-only runs. `test_corpus_setup_precedes_the_first_timer` pins the order as `['setup', 'clock', 'run', 'clock']`, which is exactly the property. Optional and non-blocking: nothing asserts the other half, that setup is *not* called with no corpus case. `test_the_harness_reports_a_failing_case_without_failing_itself` would stay green if the gate were dropped, just slower.

.github/workflows/benchmarks_weekly.yml
darkclad: **Fixed.** `timeout-minutes: 60` on the step. Combined with the existing `continue-on-error: true`, a timeout lets the job continue and conclude green. That bounds the runner hold, which was the point. A step-level kill still loses every row after the hung case, and the totals too, but a per-case timeout was a nice-to-have in the original comment, not the ask.

tests/core/benchmarks.py
darkclad: **Fixed.** Checked the accounting: `elapsed_seconds` is now computed on the error path too, so a failing case still contributes to `all cases (total)`. `startswith('corpus/')` selects exactly the rows that the `f'corpus/{corpus_id(path)}'` names produce. `test_main_reports_corpus_and_whole_run_totals` pins 3.000 and 6.000 against a hand-computed tick sequence (1+2 for corpus, +3 for `parse_only`), and the sequence is consumed exactly: 3 cases × 2 calls = 6 ticks.

tests/core/benchmarks.py
darkclad: **Fixed.** n=100, with the docstring and the rationale both updated to match.

.github/workflows/benchmarks_weekly.yml
darkclad: **Fixed.** `set -o pipefail` sits directly above the pipeline, so Python's status now crosses `tee` and `|| true` is the mechanism doing the work. The rewritten comment describes it accurately. Because the step shell has no `-e`, the summary `echo` block before it can't abort the step either, so "`continue-on-error` also covers failures before that pipeline" is belt-and-braces, but it's true.

tests/core/benchmarks.py
darkclad: **Fixed.** Each citation now carries one claim, in both the docstring and the `speculative_insertion` rationale: §6.2 for the converter shape, §6.3 for coverage, §1 for `O(2^n)`.

tests/core/benchmarks.py
darkclad: **Fixed.** `partial(_compile_corpus_file, path)` replaces the default-argument trick, and the `type: ignore[misc]` is gone. The header `print`s are plain strings, and both pylint suppressions are gone. `import pytest` also left the module with the tests, so the runnable script no longer needs pytest installed.

.github/workflows/benchmarks_weekly.yml
vjaganat90: Added, with the reasoning on the step itself. The install and schema steps stay gating, as you suggest — if those fail there is nothing to benchmark and red is the honest signal. This one is different in kind, and the generated files are already on disk by the time the assertion fires, so the corpus is as complete as that week allows either way.

.github/workflows/lint_and_test.yml
vjaganat90: `test_benchmarks.py` now has its own `PyTest Benchmark Harness` step and is out of the Oracle's file list, so the Oracle's comment describes its contents again. I checked the parallel lists you would expect to need the same treatment: `ORACLE_FILES` and `ORACLE_MODULES` never named this file, which is why P25 stayed green, so the step list and the tuples are now consistent in excluding it.

darkclad (commented):
Round 3 at `9e53c1d`. The base is still `835b941`, so `6d854f4..9e53c1d` is exactly this round: three commits, +34/−3 across three files. I checked by reading, plus CI: `lint_and_test (ubuntu-latest)` passed at this head, including the new `PyTest Benchmark Harness` step.

**Both round-2 findings are fixed, and so is the optional note on the setup gate.** The verdicts are on each thread. No new findings.

The new test does pin the gate. `main()` calls `_get_corpus_env()` before its `try`, so if the `corpus/` check were dropped, `refuse_setup`'s `AssertionError` would escape `main()`, and the broad `except` never gets a chance to catch it. The two ticks are used up exactly, one for the start and one for the error-path elapsed time, so the `2.000` total also fails if the error path ever stops recording elapsed time.

.github/workflows/benchmarks_weekly.yml
darkclad: **Fixed** `continue-on-error: true` is on the generation step in `b409afe`, with the reasoning written on the step itself. The install and schema steps still gate, which is the right split.

.github/workflows/lint_and_test.yml
darkclad: **Fixed** `test_benchmarks.py` is out of the Oracle file list and runs in its own `PyTest Benchmark Harness` step in `9e30e53`, so the Oracle comment's "no corpus" is true again. The new step passed on `ubuntu-latest` at `9e53c1d`.

tests/core/benchmarks.py
darkclad: **Fixed** In `9e53c1d`, `test_the_harness_reports_a_failing_case_without_failing_itself` swaps in a setup that raises, so dropping the gate now fails the test instead of just slowing it down. It also pins the failing case's elapsed time into `all cases (total)`.

darkclad (approved):

PR #406, opened 2026-09-08
feat(nextflow): valueFrom self-reference, absent-optional val inputs, array-typed inputs

vjaganat90:
Stacked on #401 (`nf202_basename`). Review after that branch; its commits appear here until it merges.

## What this adds

Three related Phase 2 lowerings, each closing a construct that previously failed closed.

**`valueFrom` self-reference on a boolean binding.** CWL evaluates `valueFrom` and then applies boolean flag semantics to the result. The only shape provably boolean without a JS evaluator is a bare `$(inputs.<name>)` reference restating the binding's own input, so that's the sole supported form — it lowers to the exact same `NfFlag` token the no-`valueFrom` case produces. Every other shape (a literal, a different input, `.path`/`.basename`) stays rejected.

**Absent-optional `val` inputs.** An absent optional `string`/`int`/`float`/`boolean` value now lowers instead of rejecting, when its consuming tool port is never dereferenced in that tool's command/stream/glob templates, or is referenced solely as the boolean flag it drives (a flag tests, never dereferences). The representation is a reserved `[]` sentinel, not `null` — real execution against the pinned runtime showed `Channel.value(null)` never binds, so the consuming process hangs forever. `[]` is Groovy-falsy like `null`/`false` and can't collide with a real array value since this lowering never touches array-typed ports. Every other reference position, and `path`-qualifier or array-typed absence, still reject.

**Array-typed inputs.** `{type: array, items: <scalar>}` (File/Directory/string/int/float/boolean) lowers to a port carrying its item type's qualifier plus a new array marker. A `val` array is a JSON list through `Channel.value(...)`; a `path` array is one `Channel.value(...)` element holding a Groovy list built with `file(...)` per element, so every element stages for a single process call instead of fanning the channel out. Command-line binding supports one shape: no `itemSeparator`, prefix once when non-empty, each element its own shell-quoted argv word, nothing at all when the array is empty — runtime-proven, including the empty case, against the pinned Nextflow.

## Boundaries

- `valueFrom` aliasing a *different* boolean input, a literal, or a `.path`/`.basename` suffix on a boolean binding all still reject.
- Absent-optional support is scoped to `val`-qualifier inputs that are unreferenced-or-flag-only; a general presence-gated value binding (the value analogue of a flag, for any scalar type bound directly into the command) has no approved lowering yet. `path`-qualifier absence also stays rejected — `path` channel construction always stages unconditionally, so it needs its own sentinel convention.
- An empty array is a present, valid, zero-length value, kept structurally distinct from absence: this lowering never emits or accepts the `[]` absent-optional sentinel for an array port.
- Array support is inputs-only. Rejected and unchanged: the shorthand `File[]` type form, nested arrays, a per-item `inputBinding` on the array's `items` schema, `itemSeparator`, `separate: false` on an array binding, `valueFrom` on an array binding, and array-typed outputs.

## Schema

Adds `schema_version: 5` for the new `array` command-token kind and the additive `is_array` port field; hydration accepts versions 2-4 unchanged. `is_array` is optional on hydration (defaults `False`) since no earlier version ever wrote it — the first field-level addition alongside the existing kind-tagged additions, so `_reject_newer_kinds` gates both a `kind` and a small declared set of versioned fields now.

## Design authority

Four `design:` commits approve this in order: the `valueFrom` self-reference and absent-optional lowerings, a correction to the absent-optional sentinel after real-execution proof ruled out `null`, and the array-typed input lowering, all under §10 of `design_docs/sophios_nextflow_backend.md`.


src/sophios/utils_nf.py
```diff
@@ -1056,13 +1184,58 @@ def _workflow_capability_findings(
...
+    except (ValueError, TypeError):
+        return None
+    flag_names = {token.name for token in command.tokens if isinstance(token, NfFlag)}
+    templates = [
+        *(token for token in command.tokens if isinstance(token, NfTemplate)),
+        *(stream for stream in (command.stdin, command.stdout, command.stderr) if stream),
+        *(port.glob for port in outputs if port.glob),
+    ]
+    plain, basenamed = _template_reference_names(templates)
+    unsafe = (plain - flag_names) | basenamed
```
darkclad: **🔴 `- flag_names` excuses an input that is a flag *and* a template reference.**

Subtracting `flag_names` from `plain` says "a boolean used as a flag needs no absence handling", which is true — but the same name can also appear dereferenced in a template, and that use *does* need handling. The subtraction drops it from `unsafe` on the strength of the flag use alone.

Repro at this commit — an optional `verbose` bound as a flag and also read in `arguments`:

```yaml
inputs:
  verbose:
    type: boolean?
    inputBinding: {prefix: --verbose}
arguments: ["--label=$(inputs.verbose)"]
```

Analysis passes, and with `verbose` absent the process renders `--label=[]` — the absence sentinel leaking into a command line as a literal. The finding this line exists to raise is exactly the one it suppresses.

`unsafe = plain | basenamed` restores it. A flag-only boolean never enters `plain` in the first place (it has no template reference), so the subtraction is not carrying its own weight.

src/sophios/nf_types.py
```diff
@@ -535,6 +609,11 @@ def __post_init__(self) -> None:
                 f"process {self.name!r} basename references must target path inputs: "
                 f"{', '.join(sorted(invalid))}"
             )
+        if invalid := {name for name in array_binding_names if not is_array_by_name[name]}:
+            raise ValueError(
+                f"process {self.name!r} array bindings must reference array-marked inputs: "
```
darkclad, line 615: **🔴 The array-binding check is one-directional.**

This validates that every `NfArrayBinding` targets an array-marked port. It does not validate the converse: an array-marked port referenced by a plain `NfInputReference` passes.

That combination renders `values.toString()`, so a two-item array reaches the command line as

```
--joined=[a, b]
```

instead of expanding per item. Silent, and wrong in a way that reads like a Groovy artefact rather than a compiler bug. Same reachability through `NfStream` targets and through output `glob` templates, which take the same reference type.

The fix belongs next to the existing check: collect the plain-reference names alongside `array_binding_names` and raise when one of them lands on a port with `is_array` set. Related to the channel-contract gap I flagged on line 941.

src/sophios/nf_types.py
```diff
@@ -850,11 +933,12 @@ def _validate_graph(self) -> None:
...
                     destination = self._destination_port(process_by_name, to_process, to_port)
-                    # path_kind selects the staging policy, so it is part of the
-                    # channel contract: connection order must never pick one.
+                    # path_kind and is_array select the staging/cardinality
+                    # policy, so both are part of the channel contract:
+                    # connection order must never pick one.
                     semantics = destination.qualifier + (
                         f"[{destination.path_kind}]" if destination.path_kind else ""
-                    )
+                    ) + ("[]" if destination.is_array else "")
```
darkclad, line 956: **🔴 The new `is_array` contract is enforced for workflow inputs only.**

Adding `is_array` to `semantics` is right, and it makes `NfWorkflowInputConnection` safe. But the `NfProcessConnection` case immediately below checks only that the ports exist — nothing stops a *scalar* output feeding an array-marked destination port.

That is not hypothetical: `_array_rose_from_producer`, added by this PR, builds precisely that shape. It compiles, and the destination renders `values.isEmpty()` against a `Path`, which fails at runtime with

```
groovy.lang.MissingMethodException: No signature of method: java.nio.file.Path.isEmpty()
```

— a Nextflow-level stack trace rather than a compiler diagnostic, which is the failure mode the channel-contract comment above is written to prevent.

`_source_port` already returns the producing port, so the cardinality comparison is available right there: raise when `source.is_array != destination.is_array`.

src/sophios/utils_nf.py
```diff
@@ -144,6 +155,34 @@ def cwl_type_to_nf_qualifier(cwl_type: Any) -> str:
             raise ValueError(f"unsupported CWL type for Nextflow Phase 1: {cwl_type!r}")
 
 
+def _is_array_type(cwl_type: Any) -> bool:
+    """Return whether a (non-optional-wrapped) CWL type is the array mapping form."""
+    return isinstance(cwl_type, Mapping) and cwl_type.get("type") == "array"
+
+
+def _array_item_type(items: Any) -> Any:
```
darkclad: **🔴 The per-item `inputBinding` rejection misses the standard CWL spelling.**

`_array_item_type` only inspects `inputBinding` written *inside* `items`. CWL puts a per-item binding on the array schema itself:

```yaml
type:
  type: array
  items: string
  inputBinding: {prefix: -A}
inputBinding: {prefix: -B}
```

That form is silently accepted and lowers to `-B a b`, dropping the per-item prefix entirely; `cwltool` renders `-B -A a -A b`. The PR description states this construct is rejected, so the intent is clearly to catch it — the check is just looking one level too deep.

Checking for `inputBinding` on the array mapping alongside the existing `items` check closes it.

src/sophios/input_output_nf.py
```diff
@@ -113,10 +127,35 @@ def _render_number(value: int | float) -> str:
     return format(Decimal(str(value)), "f")
 
 
+_GLOB_METACHARACTERS = frozenset("*?[]{}")
+
+
+def _glob_names_one_file(template: Any) -> bool:
```
darkclad, line 133: **🔴 This regresses a glob that legitimately arrives as a pattern.**

`_glob_names_one_file` returns false — emitting `glob: false` — for *any* templated output name, on the reasoning that an assembled name is meant literally. That holds for a name built from a basename, which is what motivated it, but not for a glob whose pattern is itself an input:

```yaml
outputs:
  matched:
    type: File
    outputBinding: {glob: "$(inputs.pattern)"}
```

With `pattern="*.txt"` this matched before this commit and now fails with a missing output, because `glob: false` tells Nextflow to take `*.txt` literally as a filename.

The `_GLOB_METACHARACTERS` set added just above is the discriminator that is already here: the decision wants to be made against the *assembled* value, where metacharacters can be seen, rather than against the mere presence of a template — which cannot distinguish a literal-name template from a pattern-valued one.

src/sophios/nf_types.py
```diff
@@ -269,15 +290,49 @@ def to_dict(self) -> dict[str, str]:
...
+        """Return a JSON-compatible representation.
+
+        Returns:
+            dict[str, Any]: The token as ``{"kind": "array", "name": ...,
+                "prefix": ...}``.
+        """
+        return {"kind": "array", "name": self.name, "prefix": self.prefix}
+
+
+NfCommandToken = NfTemplate | NfFlag | NfArrayBinding
```
darkclad, line 322: **⚪ `NfArrayBinding` joins this union but never reaches the public surface.**

`src/sophios/api/python/nextflow.py` re-exports the token types — `NfTemplate`, `NfFlag`, `NfBasenameReference` — in both its import list and `__all__`. `NfArrayBinding` is in neither, so `from sophios.api.python.nextflow import NfArrayBinding` fails, and anyone building a command through the public API cannot name the third member of the union they will get back out of it.

Commenting here because that file has no other change in this PR, so GitHub will not take a comment on it. Alphabetically `NfArrayBinding` sorts immediately before `NfBasenameReference`, on both the import and `__all__`.

darkclad (commented):
Review at `f262003`. Scoped to this PR's own commits (`2b2b805..f262003`, ~1341 additions); the #401 basename work that shows through the stack is out of scope and I did not re-review it.

Everything below was reproduced by running the lowering and the renderer against this head in a scratch worktree (deps stubbed), not read off the diff. Six findings, all inline.

**Checked and sound:** schema-5 gating (`FIELD_SCHEMA_VERSIONS` / `_reject_newer_kinds`), optional `is_array` hydration, the `_PORT` regex change (the lazy group 2 still binds correctly with the extra optional `glob: false` group), the `[]` sentinel never reaching a `path` port (path/array absence is rejected by `_absent_optional_findings` before `_default_bindings` runs — verified both ways), array-typed outputs and the `File[]` shorthand being rejected, and `_boolean_flag_reference`'s self-reference match (the `name` guard is a comparison, not a capture).

src/sophios/utils_nf.py
vjaganat90: Fixed — `unsafe = plain | basenamed`, subtraction gone. You're right that it wasn't carrying its own weight: a flag-only boolean has no template reference so it never enters `plain`, which means the subtraction could only ever remove a name that genuinely needed handling.

Your repro is now a test with the finding pinned by exact text and path (`steps[0].run.inputs.verbose`).

src/sophios/nf_types.py
vjaganat90: Fixed, next to the existing check as you suggested. Pulled the plain-reference names out into their own set (they were being computed inline inside the `references` union) and raise when one lands on an `is_array` port.

Covered all three reachable positions in the test — command token, stream target, output glob — since they all take the same reference type.

src/sophios/nf_types.py
vjaganat90: Fixed. `_source_port`'s return value was already there for the taking, so the `NfProcessConnection` case now binds both ends and raises on `source.is_array != destination.is_array`.

Worth noting the current effect is total: process outputs can't be array-marked at all (rejected a few lines up), so this rejects every process edge into an array port. That's the correct conservative position — the only way to feed an array port today is a workflow input — and it becomes a real comparison rather than a blanket rule as soon as array-typed outputs get a lowering.

The two existing `_array_rose_from_producer` tests still pass unchanged: both reject in capability analysis before the graph is built. The new test constructs the IR directly to reach the check.

src/sophios/utils_nf.py
vjaganat90: Fixed. `_array_item_type` now takes the array mapping rather than just its `items`, so it can see both spellings; the array-level one is rejected first, and the three call sites pass the mapping. Test uses your exact `-A`/`-B` shape.

Centralising it in the one function seemed better than adding the check at each caller, since there are three and they'd have drifted.

src/sophios/input_output_nf.py
vjaganat90: You're right, and thank you — that's a real regression I introduced, with a legitimate shape behind it.

I went a different way than deciding on the assembled value, because that value only exists at run time and testing it there would mean emitting Groovy that inspects the name — computation in the renderer, and the sort of thing the projection-only rule elsewhere in this stack is meant to keep out.

There is a compile-time discriminator that draws exactly the line you describe, though: *which kind* of reference is present. A basename reference is some real file's name by construction, so it is never a pattern. A plain input reference carries arbitrary data, and a glob is a perfectly reasonable thing to pass as an input — which is your case. So the rule now requires every reference in the template to be a basename; a plain reference anywhere leaves globbing on.

`glob: $(inputs.pattern)` with `"*.txt"` is pinned by a test. Two goldens moved back as a result (`${filename}` in the real-compiled renderer test, and the backtick runtime one), which is the regression showing up where it should have.

src/sophios/nf_types.py
vjaganat90: Added to both the import list and `__all__`, sorted before `NfBasenameReference` as you noted. Extended `test_public_module_exports_flag_and_command_token` to assert it, so the union's third member can't go missing again.

Thanks for routing it here — you're right that GitHub wouldn't have taken a comment on an unchanged file.

src/sophios/nf_types.py
```diff
@@ -535,6 +610,19 @@ def __post_init__(self) -> None:
...
             )
+        if invalid := {name for name in array_binding_names if not is_array_by_name[name]}:
+            raise ValueError(
+                f"process {self.name!r} array bindings must reference array-marked inputs: "
+                f"{', '.join(sorted(invalid))}"
+            )
+        # The converse direction: a plain reference calls .toString() on its
+        # channel value, so an array-marked port reached that way renders as
+        # a Groovy list literal instead of expanding per item.
+        if invalid := {name for name in plain_reference_names if is_array_by_name[name]}:
```
darkclad: **🔴 The new guard covers `NfInputReference` but not `NfBasenameReference`.**

The comment above states the reason exactly right — a plain reference calls `.toString()` on its channel value — but a basename reference has the same problem against an array-marked port, and `plain_reference_names` does not contain it.

Reproduced end-to-end from real CWL at this head. A `File[]` input read as `$(inputs.values.basename)` lowers without complaint and renders `values.name.toString()`, which is a Groovy GPath spread over the list:

```
--tag=[a.txt, b.txt]
```

In a glob position it is worse, because it collides with the `glob: false` rule you just built:

```groovy
path "${values.name}.done", glob: false
```

— a literal filename containing a bracketed list, which can never exist, and with globbing off there is nothing to rescue it.

This is also the one case that breaks the invariant your #5 fix rests on. "A basename reference is some real file's name by construction, so it is never a pattern" is true for a scalar port and false for an array port, where it is a rendered list. Closing this restores that premise rather than just fixing a render.

`basename_names` is already computed eight lines up for the path check. Folding it into this same condition — `plain_reference_names | basename_names` — is the whole fix.

darkclad (commented):
Re-review at `1d63685`. I ran the lowering and the renderer against this head rather than reading the diff — built a venv, `pip install -e .`, and drove real CWL through each case. Full suite at head: **202 passed**. The 25 `test_runtime.py` failures are all `CreateProcess` errors from `nextflow` and a modern JDK not being on this box — identical across all 25, none at conversion — so nothing regressed there.

**Five of six fixes hold. One is half-done**, and the fix for it opened a symmetric gap on the other side of the same contract. Verdicts are on each original thread; three findings inline.

| # | thread | verdict |
|---|---|---|
| 1 | `unsafe = plain \| basenamed` | Fixed |
| 2 | array-port binding contract | **Not fixed** — basename references still slip through |
| 3 | `NfProcessConnection` cardinality | Fixed |
| 4 | `_array_item_type` array-schema binding | Fixed |
| 5 | `_glob_names_one_file` | Fixed |
| 6 | `NfArrayBinding` export | Fixed |

**On your counter-proposal for #5 — you were right and I defer to it.** Deciding on the assembled value would have meant emitting Groovy that inspects the name, which is exactly the renderer-side computation the projection-only rule in this stack exists to keep out. Requiring every reference to be a basename draws the same line at compile time, and it is a stronger rule than mine: it holds for names this compiler has never seen. I checked the two moved goldens and they are a deliberate narrowing toward cwltool, not a regression — CWL defines `outputBinding.glob` as a POSIX pattern, so leaving globbing on for a plain reference is what cwltool already does. The residual cost is that a literal name like `a[1].txt` arriving through a plain reference will silently fail to match, but cwltool fails there too, so I am not filing it.

One thing worth stating plainly about the three findings below: the first two are the same defect as my original #2 and #3, reflected. Each contract is now enforced on one side of a pair — plain references but not basename references, process edges but not workflow-input edges. The asymmetry is the pattern, not the individual line.

src/sophios/nf_types.py
```diff
@@ -850,11 +942,12 @@ def _validate_graph(self) -> None:
...
                     destination = self._destination_port(process_by_name, to_process, to_port)
-                    # path_kind selects the staging policy, so it is part of the
-                    # channel contract: connection order must never pick one.
+                    # path_kind and is_array select the staging/cardinality
+                    # policy, so both are part of the channel contract:
+                    # connection order must never pick one.
                     semantics = destination.qualifier + (
                         f"[{destination.path_kind}]" if destination.path_kind else ""
-                    )
+                    ) + ("[]" if destination.is_array else "")
```
darkclad, line 956: **🔴 The cardinality check you added for process edges is still missing here.**

`NfProcessConnection` now compares `source.is_array != destination.is_array`, which is the fix I asked for and it holds. But this case — the workflow-input edge — folds `is_array` into `semantics` only to compare *connections against each other*. Nothing checks the actual params value against `destination.is_array`.

So a scalar param feeding an array-marked port constructs cleanly, round-trips through `from_dict`, and renders `params.values.collect{...}` — which in Groovy iterates a String's characters. A `params.values = "abc"` becomes three arguments, `a`, `b`, `c`.

Note what this means together with your reply on the `NfProcessConnection` thread: you observed that process outputs cannot be array-marked at all yet, so *the only way to feed an array port today is a workflow input*. That makes this the live path — the one edge type that can actually reach an array port in a real compilation is the one still unchecked.

`self.params[from_port]` is in hand right here; `isinstance(value, list) != destination.is_array` is the comparison.

src/sophios/utils_nf.py
```diff
@@ -1112,13 +1297,23 @@ def _absent_optional_findings(
...
+                    name, qualifier = None, None
+                if safe_names is not None and qualifier == "val" and name in safe_names:
+                    continue
                 detail = (
-                    "absent optional values are deferred to Phase 2"
-                    if optional
-                    else "resolves to an absent required value"
+                    "absent optional values are supported only for a val input that is "
+                    "unreferenced in its command or drives a boolean flag"
                 )
```
darkclad, line 1314: **⚪ This message now tells the user something the same commit made false.**

"unreferenced in its command **or drives a boolean flag**" describes the old `- flag_names` rule. Your fix on line 1232 is precisely that driving a boolean flag is *not* sufficient — a flag that is also dereferenced in a template still needs handling, which is the finding this string is printed for.

So the one user who most needs this text — someone whose flag+template input just got rejected by the new rule — reads a message saying the construct they used is supported. Cosmetic in the code, actively misleading in the diagnostic.

Something like "unreferenced in its command, or drives a boolean flag and is referenced nowhere else" would match what the check now does.

src/sophios/utils_nf.py
darkclad: **Fixed.** Verified at `1d63685`: the flag+template case from my repro now raises with the finding pinned to `steps[0].run.inputs.verbose`, and a flag-only absent optional still lowers cleanly, so the rule narrowed exactly where it needed to and nowhere else. One loose end that is not this line ? the rejection *message* further down still advertises the old rule; commented there.

src/sophios/nf_types.py
darkclad: **Not fixed.** `plain_reference_names` catches plain references, but `NfBasenameReference` reaches an array-marked port by the same route and still renders `values.name.toString()` ? a `File[]` read as `$(inputs.values.basename)` lowers without complaint and produces `--tag=[a.txt, b.txt]`. Details and the end-to-end repro are in a new comment on line 621; `basename_names` is already computed eight lines above your check, so it is a one-set union.

src/sophios/nf_types.py
darkclad: **Fixed.** Confirmed both directions: base `f262003` emitted `values.isEmpty()` against a single `Path`, and this head raises on `source.is_array != destination.is_array` as intended. Your point that the effect is currently total is right and the conservative position is the correct one ? but it cuts the other way too, since if a workflow input is the only thing that can feed an array port today, that edge is the one carrying the real traffic, and it is still unchecked. Filed on line 950.

src/sophios/utils_nf.py
darkclad: **Fixed.** The `-B -A a -A b` shape is rejected at this head with and without a port-level binding, and I checked the obvious regression risk ? an ordinary `File[]`/`string[]` with a port-level `inputBinding` still lowers, because `_required_type` returns only the type field so the port-level binding can never be mistaken for the array-level one. Centralising in the one function was the right call over three call sites.

src/sophios/input_output_nf.py
darkclad: **Fixed, and your approach is better than the one I proposed ? adopt it over mine.** Deciding on the assembled value would have meant emitting Groovy that inspects the name at run time, which is the renderer-side computation the projection-only rule exists to keep out; keying on *which kind* of reference is present draws the same line at compile time and holds for names the compiler has never seen. Verified: `glob: $(inputs.pattern)` with `"*.txt"` renders `path "${pattern}", emit: matched`, basename-only still gets `glob: false`, and a mixed basename+plain template keeps globbing on. I checked the two moved goldens and agree they are a narrowing toward cwltool rather than a regression ? one caveat on the premise, though: "a basename reference is never a pattern" holds for a scalar port but not an array-marked one, which is the gap on line 621.

src/sophios/nf_types.py
darkclad: **Fixed.** `NfArrayBinding` is in both the import list and `__all__`, sorted where you said, and I checked the whole surface rather than just this member ? `__all__` now names every member of both the `NfCommandToken` and `NfTemplateSegment` unions. Extending the existing export test to assert it is the right place for that to be caught.

src/sophios/nf_types.py
vjaganat90: Fixed — `plain_reference_names | basename_names`, as you said, one union.

You're right that this is the case that breaks the premise the glob rule rests on, not just a render bug. "A basename is some real file's name" holds for a scalar port and is false for an array-marked one, where it's a rendered list. Closing it here restores the invariant rather than patching downstream, so the glob rule's justification stands as written.

Pinned in both positions — command token and output glob — since the glob one is the case that collides with `glob: false`.

src/sophios/nf_types.py
vjaganat90: Fixed, and your framing was the useful part: the one edge type that can actually reach an array port today is the one I'd left unchecked.

One wrinkle worth recording. Checking it inline in the `NfWorkflowInputConnection` arm broke `test_array_marker_participates_in_channel_qualifier_consistency`: with one param feeding both a scalar and an array port, my check fired on whichever edge came first and pre-empted the consistency diagnostic, which needs both edges seen. So the value check now runs after the loop, once every destination for a parameter agrees on a shape. Inconsistent wiring is still reported as inconsistent; only a consistent-but-wrong value reaches mine.

Also: an empty sequence is legal on both sides — it's the absent-optional sentinel on a scalar port and a real empty array on an array port — and `params` is frozen, so a JSON array arrives as a tuple rather than a list. Both are pinned.

src/sophios/utils_nf.py
vjaganat90: Fixed — now "unreferenced in its command, or drives a boolean flag and is referenced nowhere else", close to your wording.

Fair catch that it was actively misleading rather than merely stale: the person who reads it is precisely the one whose flag+template input just got rejected by the new rule.

src/sophios/nf_types.py
```diff
@@ -535,6 +610,24 @@ def __post_init__(self) -> None:
...
+            )
+        # The converse direction, for both reference kinds: each calls
+        # .toString() on its channel value -- a basename reference through a
+        # GPath spread -- so an array-marked port reached either way renders
+        # as a Groovy list literal instead of expanding per item.
+        if invalid := {
+            name
+            for name in plain_reference_names | basename_names
+            if is_array_by_name[name]
+        }:
```
darkclad, line 626: **⚪ Third reference kind: `flag_names` is still outside this set.**

`plain_reference_names | basename_names` closes the two kinds I raised. `NfFlag` is the third, and it reaches an array-marked port by the same route — nothing above rejects it, since the flag check on line 603 only constrains the qualifier to `val`, and `is_array` is orthogonal to that.

The result is quieter than the previous two and stranger. The port renders with the array marker simply dropped:

```groovy
val verbose
...
${verbose ? '--verbose' : ''}
```

Groovy list truthiness then decides the flag: a non-empty list is truthy regardless of content, so `params.verbose = [false]` emits `--verbose`. No list literal appears anywhere in the output, which is what makes it harder to spot than the `--tag=[a.txt, b.txt]` case.

**Severity is genuinely lower than the last two rounds**, and I would not hold the PR for it: it is reachable through `NfProcess.from_dict` but not through the CWL lowering, which routes an array-typed input to `_array_binding` before the flag branch is considered. So it is an IR-level gap, and the IR is the layer this class exists to validate.

`plain_reference_names | basename_names | flag_names` — the message on line 629 wants a third noun to match. Worth a glance at whether the error text should stay one sentence at that point.

For what it is worth, this is the third round in which the same contract was enforced for one reference kind at a time. `references` on line 596 already names all four kinds in one place; a guard written against that union, with `array_binding_names` subtracted, would have been total from the start and would not need a fourth visit if a fifth kind arrives.

darkclad (commented):
Third pass, at `9fa290c`. Ran the lowering and the renderer against this head rather than reading the diff. Non-runtime Nextflow suite: **204 passed**; the 25 `test_runtime.py` failures are `FileNotFoundError` on a missing `nextflow` binary, not code.

**All three round-2 findings are fixed.** Verdicts on each thread. One new finding inline, lower severity than the last two rounds — IR-level only, not reachable from CWL.

**On the cardinality restructure — I went looking for a hole and did not find one.** Your justification is sound, and there is a stronger version of it worth recording: the `semantics` key *already* encodes cardinality through `("[]" if destination.is_array else "")`, so any destination disagreement raises the consistency diagnostic **inside** the loop, before the post-loop check can run. That is what makes moving the value check out safe, rather than merely convenient. I ran the inconsistent wiring with the scalar edge first and with the array edge first; both report `feeds incompatible channel qualifiers`, so no wrong value escapes through inconsistent wiring — you get the correct diagnostic instead. Consistent-but-wrong is still caught in both directions.

I also tried to break the empty-sequence carve-out and could not: deleting it fails 4 tests, including `test_absent_optional_scalar_lowers_to_the_sentinel_parameter`. `ABSENT_VAL_SENTINEL` is literally `[]` on a scalar port, so the carve-out is required by the existing sentinel design, not a loophole. The `tuple` handling is real — `_freeze_json` maps `list|tuple` to `tuple`, verified by execution.

**`test_properties.py` checked too**, since a changed assertion deserves it. It is a real fix, not a weakening: I reproduced the old check's false positive — a literal `\` followed by a backtick renders as `` `\\`${name} ``, which contains the two-character substring the old assertion looked for while being correctly escaped. The backslash-run-parity check is the actual property, and it still catches a genuinely escaped backtick.

The new tests are load-bearing rather than tautological — deleting the production carve-out turns the cardinality test red, and the basename test covers both the command-token and glob positions.

src/sophios/nf_types.py
darkclad: **Fixed.** Verified at `9fa290c` in four positions, not just the two you pinned ? basename reference in a command token, in an output glob, and in a `stdout` template, plus a plain reference in a glob; all four raise. The reason the coverage is total is that `templates` already folds in `port.glob for port in outputs`, so widening the set was sufficient and no position was left behind. One more kind sits outside the set ? `flag_names`; commented on line 622, and it is milder than this one.

src/sophios/nf_types.py
darkclad: **Fixed**, and the restructure survives scrutiny ? I went looking for a hole rather than taking the reasoning on trust. There is a stronger justification available than the one you gave: `semantics` already encodes cardinality, so a destination disagreement raises the consistency diagnostic *inside* the loop before your post-loop check can run, which is what makes moving it out safe rather than just convenient. I ran the inconsistent case with the scalar edge first and with the array edge first ? both report `feeds incompatible channel qualifiers`, so no wrong value escapes through inconsistent wiring, and consistent-but-wrong is caught in both directions. I also tried to break the empty-sequence carve-out: deleting it fails 4 tests including `test_absent_optional_scalar_lowers_to_the_sentinel_parameter`, so it is required by the `ABSENT_VAL_SENTINEL = []` design rather than a hole, and the `tuple` handling is real (`_freeze_json` maps `list|tuple` to `tuple`, confirmed by execution).

src/sophios/utils_nf.py
darkclad: **Fixed.** The new wording matches what `_safe_absence_names` actually computes ? val and not referenced by any template ? and an `NfFlag` is not an `NfTemplate`, so a flag-only boolean genuinely never enters `unsafe`. Residual imprecision not worth a change on its own: "in its command" still understates that output globs and stream targets count as references too, though the trailing clause covers the case in practice.

src/sophios/nf_types.py
vjaganat90: Fixed in `5065ab8`.

The array-port guard now covers all three non-array-binding reference kinds (`plain_reference_names | basename_names | flag_names`), so an `NfFlag` can no longer target an array-marked input and let Groovy list truthiness decide the flag. I also changed the diagnostic to the kind-independent “array-marked inputs may only be referenced by array bindings.”

Added a regression through `NfProcess.from_dict` using the hand-crafted IR shape from this finding. Verification: executable-IR suite **47 passed**, full non-runtime Nextflow suite **205 passed**, and all pre-commit checks passed.

darkclad (approved):

PR #407, opened 2026-09-08
Spec 1 follow-on: a sequence step carries its name in an id: key

vjaganat90:
**Stacked on #408 (`semrefac_11`)** — review the chain in order.

`- touch:` in a sequence parses into a `Step` whose id is the key, but `utils.get_steps_keys` reads only `step_dict['id']`, so the step's name resolves to the empty string. The form did work: until `9758e81` ("use canonical normal form(s)", May 2024) `get_steps_keys` read a step's name with `list(step)`, which on a single-key mapping yields exactly that key, and every tutorial in the repo was written in the form. That commit made the compiler read `id:` and rewrote the tutorials; the reference was never updated, so the break went unnoticed for two years. This PR ratifies that break rather than reverting it.

## Why removing it rather than implementing it

**CWL agrees with the narrow rule, and fails the same document the same way.** `Workflow.steps` is typed as an array of `WorkflowStep` with `jsonldPredicate: {mapSubject: id}` (`v1.2/Workflow.yml`), and Schema Salad applies that transformation only *"if the value of the field is a JSON object"* (`metaschema.yml`). When `steps:` is already an array, no key is lifted into `id`. Checked against cwltool:

| | |
|---|---|
| array with `id:` | valid |
| mapping, idmap | valid |
| sequence of single-key mappings | `Field 'outputSource' references unknown identifier 's/o'` |

cwltool loses the step's identity for exactly the reason Sophios does. So this was never a Sophios divergence — it was a third form documented on top of a rule inherited from CWL.

**Implementing it would have cost a diagnostic.** In a sequence, a single-key mapping is ambiguous:

```yaml
- in: {f: 1}      # today: "a step needs an id:"   after synthesis: a step named "in"
```

A mapping key can only be a step name, so the mapping form has no such ambiguity. That asymmetry is where CWL draws the line, and §1's "leaky abstraction over CWL, deliberately" is the reason to follow it: a shorthand that accepts a shape the substrate rejects breaks at the moment a user drops into raw CWL.

## What changes

`wic006` is the one code for a sequence step with no `id:`, at any key count. The single-key case supplies its own message: it quotes the step and names both readings — "I meant a step called that" and "I forgot an `id:`" — leading with the likelier one, decided by whether the key is a known step key. Reference §3.1 is now two forms, with the CWL citation.

Both rejected shapes still report the contents of their body, so a document with two mistakes reports both in one pass instead of handing back a fresh pair on the next run.

The pinned round-trip examples move to `- id: t`. Each pins a *payload* claim — a construct inside a collection literal, a string that spells an int, NaN — and used the sequence form only as a container, so the claims are unchanged.

## Worth knowing

Files outside `search_paths_wic` use the form and therefore do not compile today: `mm-workflows/docs/tutorials/multistep3.wic` and both `insert_steps_automatically_*.wic` converters. Nothing has ever told anyone, because discovery never reached them. They are leftovers of the 2024 break — written against the language as it actually behaved — so the change there is mechanical (form 3 → form 1), but it is a sophios regression being cleaned up rather than an mm-workflows mistake.

`semrefac_11` / #408 separately fixes `insert_step_into_workflow`, which wrote this same form and is why `--insert_steps_automatically` could not work.


docs/sophios_language_reference.md
```diff
@@ -143,22 +144,36 @@ steps:
...
+transformation only *"if the value of the field is a JSON object"*. When
+`steps:` is already an array, no key is lifted into `id`, so each item is a
+plain `WorkflowStep` — and a `WorkflowStep` has no field named `touch`.
+`cwltool` fails such a document with `unknown identifier`, having lost the
+step's identity exactly as Sophios does.
+
+An earlier revision of this document listed the form as supported. It has
+never worked: the step's name resolved to the empty string in every version of
+the compiler, so no workflow can have depended on it. Writing it now earns a
+diagnostic naming the two forms above instead of a failure further downstream.
```
darkclad: **🟠 "It has never worked" is not true, and it is the only justification given for a breaking removal.**

`git show 9758e81` ("use canonical normal form(s)", 2024-05-03) changed `get_steps_keys`:

```diff
-    steps_keys = []
-    for step in steps:
-        steps_keys += list(step)
+    steps_keys = [step_dict['id'] for step_dict in steps]
```

`list(step)` on a single-key mapping yields exactly that key. Before that commit the single-key sequence form was not merely supported — it was **the only working sequence form**, and every tutorial in the repo was written in it:

```
$ git show 9758e81^:docs/tutorials/multistep1.wic
steps:
- touch:
    in:
      filename: empty.txt
...
```

That same commit rewrote them all. So the form worked for the life of the project up to May 2024; a refactor dropped it; the reference was never updated. That is a regression that went unnoticed for two years, not a documentation error.

This matters in two places beyond accuracy:

1. **The PR body calls the mm-workflows change "an mm-workflows repair."** On this history it is not a repair — those files are the leftovers of a sophios regression, written against the language as it actually behaved. Framing it as fixing someone else's mistake will read badly to whoever maintains that repo.

2. **The conclusion still holds; the argument has to change.** "Nobody can have depended on it" is false, so the removal needs the real reason — which §5.5 already states well: CWL does not admit the form, implementing it would cost the forgotten-`id:` diagnostic, and a shorthand the substrate rejects breaks the moment a user drops into raw CWL. That argument survives the correction intact, and it is the stronger one, because it explains why the 2024 break should be *ratified* rather than reverted.

The same claim is repeated at `design_docs/core-refactor-design.md:325-330` ("was listed in the reference and never worked") and in `_sequence_step`'s docstring, `src/sophios/lang/parser.py:205`. Worth a line in the reference saying the form worked before `9758e81` and is being removed deliberately rather than never having existed — that is also the note that saves the next person from "discovering" the old tutorials in git history and reopening this.

src/sophios/lang/parser.py
```diff
@@ -219,12 +223,21 @@ def _sequence_step(node: yaml.nodes.Node, file: str, diags: Diagnostics) -> Step
         return _step_body(step_id, body, span, file, diags)
 
     if len(node.value) == 1:
-        key_node, value_node = node.value[0]
-        return _step(_key_text(key_node, file, diags), value_node, file, diags)
+        # Reported rather than accepted: see the docstring. The name is read so
+        # the message can quote it, which is what tells a reader whether they
+        # meant a step called that or forgot an `id:`.
+        name = _key_text(node.value[0][0], file, diags)
```
darkclad: **🟠 This branch returns without walking the step body, so every diagnostic inside it is dropped.**

`_step_body` is what reports on the contents of a step. The `id:` path on line 223 calls it; this path does not — it reports wic021 and returns `Step(id='')`. The two spellings of the same broken content therefore report differently:

```yaml
steps:
- touch:
    in:
      f: !bogus x
    out:
      - !& e
```
→ `wic021` only.

```yaml
steps:
- id: touch
  in:
    f: !bogus x
  out:
    - !& e
```
→ `wic009` (unknown tag) + `wic019` (misplaced edge definition).

So a user who made *two* mistakes — the wrong step form plus a bad tag — fixes the form, re-runs, and is handed a fresh pair of errors. That is the edit-compile cycle `ParseResult` exists to avoid; its whole promise is that one pass reports everything it can see.

Nothing prevents the walk here. The body is `node.value[0][1]`, and the step's name is already read on line 229 for the message, so a diagnostics-only parse is available:

```python
name = _key_text(node.value[0][0], file, diags)
diags.error(Code.STEP_WITHOUT_ID, …, span)
if isinstance(node.value[0][1], yaml.nodes.MappingNode):
    _step_body(name, list(node.value[0][1].value), span, file, diags)  # diagnostics only
return Step(id='', span=span)
```

with the resulting `Step` discarded, since the document is already failing. The `isinstance` guard is what keeps `- touch:` with a null body from raising here.

The same gap exists on the `len(node.value) != 1` path at line 238 (wic006), which has the same shape and the same fix available for whichever keys are not `id`. It is pre-existing there, so I would not insist — but the two are worth being consistent about.

src/sophios/lang/parser.py
```diff
@@ -219,12 +223,21 @@ def _sequence_step(node: yaml.nodes.Node, file: str, diags: Diagnostics) -> Step
...
+        diags.error(
+            Code.STEP_WITHOUT_ID,
+            f"a step in a sequence carries its name in an id: key — write "
+            f"'- id: {name}', or key the whole steps: block by name instead (§3.1)",
```
darkclad: **🟡 The suggested fix is wrong for the reading the comment above says this message distinguishes.**

The comment on lines 226-228 says the name is quoted so the reader can tell *"whether they meant a step called that or forgot an `id:`."* The message only serves the first reading. On the second it gives an actively harmful instruction:

```yaml
steps:
- in:
    f: !ii 1
```
```
wic021: a step in a sequence carries its name in an id: key — write '- id: in',
        or key the whole steps: block by name instead (§3.1)
```

Following that produces a step *named* `in` with no inputs: `f: !ii 1` is left as the body of a step called `in`, and the input the user wrote is gone. Their actual mistake was omitting `id: <something>` above `in:` — the exact case §5.5 of the design doc singles out as the reason the form was not implemented at all.

The design doc has the analysis; the diagnostic does not carry it. Naming both readings costs one clause:

```
a step in a sequence carries its name in an id: key — if 'in' is the step's
name, write '- id: in'; if you meant to give it inputs, the step is missing
its '- id:' line above (§3.1)
```

Worth keying off whether the name is one of the known step keys (`in`, `out`, `scatter`, `when`, …) to lead with the likelier reading — `- in:` is almost certainly a forgotten `id:`, `- touch:` almost certainly a name. Those keys are already enumerated for `_step_body`, so the set is at hand.

src/sophios/lang/diagnostics.py
```diff
@@ -51,6 +51,8 @@ class Code(StrEnum):
     UNKNOWN_LANG_VERSION = 'wic017'
     LANG_VERSION_CONFLICT = 'wic018'
     MISPLACED_EDGE_DEF = 'wic019'
+    LITERAL_TYPE_MISMATCH = 'wic020'
+    STEP_WITHOUT_ID = 'wic021'
```
darkclad: **⚪ `wic006 MISSING_STEP_ID` and `wic021 STEP_WITHOUT_ID` are the same sentence twice.**

Read as names alone, nothing distinguishes them. The split is on key count and nothing else: `_sequence_step` reports wic021 when the mapping has exactly one key and wic006 when it has any other number. So

```yaml
- touch: {in: {f: x}}      # one key   → wic021
- in: {f: x}
  out: [g]                 # two keys  → wic006
```

are the same user mistake — a sequence step with no `id:` — sorted by an accident of how much else was written on it.

Diagnostic codes are declared part of the public contract, so this is awkward to revisit later; wic021 is new in this PR, which makes now the cheap moment. Two options that both read better than the current split:

- **One code, two messages.** wic006 keeps the meaning "a sequence step has no `id:`", and the single-key case supplies the more specific text plus the quoted name. Message text is not contract; the code is.
- **Keep both, rename for the real distinction** — something like `SEQUENCE_STEP_KEYED_BY_NAME` for wic021, which says what the author actually wrote rather than restating what is missing.

The first is what I would do: the fix is the same in both cases, and a user who greps the reference for wic006 should not have to discover that a sibling code covers the same mistake at a different arity.

darkclad (commented):
Review at `2e8d6d6`, diffed against the `master` merge-base (`835b941`), so #408's changes are in view as well; the sequence-step work itself is 7 files, +82/-23.

**What I ran, beyond reading the diff.** Parsed every `.wic` in this repo plus the full CI corpus — `mm-workflows/examples` and `image-workflows/workflows`, 62 files — with the new parser: **no file newly fails on wic021**, so the removal does not break corpus tests. Confirmed the renderer still emits `- id:`, so the round-trip and idempotence properties hold. Confirmed `canonicalize_steps_list` already rejected the form via `require_step_id`, so the compiler path agrees with the parser rather than diverging from it. The new `Reported` rows fire exactly one diagnostic, on the right line.

**The change itself is right.** Following CWL's line rather than a superset of it is the correct call, and §5.5's argument for why implementing the form would have cost a real diagnostic — `- in: {…}` being a forgotten `id:` rather than a step named `in` — is the strongest part of the reasoning. Four findings inline; two more below that have no line in the diff to hang on.

---

**🟠 The exported JSON Schema still describes *and* accepts the removed form.** `src/sophios/lang/schema.py:177` — `_sequence_step()`.

The file is untouched by this PR, which is the problem. Its description still reads:

> 'A step written in a sequence: either it carries id:, or it is a single-key mapping naming the step.'

and there is no `required: ['id']`, so:

```python
>>> jsonschema.Draft202012Validator(wic_schema()).iter_errors(
...     {'steps': [{'touch': {'in': {'f': 'x'}}}]})
[]
```

Zero errors on the document the parser now rejects. That breaks the contract the module states for itself in `_sequence_step`'s own docstring — *"it rejects what the parser rejects without rejecting what the parser accepts"* — in the first direction. Anyone validating with the exported schema (editor integration, a pre-commit check) gets a green light on a document that fails to compile.

`'required': ['id']` on the `sequenceStep` object, the description reworded, and a row in `STRUCTURAL_VIOLATIONS` (`tests/core/test_lang_schema.py:139`) so the two directions stay pinned together.

---

**⚪ Stale "single name key" wording outside the diff.** The reference's §4.2 (`docs/sophios_language_reference.md:287`), `parser.py:284`, `ast_strategies.py:468`, and `test_compiler.py:14` ("§3.1's third step surface form") all still describe or name the removed form. The last is the one I would fix first: a test whose name points at a section that no longer has a third form reads as a stale test rather than a regression guard.

docs/sophios_language_reference.md
vjaganat90: You are right, and the converse is worth recording too: `list(step)` on `{'id': …, 'in': …, 'out': …}` yields `['id', 'in', 'out']`, so before `9758e81` the `id:` form was the broken one and the single-key form was the only working sequence spelling.

Corrected at all three sites — the reference, §5.5 of the design doc, and the `_sequence_step` docstring. The removal now rests on the CWL argument and explicitly ratifies the 2024 break rather than claiming a history that did not happen. I have also reworded the PR body: those mm-workflows files are leftovers of a sophios regression, not that repo's mistake to own.

src/sophios/lang/parser.py
vjaganat90: Both paths walk the body now, through one `_rejected_step` so they cannot drift apart later — I took the pre-existing multi-key site as well, since leaving it is how this comes back. Tests pin the exact ordered `(code, line)` list for both rejected shapes against the `id:` form baseline.

One correction to your repro: I used a misplaced edge definition inside `in:` rather than `out: - !& e`, which turns out not to be diagnosed.

src/sophios/lang/parser.py
vjaganat90: Message names both readings and leads with the likelier one, decided by whether the key is a known step key. The set comes from the existing enumeration `_step_body` already uses rather than a second hand-maintained list, and a test pins that every member of it is one `_step_body` actually acts on, so the two cannot silently disagree.

src/sophios/lang/diagnostics.py
vjaganat90: Took your first option: `wic006` is now the one code for a sequence step with no `id:`, at any key count, and the single-key case supplies the more specific message with the quoted name. `wic021` is gone from the enum, with a line in its place saying the number is retired rather than free, since it did ship in a review cycle.

The message text now carries the whole distinction, so there is a test asserting both parses report exactly `[MISSING_STEP_ID]` and pinning each message in full.

docs/sophios_language_reference.md
darkclad: **Fixed**, verified at `f6c258b` in all three places — the reference, §5.5 of the design doc, and `_sequence_step`'s docstring now say the form worked until `9758e81` and ratify the break on the CWL argument. The PR body is rewritten too, and the mm-workflows paragraph now puts the regression where it belongs. Thanks for the converse about `list(step)` on the `id:` form; that makes the history cleaner than I had it.

One wording slip in the new reference paragraph, inline on line 174 — it reads as if the tutorials were rewritten *into* the rejected form.

src/sophios/lang/parser.py
darkclad: **Fixed** for the shape I raised, and `_rejected_step` shared by both paths is the right structure — taking the multi-key site too is what keeps it from coming back.

**You are right about my repro**, and I should have caught it: `out: - !& e` is exactly where an edge definition belongs, so there was nothing there to diagnose. Your `g: !& e` under `in:` is the correct version.

One case the fix does not reach, inline on line 253: when the single key is itself a step key (`- in:`), the inner mapping is walked as the step body instead of the entry, which drops `in:`/`out:` shape errors and invents a wic010 for an input named `id`. The fixtures here use `touch`, so they take the other branch.

src/sophios/lang/parser.py
darkclad: **Fixed.** Leading with the likelier reading, decided by `Grammar.STEP_KEYS`, is better than what I suggested, and deriving the set from `INTERPRETED_STEP_KEYS` rather than restating it is the right call. Checked `- in:`, `- out:`, `- run:` and `- touch:` at `f6c258b`; each leads with the reading I would expect.

The same classification ought to decide which node gets walked, not only which sentence comes first — inline on line 253.

src/sophios/lang/diagnostics.py
darkclad: **Fixed**, and "retired, not free" is a good call — a number that shipped in a review cycle should not come back meaning something else. Checked at `f6c258b`: no reference to `wic021` or `STEP_WITHOUT_ID` is left anywhere in `src/`, `tests/`, `docs/` or `design_docs/` apart from the retirement comment.

src/sophios/lang/parser.py
```diff
@@ -219,14 +236,49 @@ def _sequence_step(node: yaml.nodes.Node, file: str, diags: Diagnostics) -> Step
...
+        named = f"write '- id: {name}' if {name!r} is the step's name"
+        forgotten = f"add the '- id:' line above if {name!r} is one of the step's own keys"
+        first, second = (forgotten, named) if name in Grammar.STEP_KEYS else (named, forgotten)
+        diags.error(
+            Code.MISSING_STEP_ID,
+            f'a step in a sequence carries its name in an id: key — {first}; {second}. '
+            f'Keying the whole steps: block by name is the other form (§3.1)',
+            span,
+        )
+        return _rejected_step(_entries(body), span, file, diags)
```
darkclad: **🟡 Line 246 decides which reading is likelier, but line 253 walks the body the same way for both — so the likelier reading of `- in:` is walked through the wrong node.**

When the key is a step key, the reading the message leads with is "forgotten `id:`", which makes the *entry* the body: `in:` is the step's `in:`. `_entries(body)` walks the value instead, so the contents of `in:` are treated as step keys. At `f6c258b`, against the `id:` form holding the same content:

| entry | `- id: s` + same content | this head |
|---|---|---|
| `- in: [a, b]` | wic003 `in: must be a mapping` | wic006 only |
| `- out: {a: 1}` | wic004 `out: must be a sequence` | wic006 only |
| `- in: {id: !ii 1}` | no diagnostics | wic006 + **wic010** `step '' already has its identity` |

The first two are the same gap the round-1 thread was about, for the other reading: fix the form, re-run, get a fresh error. The third is **a false diagnostic** — an input named `id` is legal, and here it is reported as a contradictory second identity for a step whose name is quoted as `''`.

Tag errors are not affected (`!bogus`, `!&` under `- in:` still surface, because the inner keys fall through to passthrough and `_opaque` still checks tags), which is why the new tests do not see this: both `REJECTED_ENTRIES` fixtures reach this line with `touch`, which is on the other branch.

The classification already exists on line 246, so the fix is to use it for the walk too, and to pass the name through for the other reading:

```python
        if name in Grammar.STEP_KEYS:
            return _rejected_step('', list(node.value), span, file, diags)
        return _rejected_step(name, _entries(body), span, file, diags)
```

with `_rejected_step` taking `step_id` and handing it to `_step_body` (the multi-key call on line 260 passes `''`). I ran that patch against this head:

```
- in: [a, b]                   [wic006 L2, wic003 L2]
- out: {a: 1}                  [wic006 L2, wic004 L2]
- in: {id: !ii 1}              [wic006 L2]
- touch: {id: t}               [wic006 L2, wic010 L3]  step 'touch' already has its identity
REJECTED_ENTRIES single-key    [wic006 L2, wic009 L4, wic019 L5]   unchanged
REJECTED_ENTRIES multi-key     [wic006 L2, wic009 L4, wic019 L5]   unchanged
```

A third `REJECTED_ENTRIES` row with a step-key name — `'steps:\n- in: [a, b]\n'` against `- id: s\n  in: [a, b]` — pins it. The wic010 text quoting `'touch'` rather than `''` matches what the mapping form already says for the same body.

docs/sophios_language_reference.md
```diff
@@ -143,22 +144,40 @@ steps:
...
+transformation only *"if the value of the field is a JSON object"*. When
+`steps:` is already an array, no key is lifted into `id`, so each item is a
+plain `WorkflowStep` — and a `WorkflowStep` has no field named `touch`.
+`cwltool` fails such a document with `unknown identifier`, having lost the
+step's identity exactly as Sophios does.
+
+This document listed the form as supported, and it was: until the May 2024
+normal-form refactor (`9758e81`), which made the compiler read a step's name
+from `id:` and rewrote every tutorial into the form above, it was the only
+sequence form that worked. That break went unnoticed and is now ratified
```
darkclad, line 175: **⚪ "rewrote every tutorial into the form above" reads backwards.**

The nearest form above this paragraph is the rejected `- touch:` example, so the sentence says `9758e81` moved the tutorials *into* the single-key form — the opposite of what happened, and of what your design doc says at `core-refactor-design.md:327` ("rewrote every tutorial *out of* it").

"rewrote every tutorial out of it" or "into the `id:` form" fixes it.

tests/core/test_compiler.py
```diff
@@ -0,0 +1,26 @@
...
+from sophios.compiler import insert_step_into_workflow
+from sophios.wic_types import StepId, Tool, Yaml
+
+
+@pytest.mark.fast
+def test_an_inserted_step_is_resolvable_by_id() -> None:
+    """`insert_step_into_workflow` must write a step `get_steps_keys` can read back.
+
+    It used to insert `{stepid.stem: None}` - reference §3.1's third step
+    surface form, a single-key mapping whose key is the step name - but
```
darkclad: **⚪ "reference §3.1's third step surface form" — §3.1 has two, as of this PR.**

This docstring lands in #408 and this PR removes what it points at, so it goes stale in the same stack. Something like *"a single-key mapping whose key is the step name — a form the reference no longer admits (§3.1)"* keeps the history without citing a section that no longer says it.

darkclad (commented):
Second pass, at `f6c258b`. I ran the parser and the schema against this head directly rather than reading the diff; `package` is green on it too.

**All four round-1 threads are fixed** — verdicts on each. One slip in my own repro, acknowledged on the parser thread: `out: - !& e` is valid, so it was never going to be diagnosed.

Two new findings inline, both small. The one worth a change is on `parser.py:253`: the classification that orders the wic006 message is not applied to the body walk, so the likelier reading of `- in:` is walked through the wrong node.

---

**Carried over from round 1, not yet addressed.** These were in the review body with no line in the diff, so there was no thread to reply on — easy to miss, which is why I am repeating them here.

**🟠 The exported JSON Schema still describes and accepts the removed form.** `src/sophios/lang/schema.py:187-194`, unchanged at this head:

```python
>>> list(Draft202012Validator(wic_schema()).iter_errors({'steps': [{'touch': {'in': {'f': 'x'}}}]}))
[]
>>> wic_schema()['$defs']['sequenceStep']['description']
'A step written in a sequence: either it carries id:, or it is a single-key mapping naming the step.'
>>> wic_schema()['$defs']['sequenceStep'].get('required')
None
```

Zero errors on the document the parser now rejects, against `_sequence_step`'s own docstring: *"it rejects what the parser rejects without rejecting what the parser accepts."* Fix is `'required': ['id']`, the description reworded, and a row in `STRUCTURAL_VIOLATIONS` (`tests/core/test_lang_schema.py:139`).

**⚪ Wording that still describes the removed form**, outside the diff:

- `docs/sophios_language_reference.md:290-291` (§4.2) — *"when its identity already comes from a mapping key or a single name key."*
- `src/sophios/lang/parser.py:323` — `_step_body`'s comment lists *"the single name key"* as a source of identity. After this PR nothing reaches `_step_body` with one.
- `tests/core/ast_strategies.py:467-470` — now **wrong, not just stale**: *"A single-key sequence step takes its key as the id … earns MISSING_STEP_ID only once it has more than one key."* A single-key step earns MISSING_STEP_ID as of this PR. The `- {}` row below it is still correct; only the comment explaining it is inverted.

`test_compiler.py:13-14` is the fourth; that one is in the diff, so it is inline.

docs/sophios_language_reference.md
vjaganat90: Corrected at all three sites — the reference, §5.5 of the design doc, and the `_sequence_step` docstring. The removal now rests on the CWL argument and explicitly ratifies the 2024 break. Worth recording the converse you implied: `list(step)` on `{'id': …, 'in': …}` yields `['id','in','out']`, so before `9758e81` the `id:` form was the broken one. PR body reworded too — those mm-workflows files are leftovers of a sophios regression, not that repo's mistake.

src/sophios/lang/parser.py
vjaganat90: Both paths walk the body now, through one `_rejected_step`, and I took the pre-existing multi-key site as well rather than leaving the asymmetry. Tests pin the exact ordered `(code, line)` list for both rejected shapes against the `id:` form baseline. One correction to the repro: I used a misplaced edge definition inside `in:` rather than `out: - !& e`, which turns out not to be diagnosed.

src/sophios/lang/parser.py
vjaganat90: One addition to my reply above. The test I said pins `Grammar.STEP_KEYS` against what `_step_body` consumes was, as first written, vacuous: it built its probe body *from the set under test*, so any member dropped from the set vanished from the probe too and the test stayed green. Dropping `out` and dropping `run` both passed. It now probes a fixed key list, and all three drift directions fail. Had it stayed vacuous, `- out:` drifting out of the set would have been read as a step *named* `out`, and wic006 would have led with the wrong advice.

src/sophios/lang/diagnostics.py
vjaganat90: Took your first option. `wic006` is the one code for a sequence step with no `id:` at any key count, and the single-key case supplies the more specific message with the quoted name. `wic021` is gone from the enum with a line in its place saying the number is retired rather than free, since it did ship in a review cycle. A test asserts both parses report exactly `[MISSING_STEP_ID]` and pins each message in full, since the text now carries the whole distinction.

src/sophios/lang/parser.py
vjaganat90: Fixed, and your reading of it was exactly right — the classification on the line above decides which node holds the body, and the walk was ignoring that. It now passes `list(node.value)` under the forgotten-`id:` reading and `_entries(body)` under the named reading, with `_rejected_step` taking the step id so the named arm quotes it.

I reproduced your table before and after. `- in: [a, b]` and `- out: {a: 1}` now report wic006 plus the wic003/wic004 the `id:` form gives them, and `- in: {id: !ii 1}` reports wic006 alone — the false second-identity diagnostic is gone. `- touch: {id: t}` quotes `'touch'` rather than `''`.

Rather than a single row I added a parametrised test that compares each spelling against the `id:` form holding the same content and asserts the rejected form owes exactly the baseline plus its wic006, so the two sides cannot drift; it also asserts wic010 is absent, which is the specific false diagnostic.

docs/sophios_language_reference.md
vjaganat90: Fixed — reads "rewrote every tutorial out of it into the `id:` form" now, which also matches the design doc's wording.

tests/core/test_compiler.py
vjaganat90: Reworded to your suggestion: a single-key mapping whose key is the step name, a form the reference no longer admits (§3.1). Keeps the history without citing a section that no longer says it.

src/sophios/lang/parser.py
darkclad: **Verified** the addendum at `f1893f2`, by mutation rather than by reading. Running the body of `test_step_keys_is_exactly_the_set_step_body_acts_on` against a patched `Grammar`:

```
unmodified                                           passes
'out' dropped from STEP_KEYS                         fails
'run' dropped from STEP_KEYS only                    fails
'doc' added to STEP_KEYS (body passes it through)    fails
'hints' added to STEP_KEYS (not in the probe)        fails
'run' dropped from INTERPRETED (both sides move)     fails
```

The last row is the one that matters most: before, dropping `run` from `INTERPRETED_STEP_KEYS` moved both sides together and nothing noticed. A fixed probe is the right fix, and a good catch to have made on your own test.

src/sophios/lang/parser.py
darkclad: **Fixed**, verified at `f1893f2`. `- in: [a, b]` → wic006 + wic003, `- out: {a: 1}` → wic006 + wic004, `- in: {id: !ii 1}` → wic006 alone, and `- touch: {id: t}` quotes `'touch'`. Comparing each spelling against the `id:` form holding the same content is a better test than the row I suggested — it cannot drift.

The mirror case on the other branch has the same gap, for a non-mapping value under the *named* reading (`- touch: [a, b]`) — inline on line 272 in this round's review.

docs/sophios_language_reference.md
darkclad: **Fixed**, verified at `f1893f2` — it now agrees with `core-refactor-design.md:327`.

tests/core/test_compiler.py
darkclad: **Fixed**, verified at `f1893f2`.

src/sophios/lang/parser.py
```diff
@@ -197,36 +205,104 @@ def _steps(node: yaml.nodes.Node, file: str, diags: Diagnostics) -> tuple[tuple[
...
+        )
+        # The reading decided above also decides which node holds the body. Under
+        # the forgotten-`id:` reading the entry *is* the body — `in:` is the step's
+        # own `in:` — so walking the key's value instead would read that mapping's
+        # keys as step keys: `- in: {id: x}` would report a contradictory second
+        # identity for an input legitimately named `id`, and `- in: [a, b]` would
+        # lose the wic003 the `id:` form gives it.
+        if name in Grammar.STEP_KEYS:
+            return _rejected_step('', list(node.value), span, file, diags)
+        return _rejected_step(name, _entries(body_node), span, file, diags)
```
darkclad, line 272: **⚪ Under the named reading, a non-mapping value is dropped silently — so the mapping form the message recommends hands back an error the sequence spelling did not report.**

`_entries(body_node)` turns anything that is not a mapping into an empty body. At `f1893f2`, against the mapping form holding the same value:

| entry | this spelling | `touch: …` (mapping form) |
|---|---|---|
| `- touch: [a, b]` | wic006 | **wic003** `step 'touch' must be a mapping, found a sequence` |
| `- touch: 3` | wic006 | **wic003** `…found a scalar` |
| `- touch:` (null) | wic006 | nothing — consistent |
| `- touch: {in: {f: !bogus x}}` | wic006 + wic009 | wic009 — consistent |

So a user who follows the message's own advice ("key the whole steps: block by name") re-runs and gets a new wic003. It is the mirror of the round-2 finding on the other branch: `_rejected_step`'s docstring promises that fixing the form does not hand back a fresh round of errors, and `test_a_forgotten_id_entry_is_walked_as_its_own_body` pins that promise for step-key names only.

The mapping form already has the right check in `_step` — null is a legal empty body, anything else that is not a mapping is wic003 — and this line called `_step` before the PR. Calling it again for its diagnostics closes the gap:

```python
        if name in Grammar.STEP_KEYS:
            return _rejected_step('', list(node.value), span, file, diags)
        _step(name, body_node, file, diags)  # diagnostics only, exactly as the mapping form
        return Step(id='', span=span)
```

Ran it at this head:

```
- touch: [a, b]              [wic006 L2, wic003 L2]   = mapping form + wic006
- touch: 3                   [wic006 L2, wic003 L2]   = mapping form + wic006
- touch:                     [wic006 L2]              = mapping form + wic006
- touch: {in: {f: !bogus x}} [wic006 L2, wic009 L4]   = mapping form + wic006
- touch: {id: t}             [wic006 L2, wic010 L3]   = mapping form + wic006
REJECTED_ENTRIES, both rows  [wic006 L2, wic009 L4, wic019 L5]   unchanged
FORGOTTEN_ID_ENTRIES, all 3  unchanged
repeated id:                 [wic010 L3]              unchanged
```

`_entries` has no other caller after that, so it can go. A `NAMED_ENTRIES` table mirroring `FORGOTTEN_ID_ENTRIES` — each row paired with the mapping form rather than the `id:` form, asserting `[MISSING_STEP_ID, *baseline]` — would pin both branches the same way.

src/sophios/lang/parser.py
```diff
@@ -197,36 +205,104 @@ def _steps(node: yaml.nodes.Node, file: str, diags: Diagnostics) -> tuple[tuple[
...
         return _step_body(step_id, body, span, file, diags)
 
     if len(node.value) == 1:
-        key_node, value_node = node.value[0]
-        return _step(_key_text(key_node, file, diags), value_node, file, diags)
+        # The same defect as below — a sequence step with no `id:` — so the same
+        # code. One key is the only case where a name can be quoted, so it earns
+        # the more specific text: both readings the key admits, likelier first.
+        _key_node, name, body_node = keyed[0]
+        named = f"write '- id: {name}' if {name!r} is the step's name"
```
darkclad, line 255: **⚪ Nit: a non-scalar key is quoted as a node repr.**

`_key_text` stringifies a collection key for recovery after reporting wic005, and this branch then puts that string in front of the user three times:

```yaml
steps:
- ? [a]
  : 1
```
```
wic005: mapping keys must be scalars, found a sequence
wic006: a step in a sequence carries its name in an id: key — write
        '- id: [ScalarNode(tag='tag:yaml.org,2002:str', value='a')]' if
        "[ScalarNode(tag='tag:yaml.org,2002:str', value='a')]" is the step's name; …
```

`_key_text`'s own docstring says the fallback exists so a key does not end up "silently becoming the repr of a node list" — the recovery value was never meant to be quoted. Only a scalar key is a name the message can offer back, so gating the branch on it lets everything else fall through to the plain message below:

```python
    if len(node.value) == 1 and isinstance(node.value[0][0], yaml.nodes.ScalarNode):
```

At this head that gives `wic005` + `wic006 'a step in a sequence needs an id:'`, and every pinned message in `test_lang_parser.py` is unchanged. Cosmetic — wic005 on the same line already tells the user what is wrong.

darkclad (commented):
Fourth pass, at `f1893f2` — a full re-review of the PR against `master` (`130de95`, with #408 now merged), not just a delta check. `package` is green on this head. Every claim below comes from running the parser at this head; the test files still do not collect locally for want of `cwltool`.

**On your two comments.** Thanks for picking up the body items — and for saying plainly how they were missed; that makes it easy to fix the process rather than argue about it.

- **Schema — fixed.** Checked in both directions, as you did: the removed form, `{'touch': None}`, and `[{'in': {...}}]` each give 1 error; the `id:` form, bare `id:`, and the mapping form with and without a body give 0. The three `STRUCTURAL_VIOLATIONS` rows pin it.
- **§4.2 and `_step_body`'s comment — fixed.** Your rewording of the comment is more accurate than a plain deletion would have been, since `_rejected_step` now does hand `_step_body` a name. The grep for the three phrases is empty across `src/`, `docs/` and `tests/`.
- **`mypy` rebinding and the tutorials — confirmed.** All 21 `steps:` blocks under `docs/` parse, apart from the deliberate rejected example in §3.1 and the annotated key skeleton in §2 (which is on `master` too). The seven captioned tutorial listings are YAML-identical to the `.wic` files they name.
- **`f1893f2`, repeated `id:` — good addition.** One wic010 per repeat on its own line (`id: a / id: b / id: c` → lines 4 and 5), no second contradictory-identity report stacked on it, and the id path now reads each key once rather than twice.

**This round:** two small findings inline, both in the single-key branch and both with a patch I ran. Neither is worth holding the PR for.

---

**Outside the diff, for follow-ups rather than this PR.** Both reproduce identically on `master`.

**`src/sophios/cwl_subinterpreter.py:87` and `:97` still build the removed form:**

```python
root_yaml_tree = {'wic': wic_steps, 'steps': [{cwl_tool: None}]}
...
yml = {'steps': [{cwl_tool: args_vals_new}]}
```

That is the last producer of `[{name: body}]` in the tree — the same defect #408 fixed in `insert_step_into_workflow`. It already fails in `utils.require_step_id` on `master`, so this PR does not break it, but after this PR the shape is a documented error, which makes it worth a ticket. The fix is the #408 one: `[{'id': cwl_tool, **args}]`.

**Unknown tags on the step structure itself are not reported.** The parser checks tags on keys and inside `in:` values, but not on the nodes that make up the structure:

```
steps: !foo {}             (no diagnostics)
- !foo {id: s}             (no diagnostics)
in: !foo {f: !ii 1}        (no diagnostics)
touch: !foo {in: {}}       (no diagnostics)
```

The loader rejects all four, so this is the parser accepting documents the loader refuses — the one direction `_key_text`'s docstring says the language never takes. Pre-existing and out of scope here; worth its own issue and a `_reject_unknown_tag` call at each structural node.

vjaganat90:
Both inline findings fixed, and you were right that they are the same defect as the round-2 one wearing the other branch's clothes.

**The named reading's body walk.** `_step(name, body_node, file, diags)` now, exactly as you patched it, with the resulting step discarded. Every spelling in your table matches the mapping form plus its own wic006: `[a, b]` and `3` both give wic006 + wic003, a null body gives wic006 alone, the bad tag and the second identity are unchanged. `_entries` had no other caller and is gone.

That leaves one helper doing the walk on each branch and the classification on the line above choosing which node to hand it — which is the shape the round-2 fix should have had in the first place. The promise in `_rejected_step`'s docstring is now pinned on both sides: `FORGOTTEN_ID_ENTRIES` against the `id:` form, and a new `NAMED_ENTRIES` against the mapping form, each asserting the rejected spelling owes exactly its baseline plus wic006. Reverting the `_step` call fails four of the five new rows.

**The node repr.** Branch gated on a scalar key. `- ? [a] : 1` now gives wic005 plus the plain `a step in a sequence needs an id:`, and no pinned message elsewhere moved. You are right that it is cosmetic — but the recovery value leaking into a message three times is the kind of thing that reads as a compiler bug to whoever hits it.

All three now live in one commit rather than three, since they are one statement: a sequence step's identity and body are read the way the mapping form reads them.

**On the two outside the diff** — agreed they are follow-ups, and I have not touched either here. `cwl_subinterpreter.py` building `[{name: body}]` is the same shape #408 fixed in `insert_step_into_workflow` and wants the same `[{'id': name, **args}]`. The unreported structural tags are the more interesting one: the parser accepting what the loader refuses is the direction `_key_text`'s docstring says the language never takes, and `_reject_unknown_tag` at each structural node is a contained fix. I will raise both as issues rather than growing this PR.

vjaganat90:
The two items you flagged outside the diff are done in #415, not here — `cwl_subinterpreter`'s two single-key sequence steps, and `_reject_unknown_tag` on the four structural positions that never got a call.

#415 stacks on #412 and is the sweep-up branch for exactly this: defects the chain surfaces that no PR in it owns. Both of these reproduce on `master` and belong to subjects none of #407–#412 is about, so folding either into this PR would have made its claim harder to read rather than more complete. Anything else of that shape that comes up while the chain is reviewed will go there too, rather than growing a PR you have already been through four times.

Worth saying which way I nearly went: the subinterpreter one has a real argument for living here, since a PR that says the language has exactly two step forms is arguably incomplete while the tree still produces the third — the same two-authorities-disagree shape as the exported schema. What decided it was that `require_step_id` already refuses those documents, so there is no behaviour this PR changes and nothing for a reviewer of *this* PR to check.

darkclad (approved):
LGTM

PR #408, opened 2026-09-09
Four compiler defects found by the Spec 2 oracle

vjaganat90:
**Stacked on #405** — review the chain in order. Four compiler defects, each found by the Spec 2 oracle in the branches below, none of them fixed there because that spec touches `src/` only where a property cannot otherwise be stated.

## 1. A falsy default is not the absence of a default

`_arg_has_default_or_is_optional` tested `in_tool[arg].get('default')` for *truthiness*, so an input declaring `default: 0`, `default: False` or `default: ''` was classed as having no default, hence required. The tool author's default was discarded and the value promoted to a workflow-level input the caller must supply.

Live in `mm-workflows/cwl_adapters/extract_pdbbind_refined.cwl` and `generate_conformers.cwl`, both declaring `convert_Kd_dG: {type: boolean, default: False}`. Eight corpus workflows use those tools; four bind the input explicitly and are unaffected, four do not: `docking_rescoring_weekly`, `run_diffdock_weekly`, `vs_demo_4_weekly`, `flc`.

Silently wrong rather than loudly broken — those four compile, carrying an input that should not exist, with `convert_Kd_dG` taking whatever the inputs file says instead of `False`.

**Fix:** a default satisfies an input when it is present and not null. `null` is the exception because it is the one value a non-nullable input provably cannot take; if the type does permit null, the optional arms already say so.

This predicate also gates edge inference, through `args_required`, so the two questions it answers are worth naming: "must the caller supply this" and "should inference try to fill it". One predicate keeps them from ever diverging for the same input. `required_inputs_of` in the synthetic mirror already read presence this way, so the compiler now agrees with it rather than differing; `sink.extras` (`default: []`) is the stub that exercises the point where truthiness and presence give different answers, which previously nothing did.

## 2. An uncoercible literal escaped as a bare `ValueError`

`populate_scalar_val` coerced with a naked `int(value)` / `float(value)`. `!ii` places no constraint relating a literal to the declared CWL type of the input it binds — nothing in the grammar could — so `in: {n: !ii _}` against a tool declaring `n: int` raised `ValueError: invalid literal for int() with base 10: '_'`. No code, no span, no file.

That is the failure shape CR-104 removed everywhere it looked; this site survived because it was never a `sys.exit`.

**Fix:** `wic020`, naming the input, its declared type, and the offending literal — the user's next move is to change one of those three. It catches `TypeError` alongside `ValueError`: `int()` raises `ValueError` only for a string that will not parse, and a wrong-shaped argument raises `TypeError`. `!ii [1, null]` against `int[]` is the sharp case — the null guard inspects only the top-level value, so the array branch coerced `None` one element down and the missing-input diagnostic leaked back out as a traceback.

With a diagnosis in place, an ill-typed `!ii` literal is ordinary expected behaviour for a document that binds `"abc"` to an `int`, not a declared residual defect; the docstrings that recorded it as one now say so.

## 3. A bare-string `format` made the insertion search silently never match

The whitelist search flattens each candidate tool's input formats with `utils.flatten`, which iterates a bare string character by character. A converter declaring `format: 'edam:format_2330'` rather than `format: ['edam:format_2330']` contributed single characters to the candidate set and never matched — no diagnostic, the search simply found nothing.

The consumer input was the same defect on the same pass, and worse in one respect: it stored the bare string into `in_dict['format']`, where the ordinary backward match tests `out_format in in_formats`. `in` on a `str` is substring containment, so an output declaring `format_123` matched an input declaring `format_1234` — an edge wired wrongly rather than one missed.

**Fix:** one `declared_formats` used at every point of entry — the consumer input, the whitelist candidate's inputs, and its outputs — so the two readings cannot diverge again.

## 4. Inserted steps were unreadable

`insert_step_into_workflow` wrote `{stepid.stem: None}` — a single-key mapping — while `utils.get_steps_keys` reads only `step_dict['id']`, so every inserted step's name came back as the empty string. It also runs *after* `desugar_into_canonical_normal_form`, so nothing normalised it later. This is why `--insert_steps_automatically` could not work.

**Fix:** write `{'id': stepid.stem}`, the form the rest of the compiler reads.

Deliberately **not** fixed by teaching canonicalisation to synthesise `id` from a single-key mapping. In a sequence that shape is ambiguous — `- in: {…}` is a forgotten step id, and synthesising would silently turn it into a step named `in`. #407 removes the form from the language instead, on CWL's authority; this fix stands on its own either way.

---

Each fix has a regression test that fails when that fix alone is reverted.


src/sophios/compiler.py
```diff
@@ -1167,10 +1167,22 @@ def populate_scalar_val(cwl_type: Any, value: Any, fmt: Any = None) -> Any:
...
 
             case "int":
-                return int(value)
+                try:
+                    return int(value)
+                except ValueError as e:
+                    raise SophiosError.error(
+                        Code.LITERAL_TYPE_MISMATCH,
+                        f"Input {key!r} is declared type 'int' but its literal {value!r} "
+                        "does not convert to it.") from e
```
darkclad, line 1183: **🔴 This turns `Lint And Test` red — `test_generators.py` still catches `ValueError`.**

`SophiosError` is `class SophiosError(Exception)` (diagnostics.py:132). It is not a `ValueError` subclass, so nothing that was catching the old `int(value)` failure catches the new one.

`tests/core/test_generators.py:160-163`, inside `test_the_workflows_strategy_produces_documents_the_compiler_accepts`:

```python
try:
    compile_hermetic_cwl(yml, 'oracle')
except ValueError:  # the declared `!ii` residual; see the docstring
    return
```

That `except` exists specifically for this raise site — the test's own docstring says so:

> the module docstring's pending finding — `!ii` puts no constraint relating a literal to the CWL type it binds, so `populate_scalar_val` raises a bare `ValueError` — is a real and declared residual (measured at about one document in ten)

One in ten, at `@settings(max_examples=50)`, is roughly five escaping draws per run. This is not a flake; it is a near-certain failure. And `.github/workflows/lint_and_test.yml:244` runs the file:

```
pytest tests/core/test_hermeticity.py tests/core/test_generators.py tests/core/test_equivalence.py tests/core/test_equivalences.py -m "not skip_pypi_ci"
```

`-m "not skip_pypi_ci"` does not exclude `slow`, so the test runs.

The fix is `except SophiosError` plus rewording the docstring, and the rewording is the interesting half: the residual is no longer "raises a bare `ValueError`", it is "raises a diagnosed `LITERAL_TYPE_MISMATCH`", which is a much weaker complaint. That residual arguably stops being a *declared defect* at all with this PR and becomes ordinary expected behaviour for a document that binds `"abc"` to an `int`. Worth deciding which it is while you are in there, because the docstring is currently the only place that claim is recorded.

src/sophios/compiler.py
```diff
@@ -1167,10 +1167,22 @@ def populate_scalar_val(cwl_type: Any, value: Any, fmt: Any = None) -> Any:
...
 
             case "float":
-                return float(value)
+                try:
+                    return float(value)
+                except ValueError as e:
+                    raise SophiosError.error(
+                        Code.LITERAL_TYPE_MISMATCH,
+                        f"Input {key!r} is declared type 'float' but its literal {value!r} "
+                        "does not convert to it.") from e
```
darkclad, line 1192: **🟠 `except ValueError` is narrower than the failure it is catching — `TypeError` still escapes undiagnosed, on both arms.**

`int()` and `float()` raise `ValueError` only for a *string that does not parse*. For a wrong-shaped argument they raise `TypeError`, which this `try` lets straight through:

```
int([1, 2])    -> TypeError
int({'a': 1})  -> TypeError
int(None)      -> TypeError
float(...)     -> same three
```

Reachable from a document, in three ways:

1. `!ii [1, 2]` on an input declared `int` — `populate_input_value` takes the scalar path at line 1234 and hands the list to `populate_scalar_val`.
2. `!ii {a: 1}` on an input declared `int` — same path.
3. `!ii [1, null]` on an input declared `int[]` — this is the sharpest one. The null guard at line 1200 inspects only the top-level `value`; `[1, null]` is not `None`, so it passes, and the array branch at 1216-1219 calls `populate_scalar_val('int', None, ...)` per item. `int(None)` is a `TypeError`.

So the case the guard at 1200 was written to diagnose (`MISSING_REQUIRED_INPUT`) leaks back in one element down, with no diagnostic at all.

`except (ValueError, TypeError)` covers all six, and the message you already wrote — "is declared type 'int' but its literal {value!r} does not convert to it" — is accurate for every one of them without change. Cheap to widen, and it is the difference between a `wic020` and a traceback out of a library that just committed (diagnostics.py:133-138) to "a failure the library reports, never a process it terminates".

src/sophios/compiler.py
```diff
@@ -120,7 +120,7 @@ def _arg_has_default_or_is_optional(arg: str, in_tool: dict[str, Any]) -> bool:
     canonical null-union representation. See canonicalize_type in utils_cwl.py.
     """
     arg_type = in_tool[arg]['type']
-    has_default = in_tool[arg].get('default')
+    has_default = 'default' in in_tool[arg]
```
darkclad: **🟡 This predicate also gates edge inference, so an empty-collection default now silently stops inferring.**

The change is right for the case it targets: `bool(in_tool[arg].get('default'))` treated `default: 0`, `default: false` and `default: ''` as absent, and presence is the correct question. But `_arg_has_default_or_is_optional` is not only used to decide what a user must supply. It feeds `args_required` at line 661, and `args_required` is what lines 977 and 1097 iterate to decide which inputs get an inferred edge at all.

So for `{'type': 'File[]', 'default': []}`:

- **before** — `[]` is falsy, so "no default", so required, so it appears in `args_required`, so edge inference runs and wires it to the upstream files;
- **after** — `'default' in spec` is true, so not required, so it never reaches the inference loop and the step compiles against the literal `[]`.

An empty-list default reads as "nothing unless something feeds me", which is exactly the input you most want inference to fill. Same for `{}` on a record type.

`default: null` on a non-optional type moves the other way: it now counts as having a default, where the value it carries is the one thing that cannot satisfy the input.

**I could not confirm this fires today** — I found no `default: []` in this repo's tool corpus or in the synthetic stubs, so nothing here would go red. That is also why it would go unnoticed. If you want presence semantics without the inference side effect, the narrow version is to keep presence for the "must the user supply this" question and exclude empty collections (or `None`) from the inference gate — or, since the two questions have now visibly come apart, to give them two predicates. Either way it is worth a stub with `default: []` in `synthetic_tools.py`, because right now the mirror there cannot see this either (see the review body).

src/sophios/inference.py
```diff
@@ -309,7 +309,8 @@ def perform_edge_inference(inference_use_naming_conventions: bool,
                     continue
 
                 in_tool = tool.cwl['inputs']
-                tool_in_formats = [arg_val['format'] for arg_key, arg_val in in_tool.items() if 'format' in arg_val]
+                tool_in_formats = [arg_val['format'] if isinstance(arg_val['format'], list) else [arg_val['format']]
+                                   for arg_key, arg_val in in_tool.items() if 'format' in arg_val]
```
darkclad: **🟠 Half the fix — the bare-string normalisation went to the whitelist candidate, not to the consumer input.**

`benchmarks.py:156-161`, added by this same PR, names both sites:

> `inference.perform_edge_inference` iterates `in_formats` directly, **and** `utils.flatten` does the same to a whitelist candidate's input formats — either given a bare string, the loop walks its characters instead of the one format it names, and the whitelist search never fires.

This hunk fixes the second. The first is untouched, a hundred lines up:

```python
209:  in_formats = []
210:  if 'format' in in_tool[arg_key]:
211:      in_formats = in_tool[arg_key]['format']     # bare string stays a bare string
212:      in_dict['format'] = in_formats
...
298:  for in_format in in_formats:                    # iterates characters
```

So a *consumer* input declaring `format: edam:format_1234` still walks `'e'`, `'d'`, `'a'`, `'m'`, … and `in_format in tool_out_formats` at line 325 compares single characters against whole format strings. The insertion search never fires, silently, no diagnostic — the same failure mode this PR is fixing, on the same pass.

Line 211 has a second consequence the whitelist site does not: it stores the bare string into `in_dict['format']`, which becomes `in_formats` inside `_match_outputs_of_step`. Line 65 there is

```python
not in_formats or out_format == in_formats or out_format in in_formats
```

and `in` on a `str` is substring containment, not membership. With `in_formats = 'edam:format_1234'`, an output declaring `format_123` matches as a substring of it. That is a false positive in the *ordinary* backward match, not just a missed insertion — arguably worse than the bug being fixed, because it wires an edge rather than failing to.

`in_formats = fmt if isinstance(fmt, list) else [fmt]` at line 211 fixes both, and makes `in_dict['format']` consistently a list, which is what line 65's `in` was written for.

**The new test cannot catch this**, and it is one character away from being able to: `test_inference.py:57` gives the consumer `'format': [dst_fmt]`. Make it the bare `dst_fmt` and you have the line-211 case — the test is otherwise already exactly the right fixture for it.

tests/core/benchmarks.py
```diff
@@ -0,0 +1,315 @@
...
+    insertion. "Two or more whitelisted converter tools" is the design's own
+    phrase for what this path requires (design_docs/core-refactor-design.md
+    §6.2), and it is what §6.3 names as where the documented `O(2^n)` lives.
+
+    A tool's *input* format must be a single-element list, not a bare string:
+    `inference.perform_edge_inference` iterates `in_formats` directly, and
+    `utils.flatten` does the same to a whitelist candidate's input formats —
+    either given a bare string, the loop walks its characters instead of the
+    one format it names, and the whitelist search never fires. This is not
+    documented anywhere; it was found by trying the bare-string form first.
```
darkclad: **⚪ This docstring is the best description of the bug in the tree, and the fixture below it deliberately avoids the half that is still live.**

Not a complaint about the prose — it is precise, and it is what let me pin down which half of the defect `inference.py:312` covers. The note is that lines 171 and 173 give every *input* the list form (`'format': [dst_fmt]`, `'format': [src_fmt]`) precisely because this docstring says the bare string breaks, while lines 169 and 174 leave the *outputs* bare. Correct for a benchmark — you want it to run, not to exercise a defect — but it does mean this case passes whether or not `inference.py:211` is ever fixed, so nothing in the tree is holding that half.

Once line 211 is handled, "a tool's *input* format must be a single-element list" stops being a constraint on the caller and this paragraph should say what, if anything, remains. Worth revisiting in the same change rather than leaving the tree's clearest statement of the bug describing a version of the code that has moved on.

darkclad (commented):
Review at `ae798c9`. Scope: the four commits on `semrefac_11` — `src/sophios/compiler.py`, `src/sophios/inference.py`, `src/sophios/lang/diagnostics.py`, and five test files.

All four fixes are correct in direction, and two of them I have no notes on. `insert_step_into_workflow`'s `{'id': stepid.stem}` (compiler.py:1261) is the right shape: it is what `canonicalize_steps_list` produces, what `get_steps_keys` reads, and what `add_yamldict_keyval_in` mutates — the old `{stem: None}` satisfied none of them. And `Code.LITERAL_TYPE_MISMATCH` slots into the `wic0NN` block without colliding. The new `perform_edge_inference` test builds the whole per-step context by hand and its 22 positional arguments line up with the signature, which is not nothing for a function with that shape.

**One blocker and one half-fix**, both inline:

- `tests/core/test_generators.py:162` catches `ValueError`, and `SophiosError` is not one. `lint_and_test.yml:244` runs that file. Details on the `int` hunk that now raises it.
- The bare-string `format` fix landed on one of the two sites that `benchmarks.py`'s own docstring names. Details on `inference.py:312`.

Three smaller things are inline too. One more has no line in this diff to sit on:

**`tests/core/synthetic_tools.py:148-166` is a deliberate mirror of the function you just changed, and it did not move.** `required_inputs_of` still reads `spec.get('default') is not None`, and its docstring still describes the rule this PR removed:

> Blind spot: for a falsy-but-present default (`{'type': 'int', 'default': 0}`), this function's `is not None` treats it as present while the compiler's `bool(in_tool[arg].get('default'))` treats `0` as absent, so the two disagree there — no stub has such a default, so `test_required_inputs_agree_with_the_compilers_own_rule` never exercises it.

The compiler no longer does that. `'default' in in_tool[arg]` agrees with `is not None` on `0`, `[]` and `''`, so the blind spot the docstring names is closed — but it has not gone away, it has moved to the other end: on `{'type': 'int?', 'default': None}` the mirror now says *required* and the compiler says *not*. The comment's own reason for keeping a second implementation ("a generator that asked the compiler which inputs are required, and then a property that checked the compiler honoured them, would be asking one implementation to grade itself") is exactly why `test_required_inputs_agree_with_the_compilers_own_rule` will not catch the drift: it compares the two only on the stubs that exist, and no stub has a null default. Worth moving the predicate and the docstring together, since the docstring is now describing code that is gone.

tests/core/benchmarks.py
```diff
@@ -0,0 +1,315 @@
...
+         "A workflow with two branches, each needing one whitelisted-format-"
+         "converter insertion — the design's own phrase, 'two or more "
+         "whitelisted converter tools', for what puts a compilation on this "
+         "path, and design_docs/core-refactor-design.md §6.3 names it as "
+         "where the documented O(2^n) lives. The corpus never reaches it. On "
+         "this branch it currently raises on the second fixed-point pass: "
+         "insert_step_into_workflow inserts a step with no 'id:' key, which "
+         "get_steps_keys requires — a real defect, recorded here rather than "
+         "fixed, since Spec 2 does not fix compiler defects it finds.",
+         run=_run_speculative_insertion),
```
darkclad: **⚪ This rationale is stale on arrival — the defect it records as unfixed is fixed four hunks away.**

> On this branch it currently raises on the second fixed-point pass: `insert_step_into_workflow` inserts a step with no `'id:'` key, which `get_steps_keys` requires — a real defect, recorded here rather than fixed, since Spec 2 does not fix compiler defects it finds.

`compiler.py:1261` in this same PR is `steps_mod.insert(i, {'id': stepid.stem})`. The defect is fixed, by this commit, in the diff this paragraph is part of.

The Spec 2 discipline it invokes — record, do not fix — is a good rule and I am not arguing with it. But the PR title is "Four compiler defects found by the Spec 2 oracle", so the recording and the fixing have deliberately landed together and the prose has not caught up. As written it tells the next reader that `speculative_insertion` still raises, which is the one thing this case's timing number now depends on *not* being true.

Two knock-on effects worth checking in the same pass:

- If the case raises today and does not after `compiler.py:1261`, its timing changes meaning — it was timing a path to an exception, and now times a real double insertion. Given §6.3 names this as where the documented `O(2^n)` lives, that number is about to get interesting rather than merely different, and `benchmarks_weekly.yml` is what will report it. That looks like the outcome you want; it is worth saying so here.
- `ast_strategies.py:48-66` and `test_equivalences.py:163-168` carry the same "currently unfixed" framing for defects in this stack, with line references (`compiler.py:1170`, `compiler.py:1173`) that the new `try` blocks have already shifted. Those files are not in this diff so I cannot comment inline, but they will read as wrong the moment this merges.

src/sophios/compiler.py
vjaganat90: Caught as `SophiosError` now. I took the second half of your point too: with a named diagnostic in place, an ill-typed `!ii` literal is ordinary expected behaviour rather than a declared residual, so the three docstrings that recorded it as a defect say that instead — `test_generators.py`, the pending-finding block in `ast_strategies.py`, and `test_equivalences`. The equivalence predicate itself stays, since a refused compile is still no basis for comparing two documents.

src/sophios/compiler.py
vjaganat90: Widened to `(ValueError, TypeError)` on both arms. The null array element was the one worth its own test rather than just a wider catch — the guard diagnoses the top-level case and then the array branch coerced `None` one element down, so the missing-input diagnostic leaked straight back out as a traceback.

src/sophios/compiler.py
vjaganat90: I went half your way, deliberately, and want to flag which half.

`default: null` I agree is wrong, and it is now excluded: a default satisfies an input when it is present *and not null*, null being the one value a non-nullable input provably cannot take.

On `default: []` I did not carve out empty collections. An empty list is still a value the tool author wrote, and excluding it reinstates exactly the truthiness special-casing this commit exists to remove — with the awkward edge that `default: []` and `default: 0` would then answer differently for no reason a user could predict. The deciding evidence for me was that `required_inputs_of`, the independent mirror in `synthetic_tools.py`, already read presence as `is not None`; matching it makes the compiler agree with the oracle that grades it, where a carve-out would have made them disagree in a third way.

You were right that nothing could see this. `sink.extras` now carries `default: []`, so `test_required_inputs_agree_with_the_compilers_own_rule` exercises the exact point where truthiness and presence differ, and the blind-spot paragraph that declared the gap is gone. If you still think an empty collection should invite inference, it is now a one-line change against a test that will notice — say the word and I will flip it.

src/sophios/inference.py
vjaganat90: Fixed line 211 as well, and rather than leave two normalisations that happen to agree, both sites plus the candidate's outputs now go through one `declared_formats`. The substring case is covered separately — `format_123` against an input declaring `format_1234` — since you are right that wiring an edge wrongly is the worse failure of the two. The test consumer is a bare `dst_fmt` now, as you suggested.

tests/core/benchmarks.py
vjaganat90: Rewritten, now that the input side is normalised — the paragraph no longer claims a constraint on the caller that has stopped existing.

tests/core/benchmarks.py
vjaganat90: Rewritten; the case now times a real double insertion rather than a path to an exception, and the rationale says so. On the knock-on files: both `ast_strategies.py` and `test_equivalences.py` are on master rather than above this branch, so their "currently unfixed" framing is wrong as of here and I corrected it in this PR rather than leaving it to read wrong on merge.

darkclad (commented):
Second pass, at `ac00ed6`. I compiled against this head and a few earlier ones instead of only reading the diff.

**Five of the six round-1 findings are fixed, and I checked each one.** Replies are on the threads. The `default: []` pushback is persuasive: it matches the mirror, and `sink.extras` makes the choice testable, so no flip is needed.

**Two new findings inline, both introduced this round:**

- 🔴 **Build Documentation fails on this PR and passes on master.** It is one docstring typo in `declared_formats`.
- 🟠 **The `not out_format` arm fixed `naming_conventions` by loosening `File` matching too.** Inference now wires a format-less `File` over a correctly formatted producer, and cwltool rejects that at runtime. Repro inline.

Separately, `Lint And Test` has never finished on this PR. Every run for `semrefac_11` was cancelled. `34858701972` got far enough to show the docs failure before it was cancelled. The external-adapter jobs are where the `File` issue would show up, so a re-dispatch once the fixes land would be worth it.

src/sophios/inference.py
```diff
@@ -62,9 +77,12 @@ def _match_outputs_of_step(*, out_keys: list[str], out_tool: dict, step_j: Yaml,
...
         # Great! We found an 'exact' type and format match.
         if types_match(in_dict['type'], out_dict['type']) and (  # First we have to match the types.
-                not in_formats or out_format == in_formats or out_format in in_formats):
+                not in_formats or not out_format
+                or out_format == in_formats or out_format in in_formats):
             # Then, if we have an input format or formats, the output format has to match.
-            # Otherwise, formats are optional and we match only on type.
+            # Otherwise, formats are optional and we match only on type. A format is
+            # unconstrained from either side: CWL forbids `format` on a string output, so
+            # requiring one there would make `format` on a string input unsatisfiable.
```
darkclad: **🟠 This arm applies to every type, not just `string`, so a format-less `File` can now take an edge from a correctly formatted producer.**

The reasoning in the commit is right for strings. CWL ignores `format` on non-`File` parameters, so requiring one on a `string` output makes `ambiguous.cwl` unsatisfiable. But `not out_format` also relaxes `File` inputs, and for those the old strict reading was the correct one.

Repro: three steps compiled with `compile_hermetic`.

```python
make_pdb: outputs  pdb: {type: File, format: edam:format_1476}
make_log: outputs  log: {type: File}                      # no format
use_pdb:  inputs   pdb: {type: File, format: [edam:format_1476]}
steps: make_pdb, make_log, use_pdb   # use_pdb has no in:
```

| tree | `use_pdb.pdb` wired to |
|---|---|
| master `71dd8c0` | `make_pdb/pdb` |
| `d8ba075` (before this commit) | `make_pdb/pdb` |
| **`ac00ed6`** | **`make_log/log`** |

The backward scan takes the most recent match, so the log file now takes the edge. That fails at runtime: cwltool's `cwl_utils.file_formats.check_format` raises `File has no 'format' defined` for a format-less `File` reaching an input that declares one. On master, list-form inputs got this right and only bare-string inputs had the `''`-substring accident. Normalising made the two forms agree, which was the point, but they now agree in the loose direction.

`package` cannot see this because no in-repo example has this shape. The external adapters are more likely to.

**Suggested fix:** exempt only inputs whose type is not `File`-like, since that is the case where CWL says `format` means nothing:

```python
not in_formats
or out_format in in_formats
or (not out_format and not _is_file_type(in_dict['type']))
```

`_is_file_type` would need to cover `File`, `File[]`, and the null-union and scatter-promoted array forms. Adding a `File`-typed version of the repro above to `test_an_output_declaring_no_format_still_matches_a_format_declaring_input` would pin both sides.

Alternatively, remove `format:` from the two `string` inputs in `ambiguous.cwl` and drop this arm. CWL only defines `format` for `File` parameters, and without it inference matches on type alone, which is exactly the behaviour the tutorial demonstrates.

Two smaller things on these lines. The comment at line 82 still says "if we have an input format or formats, the output format has to match", which this arm contradicts. And now that `in_formats` is always a list, `out_format == in_formats` can only be true if an output declares a list of formats. CWL gives an output parameter a single `format`, so in practice that arm is dead.

src/sophios/inference.py
```diff
@@ -31,6 +31,21 @@ def types_match(in_type: Any, out_type: Any) -> bool:
     return False
 
 
+def declared_formats(spec: Yaml) -> list[str]:
+    """The `format`s a CWL input or output declares, always as a list.
```
darkclad: **🔴 This line fails Build Documentation. Master passes.**

`Lint And Test` run `34858701972`, `ubuntu-latest`, Build Documentation step:

```
src/sophios/inference.py:docstring of sophios.inference.declared_formats:3: WARNING: Inline interpreted text or phrase reference start-string without end-string. [docutils]
##[error]Process completed with exit code 2.
```

The problem is `` `format`s ``. In RST a closing backtick cannot be directly followed by a letter, so docutils sees an opening backtick that never closes, and the docs build treats the warning as an error. I checked with docutils 0.23: this docstring as written reproduces the warning, and with ``` ``format`` values ``` in place of `` `format`s `` it is clean.

```suggestion
    """The ``format`` values a CWL input or output declares, always as a list.
```

This has been in the PR since `d8ba075`. The red run was cancelled, so it never appeared as a failed check.

tests/core/test_generators.py
```diff
@@ -159,7 +161,7 @@ def _collect(case: tuple[Document, Yaml]) -> None:
         compiled['drawn'] += 1
         try:
             compile_hermetic_cwl(yml, 'oracle')
-        except ValueError:  # the declared `!ii` residual; see the docstring
+        except SophiosError:  # an ill-typed `!ii` literal, diagnosed; see the docstring
```
darkclad: **⚪ Minor: `except SophiosError` now skips any diagnosed failure, not only ill-typed literals.**

The docstring you edited in `test_every_edge_a_document_references_is_defined_in_that_document` describes the same hazard: the one test that compiles generated documents swallows every error under a single name. Before, that was every `ValueError`. Now it is every `SophiosError`, which includes a dangling edge or a missing required input.

Checking the code keeps the skip as narrow as the docstring above claims:

```python
except SophiosError as e:
    if any(d.code is not Code.LITERAL_TYPE_MISMATCH for d in e.diagnostics):
        raise
    return
```

`Code` is exported from `sophios.lang`, next to the `SophiosError` import this commit added. Not blocking.

src/sophios/inference.py
darkclad: **🟠 This arm applies to every type, not just `string`, so a format-less `File` can now take an edge from a correctly formatted producer.**

The reasoning in the commit is right for strings. CWL ignores `format` on non-`File` parameters, so requiring one on a `string` output makes `ambiguous.cwl` unsatisfiable. But `not out_format` also relaxes `File` inputs, and for those the old strict reading was the correct one.

Repro: three steps compiled with `compile_hermetic`.

```python
make_pdb: outputs  pdb: {type: File, format: edam:format_1476}
make_log: outputs  log: {type: File}                      # no format
use_pdb:  inputs   pdb: {type: File, format: [edam:format_1476]}
steps: make_pdb, make_log, use_pdb   # use_pdb has no in:
```

| tree | `use_pdb.pdb` wired to |
|---|---|
| master `71dd8c0` | `make_pdb/pdb` |
| `d8ba075` (before this commit) | `make_pdb/pdb` |
| **`ac00ed6`** | **`make_log/log`** |

The backward scan takes the most recent match, so the log file now takes the edge. That fails at runtime: cwltool's `cwl_utils.file_formats.check_format` raises `File has no 'format' defined` for a format-less `File` reaching an input that declares one. On master, list-form inputs got this right and only bare-string inputs had the `''`-substring accident. Normalising made the two forms agree, which was the point, but they now agree in the loose direction.

`package` cannot see this because no in-repo example has this shape. The external adapters are more likely to.

**Suggested fix:** exempt only inputs whose type is not `File`-like, since that is the case where CWL says `format` means nothing:

```python
not in_formats
or out_format in in_formats
or (not out_format and not _is_file_type(in_dict['type']))
```

`_is_file_type` would need to cover `File`, `File[]`, and the null-union and scatter-promoted array forms. Adding a `File`-typed version of the repro above to `test_an_output_declaring_no_format_still_matches_a_format_declaring_input` would pin both sides.

Alternatively, remove `format:` from the two `string` inputs in `ambiguous.cwl` and drop this arm. CWL only defines `format` for `File` parameters, and without it inference matches on type alone, which is exactly the behaviour the tutorial demonstrates.

Two smaller things on these lines. The comment at line 82 still says "if we have an input format or formats, the output format has to match", which this arm contradicts. And now that `in_formats` is always a list, `out_format == in_formats` can only be true if an output declares a list of formats. CWL gives an output parameter a single `format`, so in practice that arm is dead.

src/sophios/inference.py
darkclad: **🔴 This line fails Build Documentation. Master passes.**

`Lint And Test` run `34858701972`, `ubuntu-latest`, Build Documentation step:

```
src/sophios/inference.py:docstring of sophios.inference.declared_formats:3: WARNING: Inline interpreted text or phrase reference start-string without end-string. [docutils]
##[error]Process completed with exit code 2.
```

The problem is `` `format`s ``. In RST a closing backtick cannot be directly followed by a letter, so docutils sees an opening backtick that never closes, and the docs build treats the warning as an error. I checked with docutils 0.23: this docstring as written reproduces the warning, and with ``` ``format`` values ``` in place of `` `format`s `` it is clean.

```suggestion
    """The ``format`` values a CWL input or output declares, always as a list.
```

This has been in the PR since `d8ba075`. The red run was cancelled, so it never appeared as a failed check.

darkclad (commented):
Duplicate of the review above, posted twice by mistake.

tests/core/test_generators.py
darkclad: **⚪ Minor: `except SophiosError` now skips any diagnosed failure, not only ill-typed literals.**

The docstring you edited in `test_every_edge_a_document_references_is_defined_in_that_document` describes the same hazard: the one test that compiles generated documents swallows every error under a single name. Before, that was every `ValueError`. Now it is every `SophiosError`, which includes a dangling edge or a missing required input.

Checking the code keeps the skip as narrow as the docstring above claims:

```python
except SophiosError as e:
    if any(d.code is not Code.LITERAL_TYPE_MISMATCH for d in e.diagnostics):
        raise
    return
```

`Code` is exported from `sophios.lang`, next to the `SophiosError` import this commit added. Not blocking.

src/sophios/compiler.py
vjaganat90: Following up on my own reply above, because a later round found the skip was still too wide: `except SophiosError` caught every diagnosed failure, not only the ill-typed literal. It now re-raises unless every diagnostic on the error is `LITERAL_TYPE_MISMATCH`, so a dangling edge or a missing required input fails the test rather than being counted as a skip. Detail in the thread on `test_generators.py:164`.

src/sophios/compiler.py
vjaganat90: Widened to `(ValueError, TypeError)` on both arms. The null array element earned its own test rather than just the wider catch — the guard diagnoses the top-level case and the array branch then coerced `None` one element down, so the missing-input diagnostic leaked straight back out as a traceback.

src/sophios/inference.py
vjaganat90: Fixed at line 211 too, with all three sites behind one `declared_formats` so they cannot drift apart again. The substring false positive has its own test (`format_123` against an input declaring `format_1234`), since wiring an edge wrongly is the worse of the two failures.

tests/core/benchmarks.py
vjaganat90: Rewritten now that the input side is normalised — it no longer states a constraint on the caller that has stopped existing.

tests/core/benchmarks.py
vjaganat90: Rewritten; the case times a real double insertion now rather than a path to an exception. On the knock-on files: `ast_strategies.py` and `test_equivalences.py` are on master rather than above this branch, so their "currently unfixed" framing is wrong as of here and I corrected it in this PR rather than leaving it to read wrong on merge.

src/sophios/inference.py
vjaganat90: Fixed — the first line is `Every \`format\` a CWL input or output declares` now, so no backtick is followed by a letter. I checked it the way you did rather than by eye: publishing the docstring through docutils reports no warnings, and the same probe over every docstring in the modules this chain touches finds only the seven that master already has.

vjaganat90: Fixed. The offending text was `` `format`s `` in the first line of `declared_formats` — a closing backtick directly followed by a letter, so docutils reads an opening backtick that never closes and the docs build promotes the warning to an error. It now reads `Every \`format\` a CWL input or output declares, always as a list.`

I checked it rather than eyeballing it: publishing that docstring through docutils reports no warnings, and running the same probe over every docstring in the modules this chain touches finds only the seven that master already has, so this chain adds none. Thanks for pinning it to the exact run and step — that made it a one-line confirmation instead of a hunt.

src/sophios/inference.py
vjaganat90: You are right, and it was already the live head by the time this landed: the arm is now `not out_format and not _type_permits_format(out_dict['type'])`. CWL defines `format` for File-valued parameters only, so a `string` output declares none because it cannot, while a `File` output declaring none has an unknown format and must not outrank the formatted producer. `_type_permits_format` recurses through arrays and unions so a `File[]` output is treated as File. There is a test pinning the File case does not take the edge, alongside the one for the `ambiguous.cwl` string case.

vjaganat90: Agreed, and fixed — the arm is `not out_format and not _type_permits_format(out_dict['type'])` now.

Your framing is the one I ended up with: the absence of a `format` means two different things, and only one of them is a free pass. CWL defines `format` for File-valued parameters only, so a `string` output declares none because it *cannot* — that is the `ambiguous.cwl` case, and requiring one there makes `format` on a `string` input unsatisfiable by construction. A `File` output declaring none has an *unknown* format, and letting that match would let a format-less File outrank the correctly formatted producer, which is your repro exactly. `_type_permits_format` recurses through arrays and unions, so a `File[]` output is treated as File.

Two tests hold the pair apart: one that a formatless `string` output still takes the edge, one that a formatless `File` output does not.

tests/core/test_generators.py
vjaganat90: Narrowed to exactly what the docstring claims: the handler re-raises unless every diagnostic on the error is `LITERAL_TYPE_MISMATCH`, so a dangling edge or a missing required input fails the test instead of being counted as a skip.

vjaganat90: Fixed, and you are right that the docstring already described the narrower contract than the code enforced. The handler now re-raises unless *every* diagnostic on the error is `LITERAL_TYPE_MISMATCH`:

```python
except SophiosError as diagnosed:
    if any(d.code is not Code.LITERAL_TYPE_MISMATCH for d in diagnosed.diagnostics):
        raise
    return
```

So a dangling edge or a missing required input now fails the test instead of being silently counted as a skipped draw — which was the real hazard, since this is the one test that compiles generated documents and it was swallowing everything under a single name.

src/sophios/compiler.py
darkclad: **Fixed** It now catches `SophiosError`, and the docstrings in `test_generators.py`, `ast_strategies.py` and `test_equivalences.py` treat the case as expected behaviour rather than a defect. I ran the test locally at this head with three random seeds and it passed each time. At `0e99bc9` it failed 5 of 5, and about 8% of draws hit `wic020`, so the old catch was really broken. One minor follow-up inline: the catch is now broader than the docstring says.

src/sophios/compiler.py
darkclad: **Fixed** Both arms now catch `(ValueError, TypeError)`. The null-array-element test was the right call: it pins the case where the top-level guard passes but an element one level down fails.

src/sophios/compiler.py
darkclad: **Agreed, no flip needed.** Matching the `is not None` mirror is a better reason than any carve-out I suggested, and `sink.extras: default: []` means a later change in either direction will fail `test_required_inputs_agree_with_the_compilers_own_rule`.

src/sophios/inference.py
darkclad: **Fixed** Both sites and the candidate's outputs now go through `declared_formats`, and the substring test and bare-string consumer are in. The follow-up commit `ac00ed6` introduced a new issue, though: the `not out_format` arm added for `naming_conventions` also relaxes `File` matching. Repro and suggested fix are in a new comment on `inference.py:80`.

tests/core/benchmarks.py
darkclad: **Fixed** The paragraph now says formats can be written either way, which is true at this head.

tests/core/benchmarks.py
darkclad: **Fixed** Thanks for also correcting the "currently unfixed" wording in `ast_strategies.py` and `test_equivalences.py`.

src/sophios/inference.py
darkclad: **Verified** at `5a7bd58`. Checking the output type is a better fix than the input-type check I suggested: it separates "cannot declare a format" from "declared none" at the exact place the ambiguity comes from. I recompiled both repros at this head, and both match master: `use_pdb` takes `make_pdb/pdb` rather than the format-less `make_log/log`, and `naming_conventions` step 2 is wired to `step__1__ambiguous/uniquename2` again. `test_a_formatless_file_output_does_not_outrank_the_formatted_one` covers the `File` case.

src/sophios/inference.py
darkclad: **Verified** at `5a7bd58`. I ran docutils over every docstring in `inference.py` and `declared_formats` is clean. The one warning left is in `perform_edge_inference`, whose docstring is identical to master's, so this PR adds nothing.

tests/core/test_generators.py
darkclad: **Verified** at `5a7bd58`. The handler re-raises unless every diagnostic is `LITERAL_TYPE_MISMATCH`, so it is now as narrow as the docstring says. The test passed on three random seeds locally.

darkclad (approved):

PR #409, opened 2026-09-09
feat(nextflow): shell-mode raw literal tokens, InitialWorkDirRequirement self-staging

vjaganat90:
Stacked on #406 (`nf203_optional_and_arrays`). Review after that branch; its commits appear here until it merges.

## What this adds

Two related Phase 2 lowerings, both about the execution environment — what gets staged, and how the command line is actually invoked.

**Shell-mode raw literal command tokens.** `ShellCommandRequirement` moves from deferred to supported. With no `shellQuote: false` binding present, rendering is unchanged — runtime-proven with an adversarial argument, since every command token already goes through the shell-quoting helper and is joined into one shell-invoked line, matching CWL's own shell-mode quoting. `shellQuote: false` on a prefix-free binding whose `valueFrom` is a pure CWL-author literal (no input reference in any form) now lowers to a new `NfShellLiteral` token, rendered unquoted so it can carry real shell operators — proven against the pinned Nextflow with a real append redirect. Every other `shellQuote: false` shape stays rejected with a diagnostic naming the boundary: no `valueFrom` (binds the input's own runtime value), a `prefix`, a `valueFrom` referencing an input directly or via `.path`/`.basename`, or `shellQuote: false` declared without `ShellCommandRequirement`.

**`InitialWorkDirRequirement` self-staging.** Also moves from deferred to supported, for one shape. Nextflow already stages every `path`-qualifier input under its own basename, so a listing entry that only asks for that — the bare `$(inputs.<name>)` shorthand, or a `Dirent` with no `entryname` or the self-referencing `$(inputs.<name>.basename)` — is a no-op. Staging under an explicit literal rename is also supported: a `Dirent` with a literal `entryname` lowers to a new `NfPort.stage_as` marker, rendered as Nextflow's `path ..., stageAs: '...'` option, which keeps the channel variable bound while staging under the literal name — proven end to end against the pinned Nextflow. Runtime proof also showed a renamed port's own `.name` reports the *staged* name, not the original CWL basename, so referencing that same input elsewhere in the tool's command, stream targets, or output globs is rejected: the command must hard-code the literal staged name directly, matching how the real `append.cwl` adapter's own listing entry (the bare-shorthand, own-basename shape) now lowers cleanly.

## Boundaries

- `shellQuote: false` is approved only for a literal, prefix-free binding. A binding with no `valueFrom`, a `prefix`, or a `valueFrom` referencing any input — including the exact real-world `append.cwl` shape (dynamic value, with a prefix) — stays rejected: unquoting runtime-supplied data is exactly the boundary this lowering must not cross.
- `InitialWorkDirRequirement` staging is scoped to picking an existing File/Directory input and (optionally) renaming it. `writable: true`, an `entry` that isn't a bare input reference (inline content construction), a computed `entryname`, a listing entry naming a `val`-qualifier or array-typed input, and two inputs staged under the same literal name all stay rejected with named diagnostics.
- A renamed input can't also be referenced elsewhere in the same tool's command/stream/glob templates — resolving what such a reference would mean (staged name vs. original basename) is out of scope; enforced both in capability analysis and as an `NfProcess` construction invariant.
- Staging an array of files, or a `val`-typed input, has no approved lowering yet.

## Schema

Two version bumps in this stack: `schema_version: 6` for the new `shell_literal` command-token kind, and `schema_version: 7` for the additive `stage_as` port field. Hydration accepts versions 2-6 unchanged; `stage_as` is optional on hydration (defaults `None`) since no earlier version ever wrote it.

## Design authority

Two `design:` commits approve this in order, both under §10 of `design_docs/sophios_nextflow_backend.md`:

- `9a0e99b` — shell-mode raw literal command tokens (§6, Commands).
- `a375e5d` — `InitialWorkDirRequirement` self-staging (§6, Inputs and channels).


src/sophios/utils_nf.py
```diff
@@ -290,14 +297,69 @@ def _binding_tokens(prefix: Any, value: NfTemplate, *, separate: Any = True) ->
...
+    template = _template(value_from, context=f"CWL shellQuote false value for {raw_name!r}")
+    literal_parts: list[str] = []
+    for segment in template.segments:
+        if not isinstance(segment, NfLiteral):
+            raise ValueError(
+                f"shellQuote false for {raw_name!r} references an input; only a "
+                "CWL-author literal with no input reference may be rendered unquoted"
+            )
+        literal_parts.append(segment.value)
+    return NfShellLiteral("".join(literal_parts))
```
darkclad, line 341: **🟠 A command made only of shell literals never runs.**

`_command` (line 704, outside this diff) decides the command is empty with `any(isinstance(token, NfTemplate) for token in tokens)`. An `NfShellLiteral` is not an `NfTemplate`, so a tool whose command is only raw shell text gets `true` put in front of it:

```yaml
requirements:
  ShellCommandRequirement: {}
arguments:
  - valueFrom: "cat a.txt | sort > out.txt"
    shellQuote: false
```

This renders as `true cat a.txt | sort > out.txt`. `true` ignores its arguments, `sort` reads nothing, `out.txt` is written empty, and the step exits 0.

Counting shell literals as runnable tokens there fixes it: `isinstance(token, (NfTemplate, NfShellLiteral))`. A test with no `baseCommand` and only `shellQuote: false` arguments would pin it.

src/sophios/utils_nf.py
```diff
@@ -382,6 +635,11 @@ def _input_binding_items(
         value_from = binding.get("valueFrom")
         position = _position(binding.get("position"), default=0)
         required_type = _required_type(input_definition.get("type"))
+        if binding.get("shellQuote") is False:
+            items.append(
+                ((position, 1, str(raw_name)), (_shell_literal_value(raw_name, binding, shell_mode=shell_mode),))
+            )
+            continue
```
darkclad, line 656: **🟠 The literal is emitted even when its input is null.**

CWL skips an input's binding when the input is null, `valueFrom` included. Here the literal is appended no matter what the input's value is. `_safe_absence_names` only looks at template references, and this literal has none by design, so `_absent_optional_findings` treats a null input as safe:

```yaml
inputs:
  compress:
    type: string?
    inputBinding:
      position: 10
      valueFrom: "| gzip -c"
      shellQuote: false
```

Leaving `compress` unset in CWL gives uncompressed output. The Nextflow process always pipes through `gzip`.

The two ways to fix it are (a) render the literal only when the input is present (for example, as a conditional token like `NfFlag`), or (b) remove inputs with a `shellQuote: false` binding from `_safe_absence_names`, so a null value is reported instead of silently changing the command.

src/sophios/utils_nf.py
```diff
@@ -365,8 +427,199 @@ def _array_binding(
...
+    if not isinstance(listing, list):
+        raise ValueError("InitialWorkDirRequirement listing must be a list")
+    tool_inputs = _as_mapping(tool.get("inputs", {}), error="CommandLineTool inputs must be a mapping")
+    stage_as: dict[str, str] = {}
+    by_rename: dict[str, set[str]] = {}
+    for item in listing:
+        name, rename = _iwdr_listing_item(item, tool_inputs=tool_inputs)
+        if rename is not None:
+            stage_as[name] = rename
+            by_rename.setdefault(rename, set()).add(name)
```
darkclad, line 587: **🟡 Staging an input twice keeps only one of its names.**

`stage_as` holds one rename per input, and the later entry overwrites the earlier one. Plain entries (`rename is None`) are never recorded at all. The duplicate check, both here and in `_iwdr_listing_findings`, only catches two *inputs* sharing one name, not one input staged under two names:

```yaml
listing:
  - entry: $(inputs.ref)
    entryname: a.fa
  - entry: $(inputs.ref)
    entryname: b.fa
```

CWL stages both `a.fa` and `b.fa`. This stages only `b.fa`, so a command that reads `a.fa` fails at run time. A bare `$(inputs.ref)` entry followed by a rename works the same way: CWL stages both names, and this stages only the rename.

A Nextflow `path` input has one `stageAs`, so the simplest fix is to reject any input named by more than one listing entry in `_iwdr_listing_findings`.

src/sophios/utils_nf.py
```diff
@@ -365,8 +427,199 @@ def _array_binding(
...
+                "entryname is supported only as a literal string or the input's own "
+                f"$(inputs.{raw_name}.basename); a computed or differently-referencing "
+                "entryname is deferred"
+            )
+        literal_parts.append(segment.value)
+    literal = "".join(literal_parts)
+    if not literal.strip():
+        raise ValueError("entryname must be a non-empty string")
+    if "/" in literal:
+        raise ValueError("entryname must not contain a path separator")
```
darkclad, line 475: **🟡 `*` and `?` in `entryname` are Nextflow wildcards.**

This check only rejects `/` and empty names, and so does `NfPort.__post_init__` (`nf_types.py:530`). In `stageAs`, Nextflow replaces `*` and `?` with an index, so `entryname: sample?.fq` can be staged as `sample1.fq`. The docstring on `_iwdr_rename_reference_findings` says a command should use the staged name as written, and that name would not exist.

I have not run this through Nextflow. The replacement is described in the `stageAs` docs, though, and rejecting both characters in both places costs nothing.

darkclad (commented):
Review at `f4c1a15`.

Four notes inline. The first two change what a step does with no error, so they need fixing before merge. The other two are narrower.

- **🟠 `utils_nf.py:341`:** a command made only of `shellQuote: false` literals gets `true` put in front of it, so the real command never runs and the step still succeeds.
- **🟠 `utils_nf.py:638`:** a `shellQuote: false` literal is emitted even when its optional input is null, and the absent-optional check lets that through.
- **🟡 `utils_nf.py:572`:** staging the same input under two names, or under its own name plus a rename, keeps only one of them, with no error.
- **🟡 `utils_nf.py:474`:** `entryname` allows `*` and `?`, which Nextflow's `stageAs` reads as wildcards.

src/sophios/utils_nf.py
vjaganat90: Fixed in 9f8a85f. `_command` now treats both `NfTemplate` and `NfShellLiteral` as runnable tokens, so the `true` fallback remains limited to genuinely flag-only/empty commands. The regression uses no `baseCommand` and asserts the command is exactly one raw `printf ok` literal.

src/sophios/utils_nf.py
vjaganat90: Fixed in 9f8a85f, with the contract recorded first in 245e6fa. `_safe_absence_names` now treats an input-owned `shellQuote: false` binding as presence-sensitive, so an absent optional source is rejected instead of emitting its static pipe/redirect. The capability regression covers this exact null-input shape.

src/sophios/utils_nf.py
vjaganat90: Fixed in 9f8a85f. Capability analysis now tracks every IWDR target, including no-op entries, and rejects a repeated input; `_iwdr_stage_as` has the same defensive invariant so no later path can silently overwrite. Regressions cover two renames and bare-plus-rename.

src/sophios/utils_nf.py
vjaganat90: Fixed in 9f8a85f. Both source-level `entryname` analysis and `NfPort` validation reject `*` and `?`, with regressions for `reads*.fq` and `sample?.txt`. The locked design restriction is in 245e6fa.

src/sophios/utils_nf.py
darkclad: **Fixed** At `9f8a85f`, a command made only of raw shell text renders as that text with no `true` in front, and the updated `test_command_of_only_shell_literals_still_runs_a_program` pins it. A lone `>>` now reaches the shell as a syntax error rather than a silent success, which is the better failure.

src/sophios/utils_nf.py
darkclad: **Fixed** Marking any input with a `shellQuote: false` binding as unsafe when absent was option (b), and it keeps the lowering static. The design doc now says an absent optional source is rejected, so option (a), making the literal conditional, is left for a separate lowering if it's ever needed.

src/sophios/utils_nf.py
darkclad: **Fixed** Both `_iwdr_listing_findings` and `_iwdr_stage_as` now reject a repeated input, and the finding points to where the input was first listed. The tests cover both the two-renames case and the plain-plus-rename case.

src/sophios/utils_nf.py
darkclad: **Fixed** `_iwdr_entryname` and `NfPort` both reject `*` and `?` now, so a hand-built IR can't bring the wildcard back either.

darkclad (approved):
LGTM at `9f8a85f`.

All four findings are fixed, and I checked each one against the code rather than the thread replies. Each fix has a test, and `nextflow-runtime` and `package` are green.

The change that limits schema-version checks to `processes` and `connections` isn't mentioned in the commit messages. It's a real fix, though: a param named `kind` or `stage_as` could have been rejected before. `test_hydration_treats_parameter_keys_as_opaque_user_data` covers it, so it's fine to keep here.

PR #410, opened 2026-09-09
Spec 2: canonical emission

vjaganat90:
**Stacked on #407** — review the chain in order.

Three sites let a `set`'s iteration order reach the emitted document, and two of them need no flag:

| site | reachability | measured |
|---|---|---|
| `utils_cwl.py` `{r: {} for r in set(reqs)}` | default path — any subworkflow, scatter, `when` or `valueFrom` | eight `PYTHONHASHSEED` values, **eight different `requirements:` key orders** |
| `utils_cwl.py` `list(set(new_strs))` | default path — routine `out:` extension | six seeds, **three different orders** |
| `compiler.py` `list(set(insertions))` | needs `--insert_steps_automatically` and two whitelisted converter tools | no hermetic generator reaches it |

`sorted(set(…))` at all three. `StepId` sorts lexicographically by `(stem, plugin_ns)`, asserted rather than assumed.

## Why this is stated as canonical order, not as agreement across seeds

The seed is the symptom. The defect is that a `set`'s iteration order is load-bearing in the output at all — and sortedness is checkable on every generated example without a second process, where varying the seed costs a subprocess per example and reaches two of the three sites.

It also matters for the equivalence relation below it in this stack: under a normalising comparison the `requirements` order would be *forgiven*, so a seed-varying property would have been vacuously true and `Strength.IDENTICAL` a strength nothing could satisfy.

## Three parts, because one would not be enough

- **The property** — over generated workflows, every collection that came from a set is in canonical order. Cheap, and runs on every example.
- **A total static scan** — outside an explicit allowlist, a `set` may not be iterated into an ordered structure anywhere in `src/sophios`. Total rule rather than a shape catalogue: a scan that enumerated the shapes it knew about once missed the two a PR had just fixed. Its companion feeds it each banned shape and asserts each fires, plus a false-positive check on the sanctioned form.
- **One four-seed regression example** — a hand-built workflow reaching all three sites, compiled in **fresh interpreters**. Fresh rather than pooled: this compiler mutates a module global and the tool registry, so seeded compilations sharing an interpreter could agree for the wrong reason.

Each of the three `sorted` changes is revert-checked individually, with a `diff` confirming the file actually changed before the result is trusted.


darkclad (approved):
LGTM at `a138948`. This covers that commit only; the parser changes under it belong to #407 and are reviewed there.

The `sorted(set(...))` fixes in `compiler.py`, `utils_cwl.py` and `plugins.py` are safe: `StepId` orders on two strings, and the success codes are already ints when they are sorted. The one set-built list left, `vars_workflow_output_internal`, is only used for membership tests. I ran the changed test files locally, including the property test and the four-seed subprocess test, and with the `utils_cwl.py` sorts reverted both of those fail, as they should.

Non-blocking, and older than this chain: `cwl_subinterpreter.py:87` and `:97` still build `steps:` as `[{cwl_tool: ...}]`, the form #407 removes, while `canonicalize_steps_list` requires an `id:` on every list entry (on `master` too). Your call whether that belongs in this PR, in #407, or in its own issue.

PR #411, opened 2026-09-09
Spec 2: the single-compilation predicates

vjaganat90:
**Stacked on #410** — review this after #410 and before #412. Both PRs intentionally remain based on `master`; branch ancestry supplies the stack order.

## What this PR owns

This remains the Spec 2 single-compilation predicate PR, but P32 now covers every reference Sophios resolves:

- inferred-edge candidate selection keeps the existing `types_match()` behavior;
- workflow-input references and explicit aliases are checked with a new internal, language-versioned `reference_relation(...)`;
- consuming and producing scatter are reflected in the effective endpoint types;
- Sophios raises `wic023` only when the relation is provably `DISJOINT`;
- `OVERLAPS` and `UNKNOWN` continue to final CWL validation.

The Python API uses the same relation before a compile version is selected. It rejects early only when every supported language version proves the pair disjoint; the resolved-version compiler check remains authoritative.

## Semantic boundary

Language `0.0.1` conservatively understands nullable shorthand, array shorthand, union lists, array `items`, recognized primitive equality, primitive incompatibility, and scalar/array shape mismatches. Unions overlap when any pairing overlaps, and are disjoint only when every pairing is proven disjoint.

`Any`, records, enums, named schema references, malformed declarations, missing cross-scope information, and other opaque schemas produce `UNKNOWN`. In particular, `docs/tutorials/echo_multi_scatter.wic` is pinned so an `Any` output feeding scattered string inputs cannot be rejected by Sophios.

This is deliberately a read-only linking constraint over raw passthrough data:

- preservation ownership and semantic observation are separate;
- there is no `PortType` IR in Spec 2;
- `types_match()` is unchanged and has no new rejection authority;
- production code does not import or call `cwltool.check_types()`;
- cwltool remains an external conformance oracle, while Spec 3 retains normalized type ownership.

The locked design document is changed only by `design: define the versioned reference-judgment boundary`.

## Predicate coverage

The existing namespace-injectivity, inferred-edge P32, and termination claims remain. Reference compatibility now has an independent test-side model and explicit outcome properties:

- `OVERLAPS` and `UNKNOWN` cases compile through Sophios;
- `DISJOINT` cases raise exactly one `wic023`;
- any unrelated `SophiosError` fails the property.

Coverage includes primitives, unions, arrays, consuming and producing scatter, `Any`, nested `Any`, records, enums, named references, unavailable cross-scope types, and the real tutorial regression. Static independence guards prevent production-oracle imports from the generators and prevent cwltool from entering the production language layer.

The AST inlineability filter now enforces its existing scattered-subworkflow boundary. Because that transformation cannot redistribute invocation scatter, it is not offered and cannot manufacture a false scalar endpoint.

## Validation

- complete fast suite: 918 passed, 1 skipped;
- focused compatibility suite: 32 passed;
- generator, predicate, diagnostic, Python API, corpus, and hermeticity coverage passed;
- configured cwltool conformance properties passed;
- mypy and the configured pylint gate passed.

Stack order: #410 → **#411** → #412.


src/sophios/api/python/workflow.py
```diff
@@ -75,14 +75,17 @@ def _tool_builder_source_name(value: Any) -> str | None:
...
+    from pre-empting the authoritative, resolved-version compiler judgment.
     """
-    if _contains_any_type(parameter_type) or _contains_any_type(candidate_type):
-        return True
-    return types_match(parameter_type, candidate_type)
+    return not all(
+        reference_relation(candidate_type, parameter_type, lang_version=lang_version)
+        is TypeRelation.DISJOINT
+        for lang_version in KNOWN_VERSIONS
+    )
```
darkclad: This now rejects an `Any[]` source bound to a scalar sink that is scattered afterwards. The check runs at assignment, before `scatter_on()` lifts the sink to an array, so array-vs-scalar comes back `DISJOINT` in every version. The removed `contains_any_type` shortcut used to let this through.

```python
echo.inputs.message = dup.outputs.output_array   # dup.cwl: output_array is Any[]
echo.scatter_on(echo.inputs.message)
```

This worked before and now raises `InvalidLinkError` on the first line, although the compiler accepts the resulting workflow. The same happens for a workflow input declared `Any[]` feeding a later-scattered scalar input, and for a workflow output declared `Any[]` bound to a scattered step's output.

The new parametrized test has no `Any[]` source against a scalar sink, so it doesn't catch this. Deferring the check until scatter is known (at build/compile time, or re-checking in `scatter_on()`) would fix it; so would treating array-to-scalar as not disjoint in this eager check.

darkclad (commented):
The compiler-side predicates look sound: `lang/compatibility.py` agrees with cwltool's `check_types`, scatter array-lifting matches cwltool's static typing, and the generator and the independent test model agree. One regression in the eager Python API check is noted inline.

One question on the inliner change: it reads scatter only from the call's own `parentargs`. Can scatter also arrive through `wic: steps:` metadata, and if so, is that path covered?

src/sophios/api/python/workflow.py
vjaganat90: Fixed, and the repro was exact — the check runs at assignment and `scatter_on()` lifts the sink afterwards, so it was answering a question it does not have the information for.

The rule now is that disjointness is proven here only when it holds for the sink as declared *and* for the sink lifted one array level — when no scatter arrangement could make the binding valid. `Any[]` into a scalar passes, and so does a typed array into a scalar, since scatter rescues both; `string -> File` and `string[] -> File` still reject, because nothing rescues those. That keeps the check from preempting the compiler, which is the thing its docstring says it must not do, rather than special-casing array-to-scalar.

The parametrised table has rows for both the `Any[]` and the typed-array case now, and reverting the lift makes both fail.

vjaganat90:
On your inliner question: no, that path was not covered, and it is now.

`ast.read_ast_from_disk` sets `parentargs` to `{**steps[i]}` — the call site as written under `steps:` — while `wic: steps:` metadata is merged onto the step in `compile_workflow_once`, at `compiler.py:623`, with the metadata winning. The inliner runs between those two, at `main.py:77`, so scatter spelled in the metadata was invisible to it and the subworkflow was offered for inlining.

The guard reads both now, through one helper, and the test is parametrised over four cases: scatter at the call site, scatter in the metadata, metadata carrying something that is not scatter, and no scatter anywhere. Removing the metadata branch fails the second.

darkclad (approved):
LGTM

PR #412, opened 2026-09-09
Spec 2: the canonical path

vjaganat90:
**Stacked on #411** — review and merge #411 first. Both PRs intentionally remain based on `master`; branch ancestry supplies the stack order.

## Executive summary

Spec 2 gives the compiler refactor an executable semantic contract. Its purpose is not to design the replacement compiler early; it is to make the behavior that Spec 3 must preserve precise enough that the replacement can be judged rather than trusted.

Before this work, the compiler was protected mainly by example workflows. Those examples established that familiar files still compiled, but they did not establish the compiler's stated invariants: that regrouping a workflow preserves meaning, namespacing cannot collide, emitted structure is deterministic, inferred and explicit connections are sound, compilation terminates deliberately, or the two Sophios front ends describe the same program. That was an unsafe starting point for replacing a recursive, stateful compiler with a typed IR and phase-separated pipeline.

Spec 2 closes that gap with one reusable semantic-equivalence relation, a set of independent single-compilation predicates, hermetic generators derived from the Spec 1 language model, permanent regressions for discovered counterexamples, and a narrow external-conformance boundary. The resulting oracle is also the acceptance mechanism for Spec 3: the new compiler may change internal representation, phase structure, traversal, and emitted ordering, but it must preserve the language behavior these properties define.

This PR is the capstone of that work. Earlier changes in the Spec 2 series establish the generators, equivalence relation, partition and inlining laws, deterministic emission, namespace and edge predicates, structured failure behavior, and operational benchmarks. #411 completes the Sophios-owned semantic boundary for resolved references. #412 closes the remaining canonical path from the Python surface through serialization and compilation to valid CWL and a lossless compute-request envelope.

## What Spec 2 owns

Spec 1 defines Sophios as a versioned language and states its deliberate leak over CWL v1.2. Spec 2 turns the observable semantic consequences of that language into executable invariants. Spec 3 remains responsible for the normalized `WorkflowGraph`, `PortType`, compiler phases, and replacement lowering architecture.

That separation matters. Spec 2 exercises the compiler because compilation is the behavior under protection; it does not canonize the present compiler's internal dictionaries. It inspects raw CWL port declarations only at explicitly owned linking decisions, leaves passthrough data unchanged, and does not introduce a provisional type IR. Likewise, `cwltool` is used as an external CWL conformance oracle, never as the production implementation or authority for Sophios semantics.

The semantic model has three parts.

### 1. Hermetic, language-derived inputs

Workflow strategies are built from the Spec 1 AST and a synthetic registry of tools with known signatures. Generated workflows therefore quantify over the language Sophios admits without depending on installed plugins, cached containers, search paths, or machine-specific discovery. The generators cover both valid programs and deliberately hostile siblings, and their own construct coverage and shrinking behavior are part of the contract. A failure can be reduced to a small, reproducible Sophios document rather than an environment-dependent compiler incident.

### 2. One shared meaning relation

All meaning-preserving rewrites use the same ordered equivalence relation:

| Strength | Permitted difference | Reason |
|---|---|---|
| `IDENTICAL` | None | The two paths should produce the same compiled artifact. |
| `UP_TO_EMBEDDING` | Relative `run:` placement | Embedding changes document location, not computation. |
| `UP_TO_RENAMING` | Names introduced by workflow nesting | Partitioning changes namespace spelling, not graph meaning. |

The strongest applicable relation is used for each claim. The renaming relation compares both graph structure and the multiset of invoked tools, so graph isomorphism cannot hide a workflow whose steps were replaced by different tools of the same shape. Divergences are reported at the semantic component that differs, making this relation suitable for the large differential comparison that Spec 3 will perform between the current and replacement pipelines.

This relation supports the central equivalence laws: repeat compilation is idempotent; partitioning a workflow into subworkflows preserves its meaning; inlining reverses that transformation; and compiling a Python workflow directly agrees with writing it as Sophios, loading the independent document, and compiling that document. Step reordering is intentionally not treated as equivalent because current inference semantics make order meaningful.

### 3. Predicates that do not need a second compilation

The remaining guarantees are properties of a single result:

- passthrough CWL survives without Sophios silently interpreting or rewriting fields it does not own;
- emitted collections have canonical order rather than inheriting hash-seed or set iteration order;
- namespacing is injective, so distinct ports cannot collapse to the same emitted identity;
- inferred connections and Sophios-resolved references satisfy their respective soundness rules;
- compilation reaches a fixed point or produces a structured diagnostic instead of exhausting an iteration cap silently;
- admitted failures remain diagnostics rather than uncaught implementation exceptions;
- the final artifact is valid CWL v1.2;
- the compiled CWL and job inputs can be placed, unchanged, into the existing compute-request envelope.

The properties are intentionally complementary. Path agreement alone could prove that two routes produce the same invalid document. CWL validation alone could validate only one route while the other diverged. Graph equivalence without namespace injectivity could accept an identity collision. Reference checking without conservative uncertainty could reject legal programs whose opaque CWL types Sophios does not own. Hermeticity prevents all of those conclusions from varying with the machine that evaluates them.

## Sophios-owned reference judgment

#411 places reference compatibility where Spec 2 already owns edge soundness. The internal, language-version-dispatched relation has three outcomes: `OVERLAPS`, `DISJOINT`, and `UNKNOWN`.

For language version 0.0.1, it recursively understands recognized primitives, nullable declarations, union lists, shorthand arrays, array `items`, and the effective endpoint types introduced by scatter. It proves disjointness only when the declarations justify that conclusion: incompatible recognized primitives, scalar/array shape disagreement, or unions for which every possible pairing is disjoint. `Any`, records, enums, named schemas, malformed declarations, and unavailable cross-scope information remain `UNKNOWN`.

That distinction is the rejection boundary. Explicit aliases and workflow-input references raise `wic023` only when their effective source and sink types are proven disjoint. `UNKNOWN` is accepted by Sophios and deferred to final CWL conformance. The older `types_match()` heuristic remains responsible only for choosing inferred-edge candidates; it does not acquire rejection authority. The Python API uses the same relation, so the two language surfaces cannot evolve separate compatibility policies.

This preserves Spec 1's leak boundary: port `type` and `format` declarations remain byte-preserved CWL data, while Sophios may observe them read-only for the enumerated act of linking ports. It also preserves Spec 3's design space. There is no normalized `PortType` algebra here, and production behavior does not delegate to `cwltool.check_types()`.

## What #412 closes

The final seam is the canonical user path:

```text
Python Workflow ───────────────→ direct compile ──────────────┐
       └─→ write_wic → load an independent document → compile ├─→ same CWL + job inputs
                                                               └─→ valid CWL → ComputeRequest
```

**P35 — path agreement.** The direct Python path and the serialized Sophios path previously disagreed on workflow-output references. One emitted the compiler's concrete step identifier while the other emitted a user-facing step name, and the compiler consumes an explicit `outputSource` verbatim. `workflow_document()` is now the single construction boundary and always emits the concrete spelling used by the compiler and corpus — there is no flag and no default selecting the other one. Both entry points obtain their output references from that builder. The property crosses a real serialization boundary and loads a distinct document before comparing the resulting CWL and job inputs at `IDENTICAL` strength; it is not object self-comparison.

P35 establishes agreement across the canonical multi-step topology exercised here. It does not quantify over nested workflows, scatter, `when`, or workflow-input references; the surrounding partitioning, passthrough, reference, and conformance properties cover those adjacent dimensions.

**P36 — CWL conformance.** The artifact from the canonical path is inlined and validated as CWL v1.2. This establishes that internal agreement terminates in a document accepted by the substrate Sophios declares. It does not move type authority into `cwltool`: Sophios makes its owned semantic decisions first, and external validation checks only the resulting CWL contract.

**P36a — compute-envelope integrity.** A `CompiledWorkflow` can produce an existing `ComputeRequest` whose workflow, job inputs, and identifier are the compiled values supplied to it. This is deliberately a shallow transport guarantee. It performs no additional semantic judgment, lowering, execution, or submission, and it does not make `CompiledWorkflow` the future IR. Spec 3 remains free to replace everything upstream of this stable envelope.

**Composed passthrough fidelity.** The earlier passthrough law covered a shallow workflow. The canonical-path coverage carries arbitrary passthrough freight through multiple steps and real scatter, where interpreted CWL and preserved CWL interact. This demonstrates that Sophios can add the requirements it owns without rewriting unrelated user material.

**Input-document immutability.** Binding a workflow input no longer inserts empty documentation fields, writes a literal `\\n`, or overwrites user-authored `doc` and `label` content. Useful consumer documentation may be carried forward, but absence remains absence and authored text remains intact. This removes mutation of source documentation as an accidental side effect of compilation.

**Oracle hermeticity.** The complete Spec 2 oracle surface is covered by the same static and runtime discovery boundary. Importing plugin types remains legal; invoking environment-dependent plugin discovery from the oracle does not. The canonical-path and reference-compatibility modules therefore operate under the same environmental contract as the generators and earlier properties.

## Closure and forward contract

At the tip of #412, every Spec 2 concern has an owner and the concerns compose across the full supported path:

- Spec 1 supplies the versioned language and preservation boundary.
- The generator supplies broad, shrinkable, environment-independent programs.
- The equivalence relation states what changes may and may not alter meaning.
- Single-compilation predicates cover determinism, identity, wiring, failure, and termination.
- #411 gives Sophios-resolved references a conservative, versioned rejection rule without inventing Spec 3's type system.
- #412 proves both front ends converge on one compiled artifact, that the artifact is conforming CWL, and that it reaches the existing compute boundary without reinterpretation.

This is the sense in which #412 completes Spec 2: the oracle now surrounds the compiler seam that Spec 3 will replace. It does **not** claim that the current compiler architecture is final, that generated CWL is byte-stable across compiler implementations, that CWL has been normalized into an IR, or that a compute request has been submitted or executed. Those would either pre-empt Spec 3 or cross into runtime integration.

The weekly benchmark lane remains observational and non-gating by design. It records corpus and speculative-insertion trends but makes no performance promise and does not turn incidental runner timing into language semantics.

After this PR, Spec 3 can replace the current recursive compiler behind the existing public boundary. Its burden is explicit: preserve the Spec 1 language contract and satisfy the Spec 2 oracle, while remaining free to introduce the typed graph, explicit obligations, phase separation, and normalized port types that belong to the new architecture.

Stack order: #410 → #411 → **#412**.


src/sophios/api/python/_workflow_runtime.py
```diff
@@ -314,23 +316,21 @@ def workflow_document(
...
-        else None
-    )
+    # Keyed by object identity, not by process_name: a step renamed after an
+    # output was bound to it still is the step the output names.
+    compiled_step_ids = {
+        id(step): step_name_str(
+            workflow.process_name,
+            index,
+            f"{step.process_name}.wic" if isinstance(step, Workflow) else step.process_name,
+        )
```
darkclad, line 331: **🟠 `outputSource` is spelled from `process_name`, but the compiler spells step ids from the file name.**

`step_name_str(workflow.process_name, …)` is right only when the document lands in `<process_name>.wic`. The compiler takes the prefix from the path it loads (`yaml_stem = Path(yaml_path).stem`, `compiler.py:190`). `_wic_output_path` accepts any `*.wic` name, and the docstring above says `compile_workflow_finish` uses an explicit `outputSource` verbatim. Nothing downstream fixes the mismatch.

For example, take a workflow with `process_name == 'wf'` whose output is bound to its second step, `append`:

```python
wf.write_wic('pipeline.wic')
# outputSource: wf__step__2__append/file
```

Compiling `pipeline.wic` names that step `pipeline__step__2__append`, so the output points at a step that does not exist. `to_wic_yaml()` has the same problem when its text is saved under any name other than `wf.wic`.

P35 does not catch it because `test_the_two_front_ends_compile_to_the_same_cwl` always writes `f'{via_file_workflow.process_name}.wic'`. Drawing the file name separately from `process_name` there would pin it.

The docstring rules out a second spelling, so the options are (a) have `write_wic` raise when `output_path.stem != workflow.process_name`, or (b) pass the destination stem into `workflow_document` and spell the ids from that. `to_wic_yaml()` has no destination, so either way its docstring should say the text compiles correctly only as `<process_name>.wic`.

src/sophios/compiler.py
```diff
@@ -962,15 +1030,27 @@ def compile_workflow_once(yaml_tree_ast: YamlTree,
...
+                            f"'{arg_key}' to a different source.")
+
+                    # Preserve the user's input block while carrying useful
+                    # documentation from the consuming argument.
+                    for key in ('doc', 'label'):
+                        addition = in_dict.get(key, '')
+                        if not addition:
+                            continue
+                        existing = inputs_key_dict.get(key, '')
+                        inputs_key_dict[key] = f'{existing}\n{addition}' if existing else addition
```
darkclad, line 1066: **🟡 This merge assumes `doc` is a string, but CWL also allows `string[]`.**

If either side is a list, the f-string writes its Python repr into the user's input block:

```python
existing = 'mine'
addition = ['Line one', 'Line two']
# -> "mine\n['Line one', 'Line two']"
```

The block's comment says its purpose is to preserve the user's documentation, and this output damages it.

`label` is string-only in CWL, so only `doc` is affected. I found no list-valued `doc` in this repo's adapters, so nothing here goes red today. Tools from outside the repo can still have one. Joining lists first fixes it:

```python
def _as_text(v): return '\n'.join(v) if isinstance(v, list) else v
```

Apply it to both `existing` and `addition` before the f-string.

tests/core/test_ci_coverage.py
```diff
@@ -0,0 +1,200 @@
...
+                    continue
+                tokens = shlex.split(line, comments=True)
+                # `python -m pytest -m "not skip_pypi_ci"` carries two `-m`: the
+                # first selects the module to run, the second is pytest's marker
+                # filter. Only the one after the `pytest` token is the filter.
+                start = next((i for i, t in enumerate(tokens) if t == 'pytest' or t.endswith('/pytest')), -1)
+                if start < 0:
+                    continue
+                after = tokens[start + 1:]
+                files = [t for t in after if t.endswith('.py')]
```
darkclad: **⚪ A directory argument is read as "no path named", so it counts as covering the whole rootdir.**

This keeps only `.py` tokens. `pytest tests/contrib …` therefore yields `files == []`, and line 189 treats that as the rootdir.

It does not misfire today. The only directory run is `lint_and_test.yml:291`, `pytest tests/contrib -m "not skip_pypi_ci"`, and it excludes the one marker the packaging lane excludes, so it admits nothing extra. The first `pytest tests/contrib -m skip_pypi_ci` would mark every `skip_pypi_ci` test under `tests/core` as covered, and this is the one failure this file exists to catch.

Something like `[t for t in after if not t.startswith('-') and (REPO_ROOT / t).exists()]` catches directories without picking up flag values such as `cwltool`. Line 189 then becomes a prefix check (`relative == f or relative.startswith(f.rstrip('/') + '/')`).

tests/core/test_ci_coverage.py
```diff
@@ -0,0 +1,200 @@
...
+        for test, markers in _marked_tests(path).items():
+            blocking = markers & packaging_excludes
+            if not blocking:
+                continue  # the packaging lane's default collection reaches it
+            admitted = any(
+                (not files or relative in files)          # no path named means the whole rootdir
+                and (keyword is None or keyword in test)  # -k narrows by test name
+                and not (_excluded_markers(marker) & blocking)
+                for files, marker, keyword in main_runs
+            )
```
darkclad, line 237: **⚪ `keyword in test` treats a `-k` expression as a plain substring.**

That is right for a bare keyword, but a boolean expression never matches. `run_workflows.yml:186` runs

```
pytest tests/core/test_python_api.py -k "not test_compile_python_workflows and not test_validate_generated_python_workflows"
```

That run executes nearly every test in the file, but this check says it admits none of them. It does not fire today because `test_python_api.py` has no `skip_pypi_ci` tests. The first one added there would be reported as an orphan even though it runs.

This fails loudly rather than silently, so it is the less dangerous direction. It still differs from how this file treats `-m`: `_excluded_markers` reads the shapes the repo writes and raises on anything else. Doing the same for `-k` would keep the two flags consistent: handle a bare name and `not <name>` joined by `and`, and raise otherwise.

darkclad (commented):
Review at `43a6110`.

Four notes inline. The first needs fixing before merge. The other three are smaller or do not fire today.

- **🟠 `_workflow_runtime.py:321`:** the `outputSource` in a written `.wic` is spelled from `process_name`, but the compiler spells step ids from the file name. Any `write_wic('<other>.wic')` produces a document whose outputs point at steps that do not exist. P35 always writes `{process_name}.wic`, so it cannot see this.
- **🟡 `compiler.py:1048`:** the `doc`/`label` merge assumes strings. CWL allows `doc` to be a list, and a list gets merged as its Python repr.
- **⚪ `test_ci_coverage.py:83` and `:190`:** directory arguments count as the whole rootdir, and `-k` expressions are matched as plain substrings. Neither misfires on today's workflows, but the first can hide an orphan and the second can report a false one.

src/sophios/api/python/_workflow_runtime.py
vjaganat90: Fixed, and I took the second of your two options. `workflow_document` accepts the destination stem and spells step ids from it; `write_workflow_wic` passes `output_path.stem`. A document written as `pipeline.wic` now says `pipeline__step__2__append`, and one written as `wf.wic` is unchanged, so the artifact is correct under whatever name it is saved as rather than only under one.

`to_wic_yaml()` has no destination to spell for, so its docstring now says the text compiles correctly only as `<process_name>.wic`, and why.

The new test writes the same workflow under two stems and asserts the prefix matches the file — which is what P35 could not see, since it always writes `f'{process_name}.wic'`. Reverting the stem argument fails the second stem.

src/sophios/compiler.py
vjaganat90: Fixed. Both sides go through a small `_as_text` that joins a list before the f-string, so neither a user-authored `doc` nor a consuming argument's can reach the block as a repr. `label` is untouched by the list case, as you say, but it goes through the same call rather than being special-cased — the block already treats the two keys alike.

The test is parametrised over the four combinations: argument list, user list, both lists, neither. Reverting the join fails the three that involve one.

tests/core/test_ci_coverage.py
vjaganat90: Fixed, and you were right that this is the dangerous direction of the two.

Paths are now anything that is not a flag and exists in the repo, so `pytest tests/contrib` is read as a path rather than as the absence of one, and the match became a prefix check. That run is now seen with its path instead of as the whole rootdir, which is what would have marked every marked test under `tests/core` as covered.

tests/core/test_ci_coverage.py
vjaganat90: Fixed, and given the same treatment as `-m` for the reason you name: `_keyword_admits` reads a bare name and `not <name>` joined by `and`, and raises `ValueError` on anything else. A boolean expression no longer silently matches nothing.

The `run_workflows.yml` invocation you quote is now read correctly, and there is a case asserting the refusal alongside the ones for the shapes it does read — the same pairing `_excluded_markers` has.

darkclad (approved):
LGTM at `4f311ad`.

All four findings are fixed, and I checked each one against the code rather than the thread replies. Each fix has a test that fails without it. The `pipeline` stem is the case that pins the `outputSource` fix, since `oracle` is the workflow's own name. The scatter-aware eager check in `workflow.py` and the `wic: steps:` scatter guard in `inlineing.py` are sound and covered by the new cases in `test_reference_compatibility.py`. `package` is green.

Non-blocking: `_pytest_runs` now drops any token that does not exist on disk, so a deleted file or a node id like `test_x.py::test_y` leaves a run with no paths, and that reads as the whole rootdir. No workflow writes either shape today, and pytest fails a job that names a missing file, so this is only worth a follow-up if it ever bites.

vjaganat90:
Noted, and taken as a follow-up rather than here, since this PR is approved and the shape you describe is not written anywhere today.

It is on #415 now. You are right that it is the same silent pass by another route: dropping the last path argument leaves none, and none reads as the whole rootdir. A token shaped like a path — containing `/` or `.py` — that does not resolve now raises, which is the refusal `_excluded_markers` already makes for an expression it cannot read; a bare word stays ignored, since it may be a flag's value, and a node id keeps its file part.

darkclad (approved):
LGTM at `4f311ad`.

All four findings are fixed, and I checked each one against the code rather than the thread replies. Each fix has a test that fails without it. The `pipeline` stem is the case that pins the `outputSource` fix, since `oracle` is the workflow's own name. The scatter-aware eager check in `workflow.py` and the `wic: steps:` scatter guard in `inlineing.py` are sound and covered by the new cases in `test_reference_compatibility.py`. `package` is green.

Non-blocking: `_pytest_runs` now drops any token that does not exist on disk, so a deleted file or a node id like `test_x.py::test_y` leaves a run with no paths, and that reads as the whole rootdir. No workflow writes either shape today, and pytest fails a job that names a missing file, so this is only worth a follow-up if it ever bites.

PR #413, opened 2026-09-09
feat(nextflow): executable scatter, one channel adapter, and nested workflows

vjaganat90:
Stacked on #409 (`nf204_shell_and_staging`). Review after that branch; its commits appear here until it merges.

## What this adds

Three coupled Phase 2 lowerings: executable scatter, the one channel adapter scatter needs, and one level of nested workflows.

**One channel adapter.** Scatter is the only approved topology whose two ends disagree about channel cardinality, so `NfWorkflowInputConnection` gains a single typed `adapter` attribute with a closed one-member set: the queue fan-out. It renders as Nextflow's `flatten` applied at the *consumption site* — `PROCESS(param.flatten())` — so parameter construction is unchanged from the array-typed-input form already proven in #406, and each sink derives its own queue channel. That placement is why one scattered parameter can feed two processes without a consumed-once channel; both that and the value-channel broadcast of a scattered step's other inputs now have installed-Nextflow proof. The adapter joins `path_kind` and the array marker in the parameter channel-semantics check, so a parameter is adapted at every sink or at none.

**Single-input executable scatter.** A step whose `scatter` names exactly one input (string or one-element list) runs once per element of an array-typed workflow input — runtime-proven end to end from a real compiled fixture, for `string[]` and for `File[]` whose elements are CWL File objects. Multi-input scatter is rejected, and with it every load-bearing `scatterMethod`. This is where real compilation contradicted my first design: I had planned to reject `scatterMethod` outright, but the compiler emits `scatterMethod: dotproduct` unconditionally even for single-input scatter, so that would have put executable scatter out of reach of the Python and CLI surfaces and of real-fixture runtime proof. At exactly one scattered input all three methods coincide (one task per element, output depth one), so one of those three values is now consumed as an inert restatement and anything else is rejected by name.

**Collection cardinality and empty collections**, the two §9 gates. A scattered step's output carries one value per task, so its only approved sink is a workflow output — exactly the array-typed output CWL declares. A process-to-process edge out of a scattered step is rejected: CWL gives that consumer one invocation receiving an array, while the channel drives N invocations, and the gather adapter that would close the gap is deferred. For the same reason a scattered step's own inputs must all come from workflow parameters, since a process output is a queue channel that would truncate the run to one task. Empty collections are proven, not assumed: scattering over `[]` runs zero tasks, unscattered processes in the same run still execute, the emitted channel is simply empty, and the run exits successfully — no repeat of the `Channel.value(null)` hang from #406.

**One level of nested workflows.** A step whose compiled `run` is a `Workflow` lowers by inlining, in a composition pass that runs before every other analysis: inner steps become outer steps namespaced as `<outer step>___<inner step>`, each subworkflow input is replaced by the outer binding, and references to the outer step's outputs are rewritten to the inner endpoints its `outputSource` names. Emitting a nested DSL2 `workflow` block is deferred instead, because a subworkflow call would add a new connection endpoint kind across the graph validator, the renderer, and the generated-subset reader at once, while inlining reaches the same tasks, data flow, and output files with none of that. Namespacing answers the collision gate concretely (two instantiations of one subworkflow coexist); subworkflow I/O is total in both directions. Runtime-proven from a real compiled fixture whose subworkflow input comes from an outer step's output.

## Boundaries

- Multi-input scatter and every `scatterMethod` value outside the three CWL names are rejected. A `scatterMethod` alongside single-input scatter is accepted only because it is provably inert there.
- A scattered step's outputs can reach only a workflow output. Gathering a queue channel back into one list-valued element — the gather half of scatter/gather — is deferred, as are pairing and crossing queue channels and every other operator: the adapter set is closed data the renderer maps but cannot widen.
- A scattered source must be exactly one array-typed workflow input whose item type matches the scattered port's qualifier and path kind. A process output can never carry an array (array-typed outputs have no lowering), so scattering over one is rejected rather than silently truncated. Array-typed scattered ports, undeclared and unwired scattered inputs are rejected by name.
- Nesting deeper than one level is rejected, and so is `scatter` on a subworkflow step — scattering an inlined sub-DAG is not the single-process shape scatter supports. A scattered step *inside* a subworkflow is not a separate lowering: after inlining it is an ordinary scattered step and follows the same contract.
- Subworkflow I/O must be total: every declared input bound by the step, every `out` name declared as a subworkflow output, and every subworkflow output resolving to one of its own step outputs. A subworkflow output that forwards one of its inputs is rejected like any other boundary passthrough.
- Workflow-level requirements are now closed-world. `ScatterFeatureRequirement` and `SubworkflowFeatureRequirement` are consumed as inert declarations at both levels; every other one is rejected by name, and workflow-level `hints` stay unconsumed.
- Composition findings are raised before the rest of capability analysis rather than aggregated with it — the flat graph every later pass reads cannot be built while its composition is unsupported. They still aggregate among themselves across every nested step.
- The generated scatter call is outside the reader's recognized subset, so a scattered workflow round-trips as loss-aware structure with an opaque region and cannot be promoted; pinned by a reader test.

## Schema

`schema_version: 8` for the additive `adapter` field on workflow-input connections. Hydration accepts versions 2-8 unchanged; `adapter` is optional on hydration (defaults `None`) since no earlier version ever wrote it, and `_reject_newer_kinds` gates it the way it gates `is_array` and `stage_as`. Nested workflows need no schema change at all — inlining produces the same flat executable model.

## Design authority

Three `design:` commits approve this in order, all under §10 of `design_docs/sophios_nextflow_backend.md`:

- `a8665d7` — one channel adapter for queue fan-out (§6, Inputs and channels).
- `c650ffc` — single-input executable scatter (§6, Topology).
- `c6ae03a` — one level of nested workflows (§6, Topology).


darkclad (commented):
Review at `20f0251`.

Six notes inline. The first two are the ones I would fix before merge: one gives wrong results with no error, the other breaks loading payloads that worked before this PR.

- **🔴 `input_output_nf.py:316`:** a scattered step that also takes a File or Directory from a workflow input runs one task instead of N.
- **🟠 `nf_types.py:965`:** the new `stage_as`/`adapter` version gates also match keys inside `params`, so an older payload with a workflow input named `adapter` or `stage_as` no longer loads.
- **🟡 `utils_nf.py:584`:** staging one input under two names keeps only the last name, with no error.
- **🟡 `utils_nf.py:486`:** `*` and `?` in a literal `entryname` become `stageAs` wildcards.
- **🟡 `utils_nf.py:1756`:** a nested workflow's input and output fields are never checked for unsupported keys.
- **🟡 `utils_nf.py:650`:** a `shellQuote: false` literal is still rendered when its optional input is absent.

Separately, not a bug: the same array input can't be scattered in one step and passed whole to another. The IR tests show that rejection is intended, but users get the generic "incompatible channel qualifiers" error from `nf_types.py` instead of a message that names the unsupported pattern.

src/sophios/input_output_nf.py
```diff
@@ -271,23 +295,28 @@ def _render_named_workflow(workflow: ExecutableNextflowWorkflow) -> str:
...
     raise ValueError(f"workflow input {name!r} is not connected")
 
 
 def _parameter_expression(workflow: ExecutableNextflowWorkflow, name: str) -> str:
-    port = _workflow_input_port(workflow, name)
+    connection, port = _workflow_input_sink(workflow, name)
+    # A scatter-adapted parameter carries the whole source array; the graph
+    # validator keeps every sink of one parameter in agreement, so one sink
+    # decides the construction for all of them.
+    carries_list = port.is_array or connection.adapter == "scatter"
```
darkclad, line 328: **🔴 A File/Directory broadcast input truncates the scatter to one task.**

This branch changes only the scattered sink. A scattered step's *other* inputs still render as they did before. For a scalar `path` parameter that is the `Channel.fromPath(...)` at line 329, a queue channel holding exactly one item.

Nextflow pairs a process's queue-channel inputs item by item and stops at the shortest channel. A scattered port fed `params.xs.flatten()` (3 items) plus a File port fed `Channel.fromPath(params.ref)` (1 item) runs **one** task, and the other two elements are dropped silently.

Nothing rejects this today. `_scatter_edge_findings` only looks at sources containing `/` (process outputs), so a workflow-input source is skipped. R2.17 (`test_a_scattered_step_broadcasts_its_unscattered_inputs`) uses a `string` broadcast, which renders as `Channel.value` and broadcasts correctly, so the test passes.

Either render a scalar path parameter as a value channel (`Channel.value(file(..., checkIfExists: true))`) whenever any of its sinks is on a scattered step, or reject that shape in capability analysis until it is supported. Adding a File variant of R2.17 would cover it.

src/sophios/nf_types.py
```diff
@@ -868,19 +947,23 @@ def _connection_from_dict(value: Mapping[str, Any]) -> NfConnection:
...
     # Version an additive non-kind-tagged field was introduced in, keyed by
-    # the field name it appears under. is_array predates a "kind" tag on
-    # NfPort, so it needs its own gate alongside KIND_SCHEMA_VERSIONS.
-    FIELD_SCHEMA_VERSIONS: ClassVar[Mapping[str, int]] = MappingProxyType({"is_array": 5})
+    # the field name it appears under. is_array and stage_as predate a
+    # "kind" tag on NfPort, and adapter is additive on an existing
+    # connection kind, so each needs its own gate alongside
+    # KIND_SCHEMA_VERSIONS.
+    FIELD_SCHEMA_VERSIONS: ClassVar[Mapping[str, int]] = MappingProxyType(
+        {"is_array": 5, "stage_as": 7, "adapter": 8}
```
darkclad, line 967: **🟠 `_reject_newer_kinds` also applies these field gates to the `params` subtree.**

`_reject_newer_kinds` walks every nested mapping in the payload, including `params`, and runs `mapping.get(field_name)` on each. `params` is keyed by workflow input names. So a schema-v5 payload whose workflow has an input named `adapter` (common for adapter-trimming tools) or `stage_as`, with a truthy value, now fails with `'adapter' requires executable Nextflow schema version 8`. Before this PR it loaded.

`is_array` already had this problem, but adding `stage_as` and `adapter` makes it much easier to hit, and a `params` key named `kind` whose value is `array` or `shell_literal` trips `KIND_SCHEMA_VERSIONS` the same way.

Skip `params` in the walk: call `_reject_newer_kinds` on `processes` and `connections` only. Neither gate belongs to user-supplied values. A v5 round-trip test with a `params` key named `adapter` would lock this in.

src/sophios/utils_nf.py
```diff
@@ -365,8 +439,199 @@ def _array_binding(
...
+    listing = requirement.get("listing")
+    if not isinstance(listing, list):
+        raise ValueError("InitialWorkDirRequirement listing must be a list")
+    tool_inputs = _as_mapping(tool.get("inputs", {}), error="CommandLineTool inputs must be a mapping")
+    stage_as: dict[str, str] = {}
+    by_rename: dict[str, set[str]] = {}
+    for item in listing:
+        name, rename = _iwdr_listing_item(item, tool_inputs=tool_inputs)
+        if rename is not None:
+            stage_as[name] = rename
```
darkclad, line 598: **🟡 One input staged under two names keeps only the last one.**

`stage_as[name] = rename` overwrites. The duplicate check below only catches the reverse case, two inputs sharing one name. Neither this function nor `_iwdr_listing_findings` rejects these listings:

- two Dirents with the same `entry` and different literal `entryname`s, or
- a bare `$(inputs.x)` or basename self-reference, plus a rename of `x`.

CWL stages the file under both names. Nextflow's `stageAs` can give a path input only one name, so the lowered process has only the last name. A command that uses the other name fails at runtime instead of during capability analysis.

Track every listing entry per input, including the rename-less ones, and report a finding when one input appears more than once.

src/sophios/utils_nf.py
```diff
@@ -365,8 +439,199 @@ def _array_binding(
...
+                "entryname is supported only as a literal string or the input's own "
+                f"$(inputs.{raw_name}.basename); a computed or differently-referencing "
+                "entryname is deferred"
+            )
+        literal_parts.append(segment.value)
+    literal = "".join(literal_parts)
+    if not literal.strip():
+        raise ValueError("entryname must be a non-empty string")
+    if "/" in literal:
+        raise ValueError("entryname must not contain a path separator")
```
darkclad, line 487: **🟡 `*` and `?` in a literal `entryname` become `stageAs` wildcards.**

The literal becomes the port's `stageAs`, and in Nextflow `*` and `?` in a `stageAs` name are replaced with index digits. `entryname: sample?.txt` is staged as `sample1.txt`, and `data*.csv` as `data1.csv`. CWL treats the entryname as a literal file name, so the command looks for a file that does not exist.

Reject `*` and `?` here next to `/`, since there is no escape syntax for them in `stageAs`.

src/sophios/utils_nf.py
```diff
@@ -1192,6 +1490,601 @@ def _workflow_capability_findings(
...
+            continue
+        document, inner_children = nested
+        findings.extend(_composition_findings_for_step(step, path=path))
+        findings.extend(
+            _unconsumed_field_findings(
+                document,
+                consumed=_WORKFLOW_CONSUMED_FIELDS,
+                path=f"{path}.run",
+            )
+        )
```
darkclad, line 1773: **🟡 A nested workflow's `inputs` and `outputs` fields are never checked for unsupported keys.**

This checks the nested document's top-level fields against `_WORKFLOW_CONSUMED_FIELDS`. The per-port checks that `_workflow_capability_findings` runs for the top-level workflow (`_WORKFLOW_INPUT_CONSUMED_FIELDS` at line 1409, `_WORKFLOW_OUTPUT_CONSUMED_FIELDS` at line 1457) never run on `document["inputs"]` or `document["outputs"]`. `_subworkflow_output_endpoints` reads only `outputSource`.

So a nested workflow output with `linkMerge` or `pickValue`, or a nested input with `secondaryFiles` or `loadContents`, is flattened and those fields are dropped with no warning. The same fields on the outer workflow are rejected. Running the same two port loops over the nested document with `path=f"{path}.run.inputs.{name}"` would make the two levels agree.

src/sophios/utils_nf.py
```diff
@@ -382,6 +647,11 @@ def _input_binding_items(
         value_from = binding.get("valueFrom")
         position = _position(binding.get("position"), default=0)
         required_type = _required_type(input_definition.get("type"))
+        if binding.get("shellQuote") is False:
+            items.append(
+                ((position, 1, str(raw_name)), (_shell_literal_value(raw_name, binding, shell_mode=shell_mode),))
+            )
```
darkclad, line 667: **🟡 A `shellQuote: false` literal is rendered even when its optional input is absent.**

In CWL, a null input contributes nothing to the command line, including its `valueFrom`. Here the binding becomes an unconditional `NfShellLiteral`, so for `note: string?` with `valueFrom: "| sort"` and no value supplied, `| sort` still goes into the script.

Absence analysis doesn't stop it either. `_safe_absence_names` only counts `NfTemplate` references, and the literal has no input reference, so the input counts as safe to leave absent and conversion succeeds.

Quoted literal `valueFrom`s already behaved this way before this PR. The difference now is that the text reaches the shell unquoted, so an absent input turns into a real pipe or redirect. Either render the literal conditionally on the input (like `NfFlag`), or leave shell-literal inputs out of `_safe_absence_names` so an absent value is rejected.

src/sophios/input_output_nf.py
vjaganat90: Fixed in 6df43df. The renderer detects processes driven by the scatter adapter and constructs each unadapted scalar File/Directory parameter as `Channel.value(file(...))`, preserving broadcast semantics. A renderer regression pins the channel shape, and the new R2.17 File runtime variant produced two task outputs from a two-element scatter.

src/sophios/nf_types.py
vjaganat90: Fixed in the parent at 9f8a85f and inherited by this rebased head. Version gates now scan only `processes` and `connections`; `params` remain opaque. The old-schema regression includes `kind`, `is_array`, `stage_as`, `adapter`, and `capture` keys in nested user data.

src/sophios/utils_nf.py
vjaganat90: Fixed in the parent at 9f8a85f and inherited here. Every IWDR target is tracked, including rename-less entries, and a repeated input is rejected in both capability analysis and the defensive lowering path. Tests cover two different renames and bare-plus-rename.

src/sophios/utils_nf.py
vjaganat90: Fixed in the parent at 9f8a85f and inherited here. Both source analysis and executable `NfPort` validation reject `*` and `?` in `stageAs` names, with regressions for both characters.

src/sophios/utils_nf.py
vjaganat90: Fixed in 6df43df. `_flatten_subworkflows` now applies `_WORKFLOW_INPUT_CONSUMED_FIELDS` and `_WORKFLOW_OUTPUT_CONSUMED_FIELDS` to every nested port before inlining, preserving full paths. The aggregate regression covers nested `secondaryFiles` and `pickValue` findings.

src/sophios/utils_nf.py
vjaganat90: Fixed in the parent at 9f8a85f and inherited here. An input-owned `shellQuote: false` token is now presence-sensitive during absence analysis, so a missing optional source is rejected rather than emitting a real pipe or redirect.

src/sophios/input_output_nf.py
darkclad: **Fixed** At `6df43df`, a scalar path parameter that feeds a scattered process renders as `Channel.value(file(...))`, and the File variant of R2.17 pins it at runtime. Non-blocking: `Channel.fromPath` passed `glob: false` and the new `file(...)` call does not, so a broadcast file named like `sample[1].txt` is treated as a pattern. The array branch has the same gap, so `glob: false` in both would close it.

src/sophios/nf_types.py
darkclad: **Fixed** Version gates now walk only `processes` and `connections`, so `params` keys stay opaque. The hydration test covers `adapter`, `stage_as` and `kind` as parameter keys.

src/sophios/utils_nf.py
darkclad: **Fixed** Both `_iwdr_listing_findings` and `_iwdr_stage_as` now track every listed input, rename-less entries included, and reject a repeat. The test covers two renames and a bare entry plus a rename.

src/sophios/utils_nf.py
darkclad: **Fixed** `*` and `?` are rejected in `_iwdr_entryname` and again in `NfPort` validation, with test cases for both characters.

src/sophios/utils_nf.py
darkclad: **Fixed** `_flatten_subworkflows` now runs the workflow input and output field checks on every nested port, with full paths. The test covers nested `secondaryFiles` and `pickValue`.

src/sophios/utils_nf.py
darkclad: **Fixed** An input with a `shellQuote: false` binding is no longer counted as safe to leave absent, so a missing optional value is rejected. The new capability test covers it.

darkclad (approved):
LGTM at `6df43df`.

All six findings are fixed, and I checked each one against the code rather than the thread replies. Each fix has a regression test, and the scatter broadcast fix also has a File variant of the R2.17 runtime test.

Non-blocking: the new `Channel.value(file(...))` construction drops the `glob: false` that `Channel.fromPath` passed, so a broadcast file whose name contains `[` or `{` is treated as a pattern (details on the first thread). Also, `_command` no longer adds `true` in front of a command made only of shell literals. That is right for `printf ok`, but a tool with no `baseCommand` and a lone redirect literal now renders as just the redirect. That tool shape is unusual, so it is only worth a follow-up if it ever comes up.

PR #414, opened 2026-09-09
feat(nextflow): projection-only expressions and two output capture declarations

vjaganat90:
Stacked on #413 (`nf205_scatter_and_composition`). Review after that branch; its commits appear here until it merges.

## What this adds

Two `outputBinding` capture declarations, plus the rule they are admitted under.

**Expression handling is projection-only.** The governing rule this branch adds to §6: a CWL expression is representable only as a *static projection* — a finite compile-time path into a known structure where every step has a provably identical Nextflow counterpart. Nothing is evaluated, and no expression text reaches the executable model or the generated pipeline. The load-bearing half is that a counterpart which is *close but not exact* is not representable either: the gap is never closed with compensating logic, because that logic would itself be computation. I hit that clause immediately — see the trailing-newline finding below.

**Output cardinality.** `outputEval` whose text is exactly `$(self[0])` is read as a cardinality declaration, not an expression: the port carries one value rather than a list. It lowers to a closed marker on the output port and renders as `arity: '1'`, so the author's declaration is stated in the generated pipeline instead of dropped. Admitted only beside a glob that is one literal with no wildcard and no input reference — `self[0]` is a projection only where the match set has one member by construction.

**File-text capture.** `loadContents: true` with `outputEval: $(self[0].contents)` on a `type: string` output lowers to the one `val`-qualified process output, whose value is the file's decoded text. `cwl_adapters/cat.cwl` is exactly this shape, so the repository's own adapter stops being a rejection fixture in the aggregation test and starts lowering.

**Two things real execution taught me here**, both against my first assumption:

- *There is no 64 KiB truncation to reproduce.* The brief called for capturing exactly the first 64 KiB and pinning what happens when the cut lands mid-multibyte-sequence. That is CWL **v1.0** wording. The compiler pins **v1.2**, which replaced it with a hard limit: the file must be 65536 bytes or fewer and must be read *entirely*, with a fatal error above that (`Process.yml` lines 296-302; cwltool reads one byte past the limit and raises). So there is no truncation point and no partial multibyte sequence to define. What I pinned instead is the real contract — whole file, inclusive 65536-byte limit, one byte over fails — with runtime tests at exactly 65536 and 65537.
- *The obvious idiom strips the trailing newline.* Nextflow's `eval` output type is the natural reading of "capture the file's text" and is wrong: it returns `abc` for a file containing `abc\n`. Per the rule above, re-appending a newline was not an option, so the lowering uses a generated helper reading the whole file from the task work dir instead. A revert-check that swaps the helper back to `eval` fails the newline test, so the test really discriminates.

The helper also decodes UTF-8 **strictly** — malformed bytes fail the task rather than becoming U+FFFD — matching cwltool's `str(bytes, "utf-8")`. Groovy's `getText`/`new String(...)` would have substituted silently.

## Boundaries

- **Wildcard globs are rejected, with a stated relaxation path.** `self[0]` under a wildcard makes CWL's and Nextflow's glob match *ordering* load-bearing, and that ordering is unproven. Admitting it needs a demonstration that the two orderings agree across differing filename lengths, mixed case, numeric suffixes where lexical and numeric order disagree (`part2` vs `part10`), and locale-sensitive collation. Without that proof the two runtimes could silently pick *different files*. An input-reference glob is rejected one step removed — its runtime value cannot be shown wildcard-free — and the same proof unlocks it.
- **JavaScript evaluation is ruled out on the record, not deferred.** No evaluator, no JS-to-Groovy transpiler, no embedded engine, no `node` subprocess, nothing evaluated in the renderer. Revisiting it needs a successor design document, mirroring §9's posture toward remote Nextflow execution. Written as a decision so it is not re-litigated as a deferral.
- **`InlineJavascriptRequirement` declared but unused stays legal** — the same shape of decision as #409's `ShellCommandRequirement` finding. The requirement is a declaration; the violation is a non-admitted form. Stated in the design doc so it is not "fixed" later.
- **`.dirname` and `.path` are permanently unattractive, not merely unimplemented.** Under staging a produced file's directory *is* the task work dir, which Nextflow owns and may relocate — never what the author described. A mapping would type-check, run, and mean something else, which is §1's prohibition on silently changing semantics. `.location`, `.checksum`, `.secondaryFiles`, `.format` each get their own named diagnostic too.
- **`type: int` and `type: float` are rejected** for file-text capture: parsing text into a number is computation. That is exactly where the projection-only line falls, and the diagnostic says so.
- **A captured value's only sink is a workflow output.** A process edge out of one is rejected. The reason is about the model, not the lowering: every process output until now carried the `path` qualifier, so the graph never needed a source-to-destination qualifier agreement check for process edges. Adding that check — the process-edge analogue of the §5.2 parameter rule — plus runtime proof of a captured value feeding a downstream `val` port is the self-contained follow-up.
- `nameroot`/`nameext` are **not** here; they were gated on a separate exactness proof.
- `loadContents` on an *input* parameter or input binding is rejected by name — a separate lowering, unimplemented.

## Structural scoping

The narrowness is enforced by code shape, not just prose:

- **One owner.** The admitted forms live in a single module-level frozen table (`_ADMITTED_OUTPUT_EVAL`) mapping exact CWL text to a closed marker. `capture=` is constructed at exactly one call site, from that table's result.
- **Matcher, not evaluator.** Recognition is a dict lookup on the trimmed text — strictly weaker than a tokenizer. Nothing inspects, transforms, or composes expression text. The paired *glob* still goes through the existing template tokenizer.
- **Unrepresentable by type.** `NfPort.capture` accepts only members of `ALLOWED_CAPTURES`, so no future change has anywhere to smuggle expression text — the technique that made `NfFlag` illegal outside command position.
- **Pinned by test in both directions.** `test_the_admitted_output_eval_table_is_closed_in_both_directions` asserts the table equals its expected literal set exactly, so widening it without a `design:` revision fails CI. A 17-shape parametrized rejection corpus (operators, function calls, conditionals, `${return ...}` blocks, string templates, internal whitespace variants) each asserts its exact diagnostic.

The wildcard restriction is deliberately enforced *twice* — capability analysis reports it by path, and the executable model makes the state unrepresentable — with separate revert-checks proving each half independently.

## Schema

`schema_version: 9` for the additive `capture` field on `NfPort`. Hydration accepts 2-9 unchanged; `capture` is optional on hydration (defaults `None`) since no earlier version wrote it, and `_reject_newer_kinds` gates it alongside `is_array`, `stage_as`, and `adapter`. The `loadContents` helper is emitted only into pipelines that use the capture, so artifacts for previously representable models stay byte-identical and existing artifact pairs keep promoting.

## Design authority

Three `design:` commits approve this in order, all under §10 of `design_docs/sophios_nextflow_backend.md`:

- `237ebdc` — projection-only expression handling, and JavaScript evaluation ruled out on the record (§6, Expressions).
- `7bf46d5` — the output cardinality declaration, its wildcard relaxation path, and the `.dirname` rationale (§6, Outputs and globs).
- `a55aafa` — file-text capture, the pinned-v1.2 correction, and the captured-value sink restriction (§6, Outputs and globs).


src/sophios/nf_types.py
```diff
@@ -455,19 +491,37 @@ def from_dict(cls, value: Mapping[str, Any]) -> Self:
         return cls(item["cpus"], item["memory_mb"])
 
 
+GLOB_WILDCARDS = frozenset("*?[")
```
darkclad, line 494: `GLOB_WILDCARDS` only covers `*?[`, but Nextflow's `path` glob also treats `{a,b}` and `\` as special, while CWL's glob matches them literally.

For example, a tool writes a file literally named `out{1,2}.txt` with `outputEval: $(self[0])`. The check passes and the output renders as `path 'out{1,2}.txt', arity: '1'`. Nextflow expands that to `out1.txt` or `out2.txt`, matches nothing, and the task fails, while cwltool returns the file.

Suggest checking against the renderer's `_GLOB_METACHARACTERS` set instead, or rendering capture ports with `glob: false`, since a capture glob is always a plain file name.

src/sophios/nf_types.py
```diff
@@ -868,19 +1008,23 @@ def _connection_from_dict(value: Mapping[str, Any]) -> NfConnection:
...
     # Version an additive non-kind-tagged field was introduced in, keyed by
-    # the field name it appears under. is_array predates a "kind" tag on
-    # NfPort, so it needs its own gate alongside KIND_SCHEMA_VERSIONS.
-    FIELD_SCHEMA_VERSIONS: ClassVar[Mapping[str, int]] = MappingProxyType({"is_array": 5})
+    # the field name it appears under. is_array, stage_as, and capture
+    # predate a "kind" tag on NfPort, and adapter is additive on an existing
+    # connection kind, so each needs its own gate alongside
+    # KIND_SCHEMA_VERSIONS.
+    FIELD_SCHEMA_VERSIONS: ClassVar[Mapping[str, int]] = MappingProxyType(
+        {"is_array": 5, "stage_as": 7, "adapter": 8, "capture": 9}
```
darkclad, line 1028: The `capture: 9` gate is checked against any `capture` key in the payload, including the workflow's `params`, not only port entries.

For example, a schema version 8 workflow with an input named `capture` stores `params: {"capture": ...}`. It loaded before this PR; now it fails with "'capture' requires executable Nextflow schema version 9". That contradicts the PR description's claim that versions 2 through 8 still load unchanged.

Suggest applying the field gates only to port entries.

darkclad (commented):
Reviewed the three commits on top of #413 (`20f0251..4ddeb86`). Two issues inline, both in `src/sophios/nf_types.py`.

src/sophios/nf_types.py
vjaganat90: Fixed in 0b7b9f8, with the contract recorded first in 1f6eeff. Every `single` capture now renders `glob: false` before `arity: '1'`, so CWL-literal braces and backslashes cannot become Nextflow patterns. Renderer tests cover both characters, reader promotion still round-trips, and R2.18 now executes with a file literally named `out{1,2}.txt`.

src/sophios/nf_types.py
vjaganat90: Fixed in the parent at 9f8a85f and inherited by this rebased head. Schema gates scan only executable `processes` and `connections`, never `params`. The regression hydrates an old schema with nested user keys including `capture`, `adapter`, `stage_as`, `is_array`, and `kind`.

src/sophios/nf_types.py
darkclad: Verified in 0b7b9f8. Single captures now render `glob: false`, and the R2.18 runtime test producing `out{1,2}.txt` closes the case I raised.

src/sophios/nf_types.py
darkclad: Verified in 9f8a85f. The gate now walks only `processes` and `connections`, so user keys under `params` no longer trip it, and the regression test covers it.

darkclad (approved):
LGTM

PR #415, opened 2026-09-15
Spec 2 sweep-up: defects the chain surfaces but no PR in it owns

vjaganat90:
**Stacks on #412 (`semrefac_15`)** — review the chain in order.

This branch is the sweep-up for defects the Spec 2 chain turns up that no PR in it owns. Each one here was found while reviewing another PR, reproduces on `master`, and belongs to a subject none of #407–#412 is about; putting it in one of them would make that PR's claim harder to read rather than more complete. Expect this branch to grow as the chain is reviewed.

## The last producer of the removed step form

`cwl_subinterpreter.rerun_cwltool` builds a single-step workflow twice:

```python
root_yaml_tree = {'wic': wic_steps, 'steps': [{cwl_tool: None}]}
...
yml = {'steps': [{cwl_tool: args_vals_new}]}
```

That is the sequence-of-single-key-mappings form. §3.1 admits two step forms and this is neither, so a producer of it emits a document the language rejects — the same defect #408 fixed in `insert_step_into_workflow`, and the last instance of it in the tree.

Not a regression from #407: `utils.require_step_id` already refused both documents, which means the speculative rerun could not have been working. #407 makes it a *documented* error rather than an accident, which is what makes it worth closing now.

Both sites write the `id:` form the rest of the compiler reads.

## Unknown tags on structural nodes

`_reject_unknown_tag` exists precisely so the rule has one home. Its docstring:

> It used to be written out at each position that consumes a node … and a third position, mapping keys, simply never got a copy. Restating a rule per position is how a position gets missed; now a new one has a single obvious thing to call.

Four positions never got the call — the ones that make up a document's own structure rather than its keys or values:

```yaml
steps: !foo {}                    # the steps: node
- !foo {id: s}                    # one sequence entry
touch: !foo {in: {}}              # a step body, mapping form
in: !foo {f: !ii 1}               # the in: mapping
```

All four parsed clean. The loader rejects every one, so this was the parser accepting documents the loader refuses — the direction `_key_text`'s docstring says the language never takes, and the same asymmetry as the exported schema accepting what the parser rejects.

Each new case asserts both halves: that the parser reports `wic009`, and that `yaml.load` raises on the same source. Pinning only the parser side would let the two drift apart again, which is how these four were missed in the first place.

## Scope

`rerun_cwltool` itself is not executed here — it speculatively runs a CWL runner against a cache directory. The test pins the two document shapes it constructs, and says so.

## Design-doc timelessness

One commit here is not a defect of that shape: the locked document pinned its baseline to a bare commit hash and carried a date and a test's history in prose, which the chain's own edits to §5.5 and §6.3 made visible. It rides on this branch because this is the PR in flight that touches that file. The baseline now names the state — the compiler before Spec 0 — with the hash beside it so the document's present-tense descriptions can be checked against something.


darkclad (commented):
Review at `4140bb2`. Scope: the two commits on top of #412 (`43a6110..4140bb2`), which touch `cwl_subinterpreter.py`, `lang/parser.py`, the design doc, and two test files. I installed the head in a clean venv and ran `test_lang_parser.py` and `test_compiler.py` (136 passed). Then I reverted each fix separately to see whether its test catches the revert.

Both code fixes are correct. The `id:` form is a faithful translation of `{cwl_tool: body}`, and the four new `_reject_unknown_tag` calls close the four positions they name. The new parser tests work: with the `parser.py` hunks reverted, all four cases fail.

**Two findings, both inline:**

- The subinterpreter test passes with the fix reverted. It checks two dict literals written in the test, not the documents `rerun_cwltool` builds. A test that calls the function itself is cheap, and the suggested version is verified in both directions.
- Eight more structural positions still accept an unknown tag that the loader rejects: `out:` and its entries, `wic:`, `wic: steps:`, nested sidecar steps, and the root document.

Two non-blocking notes are also inline: the order of the `id` spread, and the design doc baseline.

The PR description doesn't cover the design doc change (`4140bb2`). It isn't a defect that reproduces on `master`, so one sentence on why it belongs on this branch would help whoever reads the chain later.

tests/core/test_compiler.py
```diff
@@ -24,3 +26,85 @@ def test_an_inserted_step_is_resolvable_by_id() -> None:
...
+    `cwl_subinterpreter.rerun_cwltool` was the last producer of a single-key
+    sequence step in the tree — the same defect as the inserted step above, and
+    `require_step_id` refused both documents it built. §3.1 no longer admits
+    that form at all, so a producer of it emits a document the language rejects.
+
+    Pins the shape rather than executing `rerun_cwltool`, which speculatively
+    runs a CWL runner against a cache directory; the two literals here are the
+    two it builds.
+    """
+    assert utils.get_steps_keys(document['steps'])[0] in ('tool.wic', 'tool')
```
darkclad: **🟠 This test passes with the fix reverted.**

It asserts on two dict literals written here, and nothing ties them to `cwl_subinterpreter.py`. I checked out that file from #412, which has the old `[{cwl_tool: None}]` and `[{cwl_tool: args_vals_new}]`, and ran `-k resolvable_by_id`: 3 passed. The docstring says the literals are "the two it builds", but nothing enforces that. Also, `in ('tool.wic', 'tool')` accepts either name for either document.

The docstring skips calling `rerun_cwltool` because it runs a CWL runner. But the documents are built before that, and each branch hands its document to one function you can monkeypatch: `compiler.compile_workflow` for a tool, `ast.read_ast_from_disk` for a `.wic`. This version passes at `4140bb2`. With the fix reverted, both cases fail with `ValueError: Error! Each step dictionary must contain a non-empty string id: tag.`

```python
from sophios import ast, compiler, cwl_subinterpreter


class Captured(Exception):
    pass


@pytest.mark.fast
@pytest.mark.parametrize(('cwl_tool', 'module', 'name', 'tree_arg'), [
    ('tool', compiler, 'compile_workflow', 0),
    ('tool.wic', ast, 'read_ast_from_disk', 1),
])
def test_rerun_cwltool_builds_an_id_form_step(monkeypatch, tmp_path, cwl_tool, module, name, tree_arg):
    seen = []

    def capture(*args, **_):
        seen.append(args[tree_arg].yml)
        raise Captured

    monkeypatch.setattr(module, name, capture)
    with pytest.raises(Captured):
        cwl_subinterpreter.rerun_cwltool('', tmp_path, tmp_path, cwl_tool, {'in': {}}, {}, {}, None, tmp_path)
    assert [utils.require_step_id(s) for s in seen[0]['steps']] == [cwl_tool]
```

`Captured` isn't a `FileNotFoundError`, so the function's own `except` doesn't swallow it. The test never reaches `sub.run` and doesn't touch the cache directory.

src/sophios/lang/parser.py
```diff
@@ -187,6 +187,7 @@ def _steps(node: yaml.nodes.Node, file: str, diags: Diagnostics) -> tuple[tuple[
     Both surface forms are long-standing and remain supported; the AST is the
     same either way, so nothing downstream needs to care which was written.
     """
+    _reject_unknown_tag(node, file, diags)
```
darkclad: **🟠 Eight more structural positions still accept a tag the loader rejects.**

These calls close the four positions the change names, but those aren't all of them. At this head, `parse()` returns `ok=True` with no diagnostics for each document below, and `yaml.load(..., Loader=wic_loader())` raises on every one:

```yaml
--- !foo                          # the root document
steps: []
```
```yaml
steps:
- id: s
  out: !foo [a]                   # the out: node
```
```yaml
steps:
- id: s
  out: [!foo a]                   # a scalar out: entry
```
```yaml
steps:
- id: s
  out:
  - !foo {a: !& e}                # a mapping out: entry
```
```yaml
wic: !foo {}                      # the wic: node
```
```yaml
wic:
  steps: !foo {}                  # wic: steps:
```
```yaml
wic:
  steps:
    (1, s): !foo {wic: {}}        # a sidecar step value
```
```yaml
wic:
  steps:
    '(1, s)':
      wic: !foo {}                # the nested wic: wrapper
```

None of the functions that consume those nodes calls `_reject_unknown_tag`: `parse` for the root, `_outputs`, `_output_binding`, `_sidecar` (both its own node and the `steps:` value), and `_child_sidecar_node`.

The rule has one home now, but every position still has to remember to call it, and this change shows how easy that is to miss. You could stop tracking positions altogether with one pass in `parse()`, right after `yaml.compose`. Walk every node once with an `id()` visited set, which stays linear in distinct nodes so aliases cost nothing extra, and report each unknown tag. That covers every position, including future ones, and the per-position calls could go. If you'd rather keep the per-position calls, these eight cases fit straight into `STRUCTURAL_TAGS`, which already checks the parser and the loader together.

src/sophios/cwl_subinterpreter.py
```diff
@@ -94,7 +94,7 @@ def rerun_cwltool(homedir: str, _directory_realtime: Path, cachedir_path: Path,
             yaml_tree = ast.python_script_generate_cwl(yaml_tree, Path(''), tools_cwl)
             yml = yaml_tree.yml
         else:
-            yml = {'steps': [{cwl_tool: args_vals_new}]}
+            yml = {'steps': [{'id': cwl_tool, **args_vals_new}]}
```
darkclad: **🟡 Nit: a `config` with an `id` key now silently renames the step.**

In `{'id': cwl_tool, **args_vals_new}` the spread wins, so if `args_vals` (`json.loads(args.config)`) has an `id` key, it replaces `cwl_tool`. The step then resolves to a different tool, with no diagnostic. `{**args_vals_new, 'id': cwl_tool}` keeps the tool the watcher was started for. Unlikely in practice, and the fix is a one-token reorder.

design_docs/core-refactor-design.md
```diff
@@ -1,7 +1,8 @@
 # Sophios Core Refactor — Design
 
 **Status:** Approved. This design is locked; changes require a new revision.
-**Baseline:** `master` at `6570369`
+**Baseline:** the compiler as it stood before Spec 0. Present-tense descriptions
+of existing behaviour below refer to that state, not to any later one.
```
darkclad: **🟡 The old baseline could be checked out; the new one can't.**

`6570369` is reachable from `master` (95 commits back), so a later reader can use it: `git show 6570369:src/sophios/compiler.py` is exactly the state these present-tense descriptions refer to. "The compiler as it stood before Spec 0" makes the reader work out which commit that was, and that gets harder once the chain is squash-merged. Naming the state is a real improvement, so consider keeping the hash next to it, e.g. "the compiler as it stood before Spec 0 (`master` at `6570369`)".

§5.5 is similar. `9758e81` is still cited in `docs/sophios_language_reference.md:173` and in the `_sequence_step` docstring, so the design doc now sends the reader elsewhere for a fact it used to state itself. Not blocking.

tests/core/test_compiler.py
vjaganat90: You are right, and this is the more useful kind of finding — the test asserted on literals I had copied into it, so it was pinning my transcription rather than the code. I had flagged the gap in the docstring and then written the test the gap describes, which is worse than not claiming coverage at all.

Your version is what landed, near enough verbatim: stub the one function each branch hands its document to, let `_Captured` escape the `FileNotFoundError` handler, and assert `require_step_id` reads back the tool name. Reverting either construction now fails its branch, which the old test did not.

The `in ('tool.wic', 'tool')` looseness is gone too — each row asserts its own name.

src/sophios/lang/parser.py
vjaganat90: Taken the first of your two options: one walk in `parse()` over the composed graph, and the per-position calls are gone.

All twelve positions report now — your eight plus the four the change already named — and `_reject_unknown_tag` with its nine call sites is deleted. The walk tracks nodes by `id()`, so aliases cost nothing and a recursive alias cannot loop it. `STRUCTURAL_TAGS` carries all twelve, each still asserting both halves.

One consequence worth naming: with tags reported in one pass and structure in another, diagnostics arrived in pass order rather than reading order — `wic009` on line 4 ahead of `wic006` on line 2. `parse()` now sorts by position before returning, which is a single rule that holds however many passes there are, and it restored the interleaving the body tests already pinned. The one test whose order genuinely changed is the collection-key case, where the step's span opens before its key's.

src/sophios/cwl_subinterpreter.py
vjaganat90: Fixed — `{**args_vals_new, 'id': cwl_tool}`. You are right that it is unlikely, but the failure mode is silent retargeting at a different tool, which is worth a token.

design_docs/core-refactor-design.md
vjaganat90: Good point, and taken: the line reads `the compiler as it stood before Spec 0 (\`master\` at \`6570369\`)` now, with a clause saying the hash is there so the present-tense descriptions can be checked against it. Naming the state and giving the hash are not in tension — the prose is what survives, the hash is what makes it verifiable.

§5.5 I have left pointing at the language reference. The hash there was a date and a commit in the middle of an argument, which is what made the section read as dated; the reference is a living document and carries both.

vjaganat90:
On the PR description not covering `4140bb2` — fair, and now stated. The branch's remit is defects the chain surfaces that no PR in it owns, and a design-doc correction is not that: it is timelessness housekeeping on the locked document, which the chain's own edits to §5.5 and §6.3 made visible. It rides here because this is the branch in flight that touches that file, not because it belongs to a category.

The body says so, and names what the commit does.

tests/core/test_compiler.py
```diff
@@ -24,3 +31,129 @@ def test_an_inserted_step_is_resolvable_by_id() -> None:
...
+    seen: list[Yaml] = []
+
+    def capture(*args: Any, **_: Any) -> None:
+        seen.append(args[tree_arg].yml)
+        raise _Captured
+
+    monkeypatch.setattr(module, name, capture)
+    with pytest.raises(_Captured):
+        cwl_subinterpreter.rerun_cwltool(
+            '', tmp_path, tmp_path, cwl_tool, {'in': {}}, {}, {}, None, tmp_path)
```
darkclad: **🟡 Nothing pins the id order yet.**

The config here has no `id`, so the reorder on `cwl_subinterpreter.py:99` is untested: with `{'id': cwl_tool, **args_vals_new}` restored, all 10 tests in this file still pass. Giving the tool row a config with its own `id` pins it. I checked that this passes at `db0b6f5` and fails the `tool` case when the order is swapped back:

```python
@pytest.mark.parametrize(('cwl_tool', 'module', 'name', 'tree_arg', 'config'), [
    ('tool', compiler, 'compile_workflow', 0, {'id': 'elsewhere', 'in': {}}),
    ('tool.wic', ast, 'read_ast_from_disk', 1, {'in': {}}),
])
```

Add `config: Yaml` to the signature and pass it in place of `{'in': {}}`.

darkclad (commented):
Round 2 at `db0b6f5`. #412 was rebased under this branch, so I compared against its new head (`4f311ad`); the delta is still the two amended commits. I ran `test_lang_*.py` and `test_compiler.py` at this head: 237 passed. The one failure, `test_lang_version::test_annotation_is_declared_and_the_cwl_stays_valid`, is `spython` importing `pwd` on Windows. That is the same error the Windows CI job shows at #412's head, and it has nothing to do with this PR.

**All four round-1 findings are fixed.** I checked each one by reverting it:

- With the `_report_unknown_tags` call removed, 27 parser tests fail.
- With `cwl_subinterpreter.py` reverted to #412's version, both `rerun_cwltool` cases fail.
- With the new sort removed, the 4 order-sensitive parser tests fail.

Verdicts are on the threads. The PR description now covers the design doc change.

**Four small follow-ups inline, none blocking:** nothing pins the id order, one comment and two docstrings still describe the code as it was before the walk, and the `NOT_A_MAPPING` early return skips the new sort.

src/sophios/lang/parser.py
```diff
@@ -426,7 +470,6 @@ def _input_value(node: yaml.nodes.Node, file: str, diags: Diagnostics) -> InputV
     # here would report `wic019` alone in input position while passthrough
     # reported both, which is the two-surfaces-one-language divergence this
     # rule exists to prevent.
```
darkclad: **🟡 This comment block describes a check that is no longer here.**

"Tag check first, in the same order `_opaque` uses" sat above `_reject_unknown_tag(node, file, diags, span)`, which this change deleted, so it now reads as if it explains `_is_edge_def`. The concern it records is still met, but for a different reason: `_report_unknown_tags` covers the whole graph before `_document` runs, so `!foo {wic_anchor: x}` gets `wic009` in every position, whatever this function checks first. Either delete the block or cut it to one line saying that.

tests/core/test_lang_parser.py
```diff
@@ -495,6 +495,44 @@ def test_a_named_entry_reports_what_the_mapping_form_reports(claim: str, source:
     assert rejected == [Code.MISSING_STEP_ID, *baseline], claim
 
 
+#: A tag on a node that makes up the document's *structure*, rather than on a
+#: key or an input value. The loader rejects all four, so accepting them is the
+#: specification being more permissive than the thing it specifies.
```
darkclad, line 500: **🟡 The counts here are left over from round 1.**

`STRUCTURAL_TAGS` has twelve entries, but this comment still says "The loader rejects all four". The two docstrings also give different numbers from what happened. The test's docstring (line 523) says "the first pass covered three, left three uncovered, and a second pass found eight more", and `_report_unknown_tags` (`parser.py:106`) says it "missed three positions on its first pass and eight on its second". Round 1 added four positions, not three: `steps:`, a sequence entry, a mapping-form step body, and `in:`. Dropping the numbers would keep them from going stale again, e.g. "the loader rejects every one" and "each pass found positions the previous one missed".

src/sophios/lang/parser.py
```diff
@@ -124,7 +162,13 @@ def parse(text: str, filename: str = '<string>') -> ParseResult:
         )
         return ParseResult(None, diagnostics)
```
darkclad: **🟡 This early return skips the new sort.**

The walk now runs before this check, so a root that is not a mapping can report a tag first and then `NOT_A_MAPPING` at an earlier column: `parse('- !foo a\n')` returns `UNKNOWN_TAG` at 1:3, then `NOT_A_MAPPING` at 1:1. It is the only path in `parse()` that can return more than one diagnostic unsorted. A small `_in_reading_order(diagnostics)` helper used by both returns would keep the rule in one place, which is what the new comment below asks for.

tests/core/test_compiler.py
vjaganat90: Taken, near enough verbatim — the `tool` row's config now carries `id: elsewhere`, and the docstring says why one row differs. Reverting the key order fails that row and leaves the `.wic` one passing, which is the split you'd expect since only the tool branch merges a config into the step.

src/sophios/lang/parser.py
vjaganat90: Cut to one line. The ordering argument belonged to a check that is no longer in this function; what is worth keeping is that the walk covers the whole graph before `_document` runs, so `!foo {wic_anchor: x}` earns `wic009` here as it does anywhere else, whatever order this function checks in.

tests/core/test_lang_parser.py
vjaganat90: Dropped the numbers in all three places — the table comment, the test docstring, and the `_report_unknown_tags` docstring. They were counting the passes of a form that no longer exists, so they could only ever go stale; "each pass found positions the previous one missed" is the part that stays true.

src/sophios/lang/parser.py
vjaganat90: Fixed, and your repro was exact. `_in_reading_order` now wraps both returns.

I also gave the rule a test rather than leaving it pinned by the interleaving of the body cases: a table over the three shapes that produce more than one diagnostic — the non-mapping root, a tag above a structural error, and one below — asserting the positions come back sorted. Reverting either return fails it, the early return on the first case alone.

tests/core/test_ci_coverage.py
```diff
@@ -80,17 +80,35 @@ def _pytest_runs(workflow: Path) -> list[tuple[list[str], str | None, str | None
...
+    tokens shaped like paths are considered, and one that does not resolve
+    raises rather than being dropped. Dropping the last of them leaves no paths
+    at all, which reads as the whole rootdir again: the same silent pass, by a
+    different route. A node id keeps only its file part.
+    """
+    paths: list[str] = []
+    for token in tokens:
+        if token.startswith('-') or not ('/' in token or '.py' in token):
+            continue
+        path = token.split('::', 1)[0]
```
darkclad: **🟡 Non-blocking: a node id still widens the run, now to its file instead of the rootdir.**

`split('::', 1)[0]` keeps the file and drops the test name, and `_path_reaches` then counts every marked test in that file as collected. The commit message names this failure ("a node id would make a narrow run appear to cover every marked test"), but the fix only narrows it from the rootdir to the file. Say a main lane had this step:

```
pytest tests/core/test_leak_boundary.py::test_step_passthrough_is_byte_identical -m skip_pypi_ci
```

`_paths_of` returns `['tests/core/test_leak_boundary.py']`, so the census counts all 9 `skip_pypi_ci` tests in that file as covered while only one runs. No workflow uses a node id today, so nothing is wrong on `master`. But the failure would be a silent pass, which is exactly what this file exists to catch.

Two ways to close it:

- Carry the name through and require `test == name` next to `_keyword_admits`.
- Raise on `::` the way an unreadable `-k` raises, until a lane actually needs node ids. This is one line, and the third assert in the new test becomes a `pytest.raises`.

darkclad (approved):
LGTM at `0a3640c`.

#412 has merged, so this PR now diffs against `master`. The two commits from round 2 came through the rebase unchanged, so this round is `80b857f` and `0a3640c`. I ran `test_ci_coverage.py`, `test_lang_parser.py` and `test_compiler.py` at this head: 164 passed.

All four round-2 follow-ups are fixed. I checked the two that change behaviour by reverting them: with the early return unsorted, the non-mapping-root case of the new reading-order test fails, and with the id order swapped back, the `tool` row fails. `80b857f` is the right direction too: a path-shaped token that doesn't resolve now raises instead of widening the run to the whole rootdir, and the new test fails without the raise.

One non-blocking note inline: a node id still widens the run to its whole file. No workflow uses one today, so it can go in a follow-up.

tests/core/test_ci_coverage.py
vjaganat90: Closed rather than deferred — it is on the branch as `9aa2147`.

I took the first of your two options. `_paths_of` keeps the node id whole, validating only the file part against disk, and `_path_reaches` gained the test name so it can hold a node id to the test it names — reading past a class segment and a `[param]` suffix to find it. Your `test_leak_boundary.py::test_step_passthrough_is_byte_identical` example now reaches that test and none of its eight siblings.

I went with carrying the name rather than raising because a node id is legal pytest that this file would otherwise forbid; the refusal idiom fits an expression whose *effect* is unknown, and a node id's is not. It cost about a dozen lines over the one-line raise, which seemed the better side of that trade.

The reach table covers it in both directions, plus the class and parametrised spellings, and one case where a narrowing argument sits beside an argument naming the file whole — the wider one has to win. Reverting either half fails a row.

tests/core/test_compiler.py
darkclad: **Fixed** The test now calls `rerun_cwltool` and reads each document from the function it is handed to. With `cwl_subinterpreter.py` reverted to the old shapes, both cases fail, so the test now catches the regression.

src/sophios/lang/parser.py
darkclad: **Fixed** The single walk in `parse()` reports all twelve positions, and with its call removed 27 parser tests fail. The position sort that came with it is now one `_in_reading_order` helper and has its own test.

src/sophios/cwl_subinterpreter.py
darkclad: **Fixed** The id now comes last, and the `tool` row's config carries its own `id`, so swapping the order back fails that row.

design_docs/core-refactor-design.md
darkclad: **Fixed** The baseline now names the state and gives the hash, which gets both. Leaving §5.5 pointing at the language reference is reasonable, since `docs/sophios_language_reference.md` still names `9758e81`.

tests/core/test_compiler.py
darkclad: **Fixed** At `0a3640c`, swapping the order back fails the `tool` row and leaves the `.wic` row passing, as you describe.

src/sophios/lang/parser.py
darkclad: **Fixed** The one-line replacement says what is true now: the walk has already reported the tag before `_input_value` runs.

tests/core/test_lang_parser.py
darkclad: **Fixed** All three places now state what holds instead of counting passes, so there is nothing left to go stale.

src/sophios/lang/parser.py
darkclad: **Fixed** Both returns go through `_in_reading_order`. With the early return left unsorted, the non-mapping-root case of `test_diagnostics_come_back_in_reading_order` fails, so the new test catches it.

tests/core/test_ci_coverage.py
darkclad: **Fixed** At `9aa2147`, your `test_leak_boundary.py` example reaches only `test_step_passthrough_is_byte_identical`, not the other 8. Crediting a node id to the test it names fits better than refusing it, and undoing either half fails a row. A class-only node id reaches nothing, so the census reports those tests loudly instead of silently crediting them.

darkclad (approved):
LGTM at `9aa2147`.

This covers the node-id fix since my approval at `0a3640c`. `_path_reaches` now credits a node id only to the test it names, so the `test_leak_boundary.py` example reaches 1 of its 9 `skip_pypi_ci` tests instead of all of them. Undoing either half of the fix fails a test. I ran `test_ci_coverage.py`, `test_lang_parser.py` and `test_compiler.py` at this head: 169 passed.

PR #416, opened 2026-09-15
Make lint_and_test green on master: the type errors, and cwltool on Windows

vjaganat90:
`lint_and_test` is red on `master`. This makes it green, and then closes the gap that let it go red unnoticed.

## 1. Two type errors the lane rejects and the hook accepts

```
tests/core/test_compiler.py:69: Argument 8 to "rerun_cwltool" has incompatible type "None"; expected "Draft202012Validator"
tests/core/test_lang_parser.py:550: Item "None" of "SourceSpan | None" has no attribute "start_line"
```

The lane runs `mypy src/ examples/ tests/` with the project's mypy; the pre-commit hook runs a pinned one, in an isolated environment, over only the staged files. They disagree, and the hook is the weaker of the two.

- `rerun_cwltool`'s `validator` is not optional. Neither branch of that test reaches it, but a real `Draft202012Validator({})` is honest where `None` was not, and needs no cast to quiet the checker.
- A diagnostic's `span` is `SourceSpan | None`. The order test now counts the diagnostics it actually read a span from, so `_in_reading_order`'s claim that every parse diagnostic carries one is asserted rather than assumed.

## 2. cwltool on Windows

`import cwltool.main` pulls in `spython`, which imports `pwd` at module scope. Ten tests across `test_hermeticity.py`, `test_leak_boundary.py` and `test_lang_version.py` therefore fail on the Windows leg with `ModuleNotFoundError`.

These are not long-standing failures. Until a step named their files they ran **nowhere**, so the platform question had never been put to them. Collecting them was right; running them on Windows is not.

The invariant is that a test calling cwltool in-process cannot run off POSIX, whatever lane names it — so the test declares the dependency with `needs_cwltool`, and `conftest.py` owns what that dependency costs and skips accordingly. A step-level `if: runner.os == 'Linux'` would have fixed the one step, left the next cwltool test to rediscover the same thing, and dropped the Windows run of the parser and predicate tests in that step, which pass there.

`test_every_cwltool_test_declares_that_it_needs_cwltool` holds the halves together: a test whose body imports cwltool without the marker fails the census rather than one leg of a matrix twenty minutes later. It found all five existing sites, including two in `test_canonical_path.py` that were not in the failure list.

## 3. The example surfaces nothing ran

`examples/scripts/` was documentation that happened to be executable. CI type-checked, linted and formatted it and never executed a file: the step named "Generate Sophios Python API Workflows" runs `test_compile_python_workflows`, which discovers through `search_paths_wic` — the corpus repositories and `docs/tutorials` — and `docs/tutorials` holds no `.py` files at all. An example could stop compiling with every lane green.

Three scripts were carrying claims no test made. The claims are now `tests/core/test_example_scripts.py` and the scripts are gone:

| Surface | Why it is new coverage |
|---|---|
| `Step.when` | A conditional dropped on the way to the CWL silently makes the step unconditional. `grep -rn "\.when\b" tests/` returned nothing before this. |
| Single-input `scatter_on` | Existing scatter tests always pass `method=`. The default for one port is `dotproduct`; losing the key runs the step once on the whole array. |
| Compiled document vs. golden file | The other tests assert one key each; this compares the whole emitted workflow, its job inputs and its name. |

`tool_builder_workflow.py` stays in `examples/`: it has a walkthrough page that links it and tells a reader to run it. The test imports it rather than copying it, so there is still exactly one of it — and what it adds is the two-tool chain whose output is a string read from a file (`load_contents` with an `output_eval`), which must survive composition for the downstream `outputSource` to mean anything.

`scattering_scaling.wic` was a fixture of `test_scattering_scaling.py` that lived in `examples/`. It moves next to its test and is addressed relative to it.

**The golden file had drifted.** It predated the `sophios:lang_version` annotation, and nothing ran the comparison that would have said so — which is the case for adding the test, made by the test itself. Regenerated at its original formatting, so the diff is the two lines that actually changed.

These compile and never execute: no container, no runner, no network, about a second in total, in their own named step.

## Not addressed

`examples/secrets_echo.wic` and `examples/test_nvidia_smi.wic` are referenced by no test, workflow or page, and are reachable only through `config_basic.json`'s `search_paths_wic`. Their requirements are a secrets store and a GPU, not a platform — left alone here.


vjaganat90:
Second cause of the red lane, pushed as `36e0503`.

`lint_and_test` is red on `master` for two independent reasons, not one. The first was the type errors. The second is the Windows leg: `import cwltool.main` pulls in `spython`, which imports `pwd` at module scope, so ten tests across `test_hermeticity.py`, `test_leak_boundary.py` and `test_lang_version.py` fail there with `ModuleNotFoundError`.

I had called this pre-existing. It is not — those tests ran **nowhere** until the census step named their files, so the platform question had never been asked of them. Collecting them is right; running them on Windows is not.

The invariant is that a test calling cwltool in-process cannot run off POSIX, whatever lane names it. So the test declares the dependency with `needs_cwltool`, and `conftest.py` owns what the dependency costs and skips accordingly. A step-level `if: runner.os == 'Linux'` would have fixed this one step and left the next cwltool test to rediscover the same thing — and it would also have dropped the Windows run of the parser and predicate tests in that step, which pass there and are worth keeping.

`test_every_cwltool_test_declares_that_it_needs_cwltool` holds the two halves together: a test whose body imports cwltool without the marker fails the census, rather than one leg of a matrix twenty minutes later. It caught all five existing sites, including the two in `test_canonical_path.py` that the failure list did not reach.

Verified by forcing the platform predicate true — the eight stub cases skip rather than error — and by removing a marker, which fails the new census test.

PR #417, opened 2026-09-16
Spec 3 prerequisite: the language's opinion of the documents the compiler manufactures, and the defect it exposed

vjaganat90:
**Stacked on #416** — that PR makes `lint_and_test` green again, and this branches off it rather than off a red `master`.

Two commits. The first adds a property; the second fixes a language defect that property's own verification exposed.

---

# 1. Every document the compiler manufactures is one the language accepts

P07 already says every reachable corpus `.wic` parses, so documents a **user** wrote are covered. What nothing checked is the documents the compiler **makes**: the Python API's output, the inliner's intermediates, the tree after `wic:` metadata is merged onto a step, the documents `rerun_cwltool` builds. Those never pass through the front end, so the grammar has never had an opinion about them.

That class keeps producing defects — three in as many PRs:

| Defect | PR |
|---|---|
| `workflow_document` spelled step ids from the wrong stem | #412 |
| `cwl_subinterpreter` still emitted a step form the grammar had removed — a producer that outlived its own syntax | #415 |
| Scatter readable from two places with no owner | #407 |

§6 of the language reference states the adherence claim; today it is checked only for documents on disk. This quantifies it over manufactured ones.

### Why before the IR rather than during it

Spec 3 freezes `Lower`'s input type — the AST. If the AST cannot represent what the compiler actually generates, that invalidates the IR design rather than costing a patch. This buys the answer for the price of a test instead of a rewrite, and it is the cheap half of what T3.6 does for real.

### How the claim is made exhaustive

`MANUFACTURING_SITES` pins all 26 places in `src/sophios` that build or rewrite a document-shaped object, each classified `DOCUMENT`, `SCHEMA`, `RENDERER`, `TOOL` or `CONTRIB`. A static AST scan fails when a site appears that is not listed, so the inventory cannot fall behind the code silently.

Every `DOCUMENT` site is instrumented; anything document-shaped passing through it — arguments included, since several rewrite in place and return nothing — is dumped and parsed.

**Sites no driver reaches are named, not ignored.** `UNREACHED` lists each with its reason, and a test asserts the unreached set is *exactly* that, so a site that stops being exercised fails too rather than quietly becoming evidence nobody gathers.

### What the scan cannot see, stated rather than implied

It matches two syntactic shapes: a dict literal carrying `steps`/`wic`, and an assignment into one of those keys. Below that floor, by reading the call graph: helper indirection (`inference.perform_edge_inference` reaches a step's `in:` through two further calls into `utils_cwl`), `maybe_add_requirements`, which writes a key none of the markers name, and every `mergedeep.merge` whose result set is decided at runtime. Those manufacture *into* documents the instrumented sites already hand over, so the parse claim still covers them; what the scan cannot promise is that a brand-new site built that way announces itself.

---

# 2. The defect that verifying it exposed

Mutation-verifying the property meant injecting malformed documents at a live manufacturing site and checking each was caught. Two were. One was not:

| Injected at a reached site | Caught |
|---|---|
| A step in the removed single-key form (the #415 defect) | yes |
| A step with no `id` (wic006) | yes |
| An unknown desugared key, `wic_not_a_thing` | **no** |

The third is not a hole in the test. It is a hole in the language.

**The two spellings were equally *accepted* and unequally *safe*.** §6.1 says `!ii v` and `{wic_inline_input: v}` are equivalent, and for what they accept, they are. For what they can reject, they never were:

- A tag is a **closed namespace**. `!iii` is `wic009` at once, in all twelve positions.
- A desugared key shares its namespace with **passthrough CWL**, which §1 declares open by definition. `wic_inline_inpt` was indistinguishable from a key the compiler should carry through untouched — so the construct silently vanished and the typo rode into the emitted document.

Both spellings are hand-written, so that is a human mistake as much as a generated one.

### `wic024`

A single-key mapping in input position whose key begins `wic_` and is not one of the constructs is now reported. It lives beside the desugared table, which is the one home those spellings already have.

`lang_version` does not move: nothing depends on this grammar yet, so this is a defect in an unreleased language rather than a change to a released one.

### Construct position only — and the generators enforced that

The first version walked the whole graph and reserved `wic_` for **every** mapping key. It passed six hand-written position tests. Then `test_round_trip_preserves_structure` failed on:

```yaml
steps:
  _:
    in:
      wic_: !* _
```

An input port *named* `wic_`. Reserving a prefix in name position narrows the language well past the defect being closed — names are the user's to choose. The rule moved into `_desugared_form`:

| | |
|---|---|
| `in: {f: {wic_inline_inpt: 1}}` | **wic024** |
| `in: {wic_: !* e}` — a port named `wic_` | accepted |
| `in: {wic_port: {wic_inline_input: 1}}` | accepted |
| `hints: {wic_hint: 1}` — passthrough | accepted |
| `out: [{f: {wic_anchr: e}}]` | already caught, by `wic005` — a mapping there is not a name |

Mutation-verified both directions: removing the call fails 5 tests, removing the prefix guard — the over-reach — fails 2.

---

## Cost

The property is ~3s in its own `lint_and_test` step; a red build there means the compiler emitted something the language refuses, which is a different fact from a property failing and worth reading without bisecting. `wic024` adds no measurable time — it is a dictionary miss on a path already taken.

## One thing worth a reviewer's eye

The language reference previously said "**Write the tagged form.** The desugared form is what tooling emits." Both spellings are in fact authorable — the Python API simply skips the sugar, and every layer Sophios exposes is meant to be one a person can read and edit. §6.1 now says that, which is a change in stated guidance rather than in behaviour.


tests/core/test_manufactured_documents.py
```diff
@@ -0,0 +1,383 @@
...
+    # A declared workflow output, so the step's `out:` is built from a bound
+    # port rather than left empty: `Step._yml` spells it `[{name: value}]`, and
+    # an edge written there in the wrong spelling is the likeliest way this
+    # front end emits something the grammar refuses.
+    workflow = Workflow([touch, append], 'manufactured_py')
+    workflow.outputs.result = append.outputs.file
+    workflow.compile()
+    # Both spellings: inlined subtrees are passthrough to the parser, so the
+    # uninlined form is what puts a subworkflow step body in front of it.
+    workflow.to_wic_yaml()
```
darkclad, line 340: **🟡 This call doesn't reach the uninlined form.**

`to_wic_yaml()` defaults to `inline_subworkflows=True`, so this emits the same inlined form `compile()` already produced. The comment above says the point is the uninlined form, which would be `to_wic_yaml(inline_subworkflows=False)`. Even with that, `workflow` has no nested `Workflow`, so there is no subworkflow step body to put in front of the parser. Nesting one (e.g. wrapping `append` in its own `Workflow`) and passing the flag would make the comment true.

tests/core/test_manufactured_documents.py
```diff
@@ -0,0 +1,383 @@
...
+
+
+#: Documents chosen to drive the manufacturing sites rather than to be
+#: interesting themselves: a linear chain, a scattered step, an explicit edge,
+#: and a subworkflow — the shapes that make the compiler build new documents.
```
darkclad: **🟡 Two of the four promised shapes aren't here.**

The comment lists a linear chain, a scattered step, an explicit edge and a subworkflow. `_DRIVERS` has a linear chain, an explicit edge and a `wic:` sidecar entry. No driver scatters, and none has a subworkflow step, so the sites that build documents for those shapes are only reached if something else drives them. Either add the two documents or trim the comment to what's listed.

tests/core/test_manufactured_documents.py
```diff
@@ -0,0 +1,383 @@
...
+
+**BLIND SPOT, from that same fact.** The desugared spelling has no tags, so the
+unknown-tag rule cannot fire here. A manufactured `{'wic_not_a_thing': 1}` is
+indistinguishable from passthrough CWL, which §1 says is open by definition, and
+the parser accepts it — verified by mutation, which this test does not catch.
+What it does catch is structure: a step in the removed single-key form, a step
+with no `id`, a malformed `wic:` sidecar. Those are the defects this class has
+actually produced. A misspelled desugared key would need the vocabulary check
+`Grammar.SIDECAR_KEYS` already owns for `wic:`, extended to input position, and
+that is a language change rather than a test.
```
darkclad: **🟡 The second commit fixed this blind spot, so the docstring is now stale.**

It says a manufactured `{'wic_not_a_thing': 1}` passes and that closing it "is a language change rather than a test". `wic024` is that language change, and in input position the parser now rejects that document. Worth rewriting the paragraph to say what remains open: `wic_`-prefixed keys outside construct position, which stay passthrough by design.

tests/core/test_manufactured_documents.py
```diff
@@ -0,0 +1,383 @@
...
+    {'steps': [{'id': 'mk_file', 'in': {'name': {'wic_inline_input': 'a'}},
+                'out': [{'file': {'wic_anchor': 'produced'}}]},
+               {'id': 'sink', 'in': {'file': {'wic_alias': 'produced'}}}]},
+    {'wic': {'steps': {'(1, mk_file)': {'wic': {}}}},
+     'steps': [{'id': 'mk_file', 'in': {'name': {'wic_inline_input': 'z'}}}]},
+]
+
+
+@pytest.mark.fast
+@pytest.mark.slow
```
darkclad: **🟡 Marked both `fast` and `slow`.**

No other test in `tests/` carries both. `pytest -m fast` would pick up 100 full hermetic compiles with no deadline, and `pytest -m "not slow"` would skip it. Given the cost, `slow` alone looks like the intent.

src/sophios/lang/parser.py
```diff
@@ -551,12 +561,47 @@ def _desugared_form(
...
+    and reserved the prefix in *name* position too — `in: {wic_: !* x}`, a port
+    someone may legitimately call `wic_`, which the generators found at once.
+    Passthrough stays open: `wic`, `wicked` and `my_wic_key` are ordinary CWL.
+    """
+    if not key.startswith(CONSTRUCT_PREFIX):
+        return
+    diags.error(
+        Code.RESERVED_KEY,
+        f"{key!r} is not a Sophios construct, and a single-key mapping beginning 'wic_' is read as "
+        f'one. The constructs are: {", ".join(sorted(Key.ALL))}',
```
darkclad: **🟡 The suggestion list includes a key that is also an error here.**

`wic024` only fires in input position, but `Key.ALL` includes `wic_anchor`, which in that position is `wic019`. Someone who reads the list and fixes `wic_anchr` to `wic_anchor` trades one error for the next. `Forms.DESUGARED_KEYS` is exactly the set that is valid here, and it is already derived so it can't drift: `", ".join(sorted(Forms.DESUGARED_KEYS))`. `Forms` is defined below this function, but it's only read at call time, so that's fine.

src/sophios/lang/parser.py
```diff
@@ -164,7 +173,8 @@ def parse(text: str, filename: str = '<string>') -> ParseResult:
     if root is None:  # An empty document is well-formed and carries nothing.
         return ParseResult(Document(span=whole), diagnostics)
 
-    _report_unknown_tags(root, filename, diagnostics)
+    for node in _every_node(root):
+        _report_unknown_tag(node, filename, diagnostics)
```
darkclad, line 177: **🟡 Nit: a comment still names the old function.**

This splits `_report_unknown_tags` into `_every_node` and `_report_unknown_tag`, but the comment in `_input_value` (line 484, outside the diff) still says "`_report_unknown_tags` covers the whole graph before this runs". It should name these two.

darkclad (commented):
Review at `797307a`. The new `wic024` rule is correct: it fires only on a single-key mapping in input position, and a port named `wic_` is still accepted in name position. There are no correctness findings. There are six small ones, all inline: four in the new test module (two comments that promise coverage the test doesn't have, a docstring the second commit made stale, and a double `fast`/`slow` marker) and two in `parser.py`.

Out of scope, not a finding: the compiler never calls `sophios.lang.parse`, so `wic_inline_inpt` still compiles silently from the CLI. The description already says nothing depends on this grammar yet.

tests/core/test_manufactured_documents.py
vjaganat90: Fixed in `03666ff`. I removed the inaccurate inlined/uninlined claim; this call is now described narrowly as exercising the direct Python API serialization entry point alongside the compile path.

tests/core/test_manufactured_documents.py
vjaganat90: Fixed in `03666ff`. I trimmed the inventory comment to the three shapes the drivers actually contain: a linear chain, an explicit edge, and a `wic:` sidecar.

tests/core/test_manufactured_documents.py
vjaganat90: Fixed in `03666ff`. The module docstring now says that single-key `wic_` mappings in input position are caught, and identifies only `wic_`-prefixed keys outside construct position as the remaining intentional passthrough blind spot.

tests/core/test_manufactured_documents.py
vjaganat90: Fixed in `03666ff`. Removed `pytest.mark.fast`; the generated-input property is now marked `slow` only.

src/sophios/lang/parser.py
vjaganat90: Fixed in `03666ff`. The `wic024` diagnostic now derives its suggestions from `Forms.DESUGARED_KEYS`, so it lists only constructs valid in input position and no longer suggests `wic_anchor`.

src/sophios/lang/parser.py
vjaganat90: Fixed in `03666ff`. Updated the stale comment to name `_every_node` and `_report_unknown_tag`.

tests/core/test_manufactured_documents.py
vjaganat90: Follow-up in `bccd03d`: agreed that “remaining blind spot” overstated the situation. I verified the construct positions directly: input inline/anchor typos produce `wic024`, a correctly spelled input anchor produces `wic019`, and malformed output bindings produce `wic005`. The paragraph now says those positions are explicitly closed and distinguishes ordinary passthrough elsewhere as intentional, not a missed construct.

tests/core/test_manufactured_documents.py
darkclad: **Fixed** The comment now says only what the call does, so it no longer promises uninlined-form coverage the test never had. Trimming the claim fits better than adding a nested workflow just to back it.

tests/core/test_manufactured_documents.py
darkclad: **Fixed** The comment now lists the three shapes `_DRIVERS` actually contains, so it no longer claims coverage the list doesn't have.

tests/core/test_manufactured_documents.py
darkclad: **Fixed** At `bccd03d` I parsed each case the new paragraph describes, and every one behaves as stated: a misspelled input key gives `wic024`, `wic_anchor` in input position gives `wic019`, and any other `out:` value gives `wic005`. A port named `wic_` and a multi-key mapping still pass through, which matches the passthrough sentence.

tests/core/test_manufactured_documents.py
darkclad: **Fixed** The property test is now `slow` only, and `pytest -m slow` on this file runs it and passes at `bccd03d`.

src/sophios/lang/parser.py
darkclad: **Fixed** The `wic024` message now suggests `wic_alias, wic_inline_input, wic_raw_cwl`, so it no longer points anyone at `wic_anchor`. Because the list comes from `Forms.DESUGARED_KEYS`, it stays in step with the table.

src/sophios/lang/parser.py
darkclad: **Fixed** The comment in `_input_value` now names `_every_node` and `_report_unknown_tag`, the functions that actually run.

darkclad (approved):
LGTM at `bccd03d`.

All six comments from my review at `797307a` are addressed in `03666ff` and `bccd03d`. The code change is that `wic024` now suggests only keys that are valid in input position. I ran `test_lang_parser.py` and `test_manufactured_documents.py` at this head, including the slow property test: 159 passed. I also parsed the cases in the new "Construct coverage" docstring, and each gives the diagnostic it describes.

PR #419, opened 2026-09-16
Consolidate: register ids and review narration leave the source, and a scan keeps them out

vjaganat90:
The source describes the code, not the process that produced it. A reader has the source; they do not have the trackers, the pull requests, or the review history.

## The standing decision this restores

The TDD tracker records, dated **2026-08-27**:

> **P-numbers are tracker-internal (maintainer decision).** They no longer appear in the code: this register is offline, so an in-code "P14" is an opaque token to any future reader. […] The repository now contains no register tokens; this tracker is their only home.

It was applied and then eroded to **138 sites across 33 files** — property and counterexample register ids, task and spec numbers, pull-request numbers, and sentences narrating how a defect came to light.

It eroded because it was a note in an offline document and nothing checked it. So it is now a test.

## `tests/core/test_prose_boundary.py`

| Rule | Was | Now | What a green run proves |
|---|---:|---:|---|
| no tracker identifiers | 136 in 27 files | 0 | **The whole claim.** An id is mechanical, and every comment, docstring and string literal is read. |
| none of the known narration phrasings | 32 in 11 files | 0 | **Only that** — see below. |

Each rule ships with samples proving it fires, and samples proving it leaves real prose alone — a scan never seen to fail is a scan whose green means nothing.

### The narration rule is a blacklist, and says so

There is no regex for "this sentence is about how the code came to be". English has unbounded ways to write it, and presenting a phrase list as a semantic boundary makes a green run mean more than was checked — which is how this PR twice claimed a clean tree that was not clean. The test is named `test_no_known_narration_phrasing_reaches_the_source`, and its docstring states what it does not cover. It catches the forms this repository keeps writing; judgement covers the rest, at review.

### The scan reads every string, not the ones that looked like prose

Collecting only lines that *begin* with `#` misses a comment after code. Reading docstrings and assert messages misses a `rationale=` field printed when a property fails — which carried a pull-request number and a property id that nothing saw. Singling out the string shapes that happen to be prose today is how the next shape is missed, so **every string constant is read**; across the whole tree that costs one fix. What it cannot see, and does not claim to, is text assembled at runtime: an id interpolated from a variable is invisible to any static scan.

Four vocabularies were narrower than what they name. A task id was matched as `T2.4` but not as "Task 5" or "Tasks 3-7", which is how most of them are written; a property id matched at two digits but not one; `round \d` does not match "round three"; and `CR-\d+` was absent entirely, so `CR-104` sat in `compiler.py` and `test_predicates.py` while the scan reported a clean tree.

A `design_docs/` path also exempted its entire line, so any id beside the path rode along — the exemption exists for a spec number, which beside the document names a section the reader can open, and now covers that alone.

**Narration is matched per logical block, not per physical line.** A comment wraps at the margin, so "Found by / mutation" spans two comment tokens and matches neither half; reading lines lets any phrasing escape by being long enough. A run of consecutive comments is one block and a docstring is one block. Blocks are collected *separately* from the line scan rather than replacing it, because the `design_docs/` exemption belongs to one line and joining would let a path on one line excuse an id on another.

**One exemption.** This module, which cannot state the rule without spelling an example of what the rule rejects. `diagnostics.py` was exempt too, on the grounds that it defines the `wic0NN` codes — but those were never in the pattern, so the exemption bought nothing and would have hidden an ordinary tracker id in the one file nobody was checking. It is gone.

## Every fact survived; only its provenance went

> *Before:* "It held exactly one entry: CE-13's `!&` bound to a step input, which the compiler rejected with wic011 … `semrefac_7.1` settled that as a grammar defect … mypy is what reported the expiry."
>
> *After:* "Its one entry was an `!&` edge definition bound to a step *input*; that is now a type error, because `EdgeDef` is not a member of the `InputValue` union, so the construct cannot be built here to be filtered out."

Same constraint, no archaeology, and it explains the present design rather than the discovery.

## Four module docstrings were narratives

`ast_strategies.py` 64 → 12 · `equivalence.py` 79 → 12 · `test_canonical_emission.py` 52 → 13 · `test_manufactured_documents.py` 40 → 13.

Nothing load-bearing was deleted. The registers buried in them **moved to where a reader meets them**:

- `equivalence.py`'s `CANNOT DETECT` register → onto `equivalent()`, the public entry point a caller actually reads
- its `WILL NOT READ` rule → onto `_steps_of`, the reader it governs
- `ast_strategies.py`'s `CANNOT GENERATE` and the well-formed-is-not-well-typed caveat → onto `documents()` and `compilable_documents()`

## Verification

`mypy`, `autopep8` and `pylint` all clean at the lane's own invocation. No behaviour change: this PR touches prose, one new test module, and no production code path.

Each rule ships samples proving it fires and samples proving it leaves real prose alone, and each widening is mutation-verified — revert it and a test fails. Two of those mutations initially passed, because the samples I wrote overlapped other alternatives in the pattern rather than isolating their own; they are isolated now.

## The module-docstring line limit is gone

It was a numeric gate standing in for a semantic judgement. Line count cannot tell a local contract from a historical essay, and it was satisfiable by moving text into a function docstring — which is exactly what happened, twice, and both times it caused a defect:

- `_steps_of` ended up stating the same rule twice, with its summary line stranded in the middle.
- `test_manufactured_documents.py` lost the paragraph explaining why dumping a manufactured document and re-parsing it is a *fair oracle* rather than a round-trip through a lossy form, and why a manufactured inline-input or anchoring mistake cannot pass silently.

That paragraph is restored, onto `_parses`, which is what it governs. The four docstring rewrites the limit prompted were improvements on their own merits and stay; what goes is the standing gate.

## The generator's construct gap is derived, not described

`documents()` said it covered "the whole language" while declaring `!cwl` undrawable, and the coverage test claimed every construct the language has while quantifying over an inventory that omits it. The design requires every AST construct kind to appear in the bounded sample, so the gap is now checked rather than written down: the kinds come from the `InputValue` union, and each must appear in `CONSTRUCTS` or in `NOT_GENERATED` with a reason. Adding an AST node fails that test until someone either draws it or records why they cannot.


tests/core/test_prose_boundary.py
```diff
@@ -0,0 +1,191 @@
...
+        path (Path): The module to read.
+
+    Returns:
+        list[tuple[int, str]]: Line number and text, for prose only. Code is
+            excluded so that a string literal under test is not mistaken for
+            a docstring.
+    """
+    lines: list[tuple[int, str]] = []
+    for offset, text in enumerate(path.read_text(encoding='utf-8').splitlines(), start=1):
+        if text.lstrip().startswith('#'):
```
darkclad: Only whole-line `#` comments and docstrings are collected, so a comment after code and an assert message are never scanned. These pass today: `# CE-06` at `test_lang_parser.py:960`, `P02/P09` at `test_manufactured_documents.py:42`, `P29` in the assert message at `test_equivalences.py:285`, `#383's P4` at `transformations.py:226`. Tokenizing with `tokenize` and taking every `COMMENT` token would catch the trailing ones.

tests/core/test_prose_boundary.py
```diff
@@ -0,0 +1,191 @@
...
+from .source_scan import REPO_ROOT, package_files, parsed
+
+#: Tracker identifiers. Two deliberate absences: `wic0NN`, because a diagnostic
+#: code is public contract, matchable by a caller and documented; and a spec
+#: number cited alongside `design_docs/`, because that names a section of a
+#: document the reader has. A bare spec number names nothing they can open.
+TRACKER_TOKENS: Final = re.compile(
+    r'\bP\d{2}[a-c]?\b'      # property register ids
+    r'|\bCE-\d+\b'           # counterexample register ids
+    r'|\bT\d\.\d+\b'         # task ids
```
darkclad: `T\d\.\d+` matches only the `T2.4` form. The tree still has 17 spelled-out ones ("Task 5", "Tasks 3-7", "Tasks 2-8") across `ast_strategies.py`, `equivalence.py`, `test_generators.py`, `test_hermeticity.py`, `test_canonical_emission.py`, `test_equivalence.py` and `test_predicates.py`; `git grep -nE "Tasks? [0-9]"` lists them.

tests/core/test_prose_boundary.py
```diff
@@ -0,0 +1,191 @@
...
+    r'|\bSpec \d\b'          # specification numbers
+    r'|(?<![\w/])#\d{3,}\b'  # pull request numbers
+)
+
+#: Narration of how the code came to be. The rule is that a reason survives and
+#: its provenance does not, so these read as evidence that a rewrite was skipped.
+PROCESS_NARRATION: Final = re.compile(
+    r'an earlier version|the first draft|review (?:found|caught|added)'
```
darkclad: The tree's own wording is "an earlier draft", which this pattern misses. It is still at `ast_strategies.py:265`, `equivalence.py:222` and `:313`, and `test_equivalence.py:543`.

tests/core/test_prose_boundary.py
```diff
@@ -0,0 +1,191 @@
...
+
+    Args:
+        text (str): One line of docstring or comment.
+
+    Returns:
+        bool: True for a register, task, spec or pull-request id. A spec number
+            beside a `design_docs/` path is a section of a document in this
+            repository, so it is a reference and not bookkeeping.
+    """
+    return 'design_docs/' not in text and TRACKER_TOKENS.search(text) is not None
```
darkclad: The `design_docs/` exemption skips the whole line, so a CE, P, task or PR id sitting next to a `design_docs/` path also passes. The docstring says the exemption is for spec numbers only; dropping just the `Spec \d` alternative on those lines would match it.

darkclad (commented):
The scan passes at e6347f0 while tracker ids and narration it is meant to reject are still in the tree, so the "0 remaining" claim does not hold yet. Four gaps are inline on `test_prose_boundary.py`: trailing comments and assert messages are not scanned, spelled-out "Task N" ids and "an earlier draft" are not matched, and a `design_docs/` path exempts every id on its line. The other comments are prose the rewrite left untrue or broken.

tests/core/test_generators.py
```diff
@@ -72,12 +72,12 @@ def _collect(document: Document) -> None:
...
-    going red — `documents()` keeps producing the whole language, so P26 stays
-    green no matter what the filter removes. Found by mutation: an exclusion
-    predicate of `lambda d: True` empties the strategy entirely and every test
-    in this file passed.
+    `compilable_documents()` is what the compile-driving properties quantify
+    over, and its filter is the one place a construct can leave the oracle's
+    reach without anything going red — `documents()` keeps producing the whole
+    language, so coverage stays green no matter what the filter removes. An
+    exclusion predicate of `lambda d: True` empties the strategy entirely, and
+    every test in this file still passes.
```
darkclad: "every test in this file still passes" reads as the current state, but the test below exists to fail when the exclusion empties the strategy. This needs to say that without the check they would have passed.

tests/core/test_canonical_emission.py
```diff
@@ -1,55 +1,15 @@
...
-  3. `test_one_workflow_compiles_identically_under_four_hash_seeds` — one
-     hand-built workflow, compiled in four fresh interpreters: the literal
-     claim the property register used to make, kept as a separate example
-     because the property above deliberately does not assert it end-to-end
-     (see its own CANNOT DETECT paragraph).
+`set` iteration is hash-seeded, and three sites in the compile path once let
+that order out: `maybe_add_requirements` set the emitted `requirements:` key
+order (eight seeds, eight orders), `add_yamldict_keyval_out` set a step's
+`out:` list order (six seeds, three orders), and the speculative-insertion list
+in the compiler does the same behind a default-off flag.
```
darkclad: "does the same" is no longer true: `compiler.py:1165` is `insertions = sorted(set(insertions))`.

tests/core/test_canonical_emission.py
```diff
@@ -228,7 +188,7 @@ def _site_of(line: int, spans: list[tuple[int, int, str]]) -> str:
 def _set_iteration_violations(tree: ast.AST) -> list[tuple[int, str]]:
     """Every place a `set(...)` call is iterated into an ordered structure.
 
-    Total rule, not a shape catalogue — the lesson from #384's review round
+    Total rule, not a shape catalogue
```
darkclad: Removing the #384 reference left a fragment: "Total rule, not a shape catalogue / three, where a scan…". "the PR had just fixed" still points at a PR the text no longer names.

tests/core/equivalence.py
```diff
@@ -245,7 +205,16 @@ def _stem(step_id: str) -> str:
...
+    admits that form and Sophios emits neither it nor an array-form `in:`;
+    treating an unhandled shape as "nothing here" made a non-list `steps:`
+    become `[]`, so two documents differing only in their steps produced two
+    empty graphs and compared *equal*. The one verdict a comparison oracle must
+    never invent is "no difference", so a shape it cannot read has to be loud.
+
+    The document's steps, in the one surface form the compiler emits.
 
     A list, every entry a mapping carrying `id`. Anything else raises, per the
     module docstring's WILL NOT READ: CWL admits `steps:` as a mapping keyed by
```
darkclad: The module docstring no longer has a WILL NOT READ section, so this pointer is dangling. "which an earlier draft did" two lines down is also narration the new scan is meant to catch. The rewritten module docstring also dropped that the guard applies only to UP_TO_RENAMING.

tests/core/test_equivalences.py
```diff
@@ -164,12 +164,12 @@ def _hits_the_scalar_coercion_gap(document: Yaml) -> bool:
...
     argument's declared type, with a `LITERAL_TYPE_MISMATCH` diagnostic.
     `ast_strategies.py` documents why such documents are still drawn — they
-    are well-formed, and unlike CE-13 this is not a single AST-shape predicate
+    are well-formed, and this is not a single AST-shape predicate
     — but a *refused* compilation is no basis for an equivalence. Left
     unfiltered here, every draw that hits it would abort *both* compiles this
     property needs before `equivalent()` is ever called, smothering the
     property this task is actually responsible for. Excluded the same way
-    `compilable_documents()` excludes CE-13: narrowly, by the exact predicate
+    `compilable_documents()` excludes an edge definition in input position: narrowly, by the exact predicate
```
darkclad: `compilable_documents()` excludes nothing today: `NOT_YET_COMPILABLE` is `{}` (`ast_strategies.py:352`).

tests/core/wic_corpus.py
```diff
@@ -22,7 +22,7 @@
 #: In-repo fallback when no config exists: a pure read, so collecting the
 #: suite on a fresh machine provisions nothing (`pytest --collect-only` used
 #: to write ~/wic/global_config.json and copy adapters as a side effect,
-#: a finding of the #383 review).
+#: which a round-trip property found).
```
darkclad: "which a round-trip property found" isn't true. A round-trip property compares documents and cannot observe files written to `~/wic` during collection. If the provenance must go, drop the clause instead of replacing it.

src/sophios/lang/parser.py
```diff
@@ -612,7 +612,7 @@ class Forms:  # pylint: disable=too-few-public-methods  # a namespace, not a typ
 
     Kept side by side and in one namespace because the language's claim is that
     the two spellings are equivalent (§6.1). Splitting them into separate
-    module globals is how they drift apart — which is exactly the defect CE-02
+    module globals is how they drift apart — which is exactly the defect this
```
darkclad: "the defect this recorded": with the id gone, "this" has nothing to refer to. Name the defect (the two tables drifting apart) directly.

src/sophios/lang/parser.py
```diff
@@ -840,7 +840,7 @@ def _opaque(node: yaml.nodes.Node, file: str, diags: Diagnostics,
     if _is_edge_def(node):
         # Both spellings, in every position `_opaque` walks. Catching only the
         # tagged one would let `{wic_anchor: n}` through exactly where `!& n`
-        # is refused, which is the two-surfaces-one-language divergence CE-02
+        # is refused, which is the two-surfaces-one-language divergence this
```
darkclad: Same as above: "the … divergence this was about" lost its referent.

tests/core/strategies.py
```diff
@@ -14,7 +14,7 @@
 identifiers = st.text('abcdefghijklmnopqrstuvwxyz_', min_size=1, max_size=8)
 #: Scalar payloads as YAML source text, spelled to include exactly what a
 #: [a-z_]-only alphabet can never produce: type-ambiguous quoted strings,
-#: exponent floats and specials, null, dates. The #383 review traced four of
+#: exponent floats and specials, null, dates. Four round-trip defects traced to
```
darkclad, line 17: "traced to its findings to the old alphabet's blind spot" doesn't parse, most likely left over from the edit.

src/sophios/lang/compatibility.py
```diff
@@ -2,7 +2,7 @@
 
 This module answers one deliberately narrow question: may Sophios reject a
 user-authored reference before handing the emitted document to CWL?  It does
-not normalize declarations into the typed IR reserved for Spec 3, and it does
+not normalize declarations into the typed IR that replaces it later, and it does
```
darkclad: "the typed IR that replaces it later" now reads as this module being replaced. The original meant the IR replaces the legacy handling. Consider "the typed IR planned to replace the legacy path".

tests/core/test_prose_boundary.py
vjaganat90: Comments are tokenized now, so one after code is seen, and assert messages are collected as a third surface — a message is prose a reader meets at the moment the test fails, so an id there is as unopenable as one in a comment. All four sites you named are gone.

tests/core/test_prose_boundary.py
vjaganat90: Added the spelled-out form. Seventeen sites; nearly all were "Tasks 3-7" standing for the compile-driving properties, which the surrounding prose already names — the id was redundant with its own explanation.

tests/core/test_prose_boundary.py
vjaganat90: Added to the narration pattern. The four sites are restated as constraints rather than as what a draft used to do.

tests/core/test_prose_boundary.py
vjaganat90: Narrowed to the spec number alone, which is what the docstring always claimed. A test now pins that a property or task id sitting beside the path is still caught.

tests/core/test_generators.py
vjaganat90: Reworded: without the check below, every test in this file would still pass.

tests/core/test_canonical_emission.py
vjaganat90: Corrected — that site deduplicates through a set rather than iterating out of one.

tests/core/test_canonical_emission.py
vjaganat90: Reduced to the rule, with the two shapes a catalogue misses kept as examples. Nothing points at a PR any more.

tests/core/equivalence.py
vjaganat90: Both pointers name `_steps_of` now. I also folded the duplicated explanation back into one docstring with the summary at the top, and restored the scope you flagged: `_steps_of` feeds `_dataflow`, whose labelled graph is the part of a document UP_TO_RENAMING compares.

tests/core/test_equivalences.py
vjaganat90: Dropped — the sentence compared this exclusion to one that is currently empty, so it described nothing. It now states the narrowing rule directly.

tests/core/wic_corpus.py
vjaganat90: Dropped the clause and kept what matters: the fallback is a pure read, and reading the user config instead is what makes collection provision the environment it is meant to inspect.

src/sophios/lang/parser.py
vjaganat90: Names the defect directly now — the two tables drifting apart.

src/sophios/lang/parser.py
vjaganat90: Same: one language, two surfaces, diverging.

tests/core/strategies.py
vjaganat90: Fixed, and "its" now has an antecedent.

src/sophios/lang/compatibility.py
vjaganat90: Taken as suggested: the typed IR planned to replace the legacy path.

tests/core/test_prose_boundary.py
darkclad: At 4a622a4 three spots from this rewrite still need a fix. `test_generators.py:231` still says "the earlier draft satisfied that…" and `test_lang_parser.py:1113` still says "(the reviewer's method, made permanent)"; the pattern only matches "an earlier" and `review` followed by a verb, so neither is caught, and `(?:an|the) earlier` would close the first. Also the rewrite at `test_equivalence.py:542-543` repeats a word: "…in any compiled document. Those / Those sources name no producer". Please fix these three.

darkclad (approved):
LGTM

PR #420, opened 2026-09-16
Consolidate: the Python API tests say what they pin, and are named for what they cover

vjaganat90:
**Stacked on #419.** Two files held the Python API's entire test surface — 1105 and 369 lines — with **fifty-eight tests carrying no docstring**. Six modules in the suite had undocumented tests, sixty-six in all; these two held all but eight of them. A test whose sole statement of intent is its own name tells a reader what it touches and never why.

## Sixty-five docstrings, written from what each test asserts

Not from the names. Where a reason is knowable it is given:

- `test_scatter_rejects_unbound_foreign_or_scalar_inputs` — three refusals, "because each would otherwise produce a document that compiles and then fails at runtime for a different reason"
- `test_falsey_inline_values_are_preserved` — "an empty string is a value, not an absent one. A truthiness test anywhere on this path would drop it silently"
- `test_workflow_requires_steps_in_constructor` — why it asserts on `__dict__` rather than behaviour: "an inherited or dynamically added method would still be callable"
- `test_output_target_directories_compile_as_runner_local_inputs` — why the input is typed `string` and not `Directory`: the directory does not exist until the step runs

## Renamed for the surface they cover

`test_python_api.py` → **`test_python_api_workflow.py`**
`test_tool_builder.py` → **`test_python_api_tool_builder.py`**

Twelve files carried the old paths: six workflow files (`lint_and_test`, `lint_and_test_macos`, `run_workflows`, `run_workflows_weekly`, `fuzzy_compile_weekly`, `benchmarks_weekly`), three dev-doc pages, and three test modules including a cross-module import. `test_ci_coverage`'s `_paths_of` **raises** rather than skips on a lane path that does not resolve, which is what proves they were all found.

`pylint` now reports no `missing-function-docstring` anywhere in the repository.

## A real defect this surfaced, already on `master`

The full fast suite failed on `test_tag_decisions_agree_across_positions`:

```
'{wic_: x}': input=['wic024'], passthrough=[]
```

`wic024` — the misspelled-construct rule from #417 — fires in **construct position only**, by design: a name or a passthrough key may carry the `wic_` prefix. That property compared *every* diagnostic across the two positions with `wic024` subtracted, which makes each future position-specific code another subtraction. It compares the unknown-tag verdict directly now, which is what its name always claimed.

It is intermittent, it is on `master`, and nothing on this branch caused it. Per the TDD guide, an intermittent property failure is a bug report rather than a flake, so it is fixed here: the comparison excludes `RESERVED_KEY`, with the asymmetry stated as the reason. Run repeatedly to confirm, and the rest of the property is untouched.

## Deliberately not done

A stricter "every public callable is documented" rule was written, run, and **removed**. It flags 97 public callables in `src/` — 38 of them `tool_builder.py` builder methods like `.label(x)`. `pyproject.toml` disables `missing-class-docstring` with a stated reason ("mandatory docstrings add noise before they add clarity"), so imposing the stricter rule is a policy change, not a cleanup, and not mine to make in passing.

## Verification

`mypy`, `autopep8` and `pylint` clean at the lane's own invocation. No production code changed.

## The rename had a cost the census could not see

`pytest tests/core/test_tool_builder.py -k test_tool_builder` selected all fourteen tests, because `-k` matches the module's name as well as each test's own — every test in the file inherited the match from the file it was in. Renaming to `test_python_api_tool_builder.py` took that away from the three whose own names do not repeat the subject, so the step quietly went to eleven. Three workflows carried the line; the filter selects nothing the bare path does not, so it goes.

`_keyword_admits` in `test_ci_coverage.py` modelled `-k` against the test's name alone, which is the same divergence one level up: the guard could not have caught this, because it did not know the mechanism. It now matches what pytest matches. No invocation in the tree changes verdict today — but a future `-k` naming a module would have been read as selecting nothing, and reported tests that do run as collected by no lane.


.github/workflows/run_workflows.yml
```diff
@@ -179,11 +179,11 @@ jobs:
 
     - name: PyTest Tool Builder
       if: always()
-      run: cd sophios/ && pytest tests/core/test_tool_builder.py -k test_tool_builder
+      run: cd sophios/ && pytest tests/core/test_python_api_tool_builder.py -k test_tool_builder
```
darkclad: The rename drops 3 tests from this step. `-k` also matches the file name, so `test_tool_builder.py` matched `-k test_tool_builder` and all 14 tests ran. `test_python_api_tool_builder.py` does not match, so these now get deselected:

- `test_old_tool_builder_module_name_is_not_available`
- `test_command_line_tool_constructor_hides_internal_fields`
- `test_structured_port_references_do_not_accept_raw_strings`

`test_ci_coverage.py` doesn't catch this because it ignores `-k`. Every test in this file belongs to the tool builder, so dropping `-k` restores the old behavior. The same line is in `fuzzy_compile_weekly.yml:115` and `run_workflows_weekly.yml:130`.

.github/workflows/fuzzy_compile_weekly.yml
```diff
@@ -108,11 +108,11 @@ jobs:
         # WIC Python API workflows as well as the WIC Python API itself.
       - name: Validate Sophios Python API Workflows (*.py -> *.wic)
         if: always()
-        run: cd sophios/ && pytest tests/core/test_python_api.py -k test_validate_generated_python_workflows
+        run: cd sophios/ && pytest tests/core/test_python_api_workflow.py -k test_validate_generated_python_workflows
 
       - name: PyTest Tool Builder
         if: always()
-        run: cd sophios/ && pytest tests/core/test_tool_builder.py -k test_tool_builder
+        run: cd sophios/ && pytest tests/core/test_python_api_tool_builder.py -k test_tool_builder
```
darkclad: Same `-k test_tool_builder` issue as in `run_workflows.yml:182`: the renamed file no longer matches, so 3 tests are deselected.

.github/workflows/run_workflows_weekly.yml
```diff
@@ -123,11 +123,11 @@ jobs:
       # Sophios Python API workflows as well as the WIC Python API itself.
     - name: Validate Sophios Python API Workflows (*.py -> *.wic)
       if: always()
-      run: cd sophios/ && pytest tests/core/test_python_api.py -k test_validate_generated_python_workflows
+      run: cd sophios/ && pytest tests/core/test_python_api_workflow.py -k test_validate_generated_python_workflows
 
     - name: PyTest Tool Builder
       if: always()
-      run: cd sophios/ && pytest tests/core/test_tool_builder.py -k test_tool_builder
+      run: cd sophios/ && pytest tests/core/test_python_api_tool_builder.py -k test_tool_builder
```
darkclad: Same `-k test_tool_builder` issue as in `run_workflows.yml:182`: the renamed file no longer matches, so 3 tests are deselected.

tests/core/test_prose_boundary.py
```diff
@@ -0,0 +1,191 @@
...
+
+    Returns:
+        list[tuple[int, str]]: Line number and text, for prose only. Code is
+            excluded so that a string literal under test is not mistaken for
+            a docstring.
+    """
+    lines: list[tuple[int, str]] = []
+    for offset, text in enumerate(path.read_text(encoding='utf-8').splitlines(), start=1):
+        if text.lstrip().startswith('#'):
+            lines.append((offset, text))
```
darkclad: This only counts a line as a comment when the whole line starts with `#`, so comments after code are never scanned. Three tracker ids already get through at this head:

- `tests/core/test_lang_parser.py:960`: `# CE-06: name tag on a collection`
- `tests/core/test_manufactured_documents.py:42`: `# ... pinned by P02/P09`
- `tests/core/test_manufactured_documents.py:44`: `# outside the core zone (Spec 0)`

Collecting `tokenize.COMMENT` tokens instead would catch both whole-line and trailing comments without mistaking a `#` inside a string for one.

darkclad (commented):
Two findings inline. Most of the PR is docstring and comment rewording; with docstrings ignored, only `tests/core/test_lang_parser.py` and the new `tests/core/test_prose_boundary.py` change code, and the new prose checks pass at this head.

.github/workflows/run_workflows.yml
vjaganat90: Confirmed — the filter selected nothing the bare path does not, because every test in the file is a tool-builder test, so it goes. Same in the two weekly workflows.

.github/workflows/fuzzy_compile_weekly.yml
vjaganat90: Dropped here too.

.github/workflows/run_workflows_weekly.yml
vjaganat90: Dropped here too.

tests/core/test_prose_boundary.py
vjaganat90: Tokenized, so both forms are seen. I also fixed the reason the census could not have caught your other finding: `_keyword_admits` matched the test name alone, while pytest matches the module name too — which is exactly how the old file name satisfied the `-k` for every test inside it.

.github/workflows/run_workflows.yml
darkclad: Verified at 9f2eafc: the step runs the whole file, and `_keyword_admits` now matches the module name the way pytest does, with the rename pinned in the parametrized cases. Good catch on the census being the reason this slipped through.

.github/workflows/fuzzy_compile_weekly.yml
darkclad: Verified at 9f2eafc.

.github/workflows/run_workflows_weekly.yml
darkclad: Verified at 9f2eafc.

tests/core/test_prose_boundary.py
darkclad: Verified at 9f2eafc: comments are tokenized and every string constant is scanned, and the three ids I listed are gone. `test_prose_boundary.py` and `test_ci_coverage.py` pass locally (50 tests), and moving the probe to `tmp_path` also settles the one thing I was going to raise about the new test.

darkclad (approved):
LGTM

PR #421, opened 2026-09-16
Consolidate: one stub builder instead of four, and the encodings nine files left to the locale

vjaganat90:
**Stacked on #420.** The last of the three.

## One stub builder instead of four

`_clt` was written four times — `synthetic_tools`, `benchmarks`, `test_inference`, `test_reference_compatibility` — as the same dict with two variations: one added the EDAM namespace, two applied canonical normal form.

Now one exported `clt(inputs, outputs, *, javascript=False, canonical=False)`. `benchmarks.py` calls it directly; the two modules that need canonical normal form bind that expectation, which is the only binding that says something a call site does not:

```python
#: `clt` in canonical normal form — the shape the inference and
#: explicit-edge paths read.
_clt = partial(clt, canonical=True)
```

**No call site moved.** That matters: the call sites in `test_inference` are multi-line dict literals, and three earlier attempts to rewrite them with a regex produced unbalanced parentheses and an insertion landing inside a string. The local definitions were removed by AST line-span and the name rebound, which touches one line per file.

## The mechanical backlog

- three unused imports — two of them left behind by the consolidation above
- an empty-sequence comparison spelled as `== []`
- **nine call sites across three files opening text with no encoding.** Not style: the default follows the locale, so the same test reads different bytes on a different machine. Same class as the Windows failure in #416.

## What I did *not* do, and why

**The twelve `unreachable` blocks stay.** My survey listed them as the highest-value dead code. Reading three — `main.py:201`, `benchmarks.py:96`, `test_setup.py:32` — shows plainly reachable code whose very next line uses the value assigned. Pylint mis-attributes flow around `sys.exit`, which it treats as `NoReturn`.

Deleting code on a linter's word, without reading it, would have been the defect rather than the fix. The item is withdrawn from the plan.

**The `compiler.py` family is untouched** — `compiler.py`, `inference.py`, `ast.py`, `inlineing.py`, `utils_cwl.py`, `cwl_subinterpreter.py`. The IR migration deletes or replaces them, so consolidating them is work thrown away twice.

## Verification

`mypy` clean · `autopep8` clean · `pylint` **9.80** (9.73 before this chain) · 1058 fast tests pass, `tests/contrib` included. Exit codes captured directly rather than read off a filtered tail.

## The namespace contract, made explicit

Folding four builders into one had to choose a `$namespaces` behaviour, and the first choice was silent: only `synthetic_tools.py`'s builder had ever emitted `edam`, so three fixture families gained a key they never had. Invisible at run time — the compiler adds the namespace itself and format matching does not read it — which is exactly why it needed deciding rather than inheriting.

`clt` now carries the prefixes a tool's own formats name, and nothing when none do. That restores the fixtures declaring no format, keeps the ones that declare one, and stops a builder shared by every module being specific to one ontology. Five tests pin the shapes: absent, string format, list format, unprefixed format, and raw against canonical against JavaScript.


tests/core/provocations.py
```diff
@@ -70,6 +70,7 @@ def _provoke_subworkflow_invalid() -> None:
 
     class _RefusesEverything:  # pylint: disable=too-few-public-methods
         def validate(self, _tree: object) -> None:
+            """Accept any tree: this stands in for a validator the caller must supply."""
```
darkclad: **🟡 The docstring says the opposite of the method.**

"Accept any tree", but `validate` always raises `ValueError`, so it refuses every tree. The class name `_RefusesEverything` and the comment at line 77 both say so. Something like "Refuse any tree: a stand-in for a validator the caller must supply."

tests/core/equivalence.py
```diff
@@ -245,7 +205,16 @@ def _stem(step_id: str) -> str:
...
+    """WILL NOT READ what the compiler does not emit.
+
+    Raises `TypeError` on a mapping-form `steps:` rather than coercing it. CWL
+    admits that form and Sophios emits neither it nor an array-form `in:`;
+    treating an unhandled shape as "nothing here" made a non-list `steps:`
+    become `[]`, so two documents differing only in their steps produced two
+    empty graphs and compared *equal*. The one verdict a comparison oracle must
+    never invent is "no difference", so a shape it cannot read has to be loud.
+
+    The document's steps, in the one surface form the compiler emits.
```
darkclad: **🟡 WILL NOT READ moved here, but two pointers still send readers to the module docstring.**

Lines 220 and 250 (`_in_of`) both say "per the module docstring's WILL NOT READ", and the module docstring no longer has that section. They should point at `_steps_of`.

The move also left this docstring saying the same thing twice. Lines 212–215 and 220–223 both explain that coercing to `[]` made two documents compare equal. The one-line summary ("The document's steps, in the one surface form the compiler emits.") now sits in the middle instead of at the top.

tests/core/test_inference.py
```diff
@@ -17,20 +19,15 @@
 from sophios.utils import step_name_str
 from sophios.utils_cwl import desugar_into_canonical_normal_form
 from sophios.utils_graphs import get_graph_reps
+from .synthetic_tools import clt
```
darkclad, line 21: **🟡 Unused imports left behind, and the new import is out of order.**

With the local `clt` gone, `CWL_VERSION` (line 18), `desugar_into_canonical_normal_form` (line 20) and `Cwl` (line 23) have no users in this file. Also, `from .synthetic_tools import clt` sits between two `sophios` imports, and `from functools import partial` is in its own block after `typing`. The CI pylint step runs over `tests/`, so it will report unused-import and wrong-import-order here. The same ordering problem is at lines 8 and 22 of `test_reference_compatibility.py`, where every import is still used.

tests/core/test_canonical_emission.py
```diff
@@ -228,7 +188,7 @@ def _site_of(line: int, spans: list[tuple[int, int, str]]) -> str:
 def _set_iteration_violations(tree: ast.AST) -> list[tuple[int, str]]:
     """Every place a `set(...)` call is iterated into an ordered structure.
 
-    Total rule, not a shape catalogue — the lesson from #384's review round
+    Total rule, not a shape catalogue
```
darkclad: **🟡 This sentence broke when the history was cut.**

"Total rule, not a shape catalogue" has no punctuation, and the next line starts mid-thought ("three, where a scan that enumerated the shapes it knew about missed the two the PR had just fixed"). "The PR" and "three" now refer to nothing. Either restore enough context or reduce it to the rule itself: walk every node and match the offending node's own shape.

tests/core/strategies.py
```diff
@@ -14,7 +14,7 @@
 identifiers = st.text('abcdefghijklmnopqrstuvwxyz_', min_size=1, max_size=8)
 #: Scalar payloads as YAML source text, spelled to include exactly what a
 #: [a-z_]-only alphabet can never produce: type-ambiguous quoted strings,
-#: exponent floats and specials, null, dates. The #383 review traced four of
+#: exponent floats and specials, null, dates. Four round-trip defects traced to
```
darkclad, line 17: **🟡 Garbled after the edit.**

"Four round-trip defects traced to its findings to the old alphabet's blind spot" has one "to" too many, and "its" has no antecedent. Probably "Four round-trip defects traced to the old alphabet's blind spot; these are its complement."

tests/core/test_equivalence.py
```diff
@@ -513,9 +513,9 @@ def test_isomorphism_alone_would_not_be_enough() -> None:
     documents have the same shape — two nodes, one edge — and a different tool
     at every node, so a bare matcher calls them equivalent. That is precisely
     the kind of thing an IR migration could get wrong, which is why the
-    relation Spec 3 imports must not be the one already in the tree.
+    relation a migration imports must not be the one already in the tree.
 
-    Asserts the *verdict*, not the path string. An earlier version asserted
+    Asserts the *verdict*, not the path string. Asserting
```
darkclad, line 518: **🟡 Sentence fragment.**

"Asserting `'tool multiset' in found.path`, which is this project's recurring failure…" has no main verb. It reads as though "Not" or "It used to assert" was cut. Something like "It used to assert `'tool multiset' in found.path`, which…" would fix it.

tests/core/wic_corpus.py
```diff
@@ -22,7 +22,7 @@
 #: In-repo fallback when no config exists: a pure read, so collecting the
 #: suite on a fresh machine provisions nothing (`pytest --collect-only` used
 #: to write ~/wic/global_config.json and copy adapters as a side effect,
-#: a finding of the #383 review).
+#: which a round-trip property found).
```
darkclad: **🟡 A round-trip property can't have found this.**

Writing `~/wic/global_config.json` during `pytest --collect-only` is a side effect of collection, and a round-trip property only compares values. It's also still an account of how the bug was found, which is the kind of text this PR removes elsewhere, and `test_prose_boundary.py` doesn't catch it. Dropping the clause leaves the part that matters: the fallback is a pure read, so collecting on a fresh machine provisions nothing.

darkclad (commented):
Review at `b7999be`. There are no runtime findings, and the new `test_prose_boundary.py` passes locally (17 tests). The rename left nothing pointing at `test_python_api.py` or `test_tool_builder.py`. The shared `clt` adds an `edam` `$namespaces` entry the three local copies lacked, but the compiler adds it anyway and format matching doesn't read it. Skipping `wic024` in the input-vs-passthrough comparison matches the parser, which only reports it in input position.

The seven inline comments are all about prose or imports: a docstring that says the opposite of its method, pointers to a docstring section that moved, unused imports left by the `clt` extraction, three sentences broken by the history cleanup, and one comment whose claim can't be true.

tests/core/provocations.py
vjaganat90: Taken as worded: "Refuse any tree: a stand-in for the validator the caller must supply."

tests/core/equivalence.py
vjaganat90: Both pointers name `_steps_of` now, the duplicate explanation is folded into one docstring with the summary back at the top, and the UP_TO_RENAMING scope is stated where the guard is read.

tests/core/test_inference.py
vjaganat90: All three unused imports removed and both files reordered.

tests/core/test_canonical_emission.py
vjaganat90: Reduced to the rule itself, keeping the two shapes a catalogue misses as examples.

tests/core/strategies.py
vjaganat90: Fixed close to your reading.

tests/core/test_equivalence.py
vjaganat90: Rewritten as a conditional, so the sentence has a verb and still says what the alternative would cost.

tests/core/wic_corpus.py
vjaganat90: Clause dropped. What is left is the part that matters — the fallback is a pure read — plus what reading the user config instead would do.

tests/core/test_equivalence.py
```diff
@@ -540,8 +540,8 @@ def test_the_dag_check_sees_a_port_rewired_onto_a_different_workflow_input() ->
     Every argument a step does not receive from another step is emitted as
     `in: {arg: {source: {step_id}___{arg}}}` against a declared workflow-level
     input — the majority of the bindings in any compiled document. Those
-    sources name no producer, so an earlier draft skipped them entirely, and
-    with them skipped the two documents below were UP_TO_RENAMING-equal: both
+    Those sources name no producer, and skipping them on that ground makes the
```
darkclad: **🟡 "Those" is doubled.**

The rewording in `e1922a1` starts this line with "Those", but line 542 already ends with it, so the docstring reads "…any compiled document. Those Those sources name no producer". Dropping one of them fixes it. Everything else from the first round checks out at `af3dfdf`.

darkclad (commented):
Review at `af3dfdf`. The new commit `4a622a4` holds up: all 27 tests in `test_prose_boundary.py` pass, no module docstring is over 25 lines, and the restored paragraph on `_parses`, the removed `diagnostics.py` exemption and the probe file in `tmp_path` all check out. There are two small comments, both on the narration regex. The doubled "Those" at `test_equivalence.py:543` is still open.

tests/core/test_prose_boundary.py
```diff
@@ -0,0 +1,272 @@
...
+#: recurring forms and nothing else, and the test that applies it says so --
+#: a green run means these phrasings are absent, not that no narration is.
+#:
+#: Case-sensitive in one place on purpose: `this PR` is narration, `this
+#: property` is not, and a case-insensitive `this pr` matches both.
+PROCESS_NARRATION: Final = re.compile(
+    r'an earlier (?:version|draft)|the first draft'
+    r'|[Rr]eview\b[^.]{0,25}?\b(?:found|caught|added|proved|showed)'
+    r'|found by mutation|round (?:\d+|one|two|three|four|five)\b'
+    r'|as it stood before|semrefac|this PR\b|this pull request'
```
darkclad: **🟡 Case-sensitive everywhere, not in one place.**

The comment says the pattern is case-sensitive only for `this PR`, but `re.IGNORECASE` was removed from the whole pattern. A phrase at the start of a sentence now slips through. "An earlier version…", "The first draft…", "Found by mutation…", "Round 3…", "As it stood before…" and "This pull request…" all matched before this commit and don't now. Nothing in the tree uses those forms today, so the scan stays green.

Keeping the flag and scoping the exception matches the comment: `re.IGNORECASE` on the pattern, with `(?-i:[Tt]his PR)\b` in place of `this PR\b`. I checked that this matches "An earlier version", "Round 3" and "This PR", and leaves "this property" alone.

tests/core/test_prose_boundary.py
```diff
@@ -0,0 +1,272 @@
...
+#: review", and no regex closes that. It is a blacklist that catches the
+#: recurring forms and nothing else, and the test that applies it says so --
+#: a green run means these phrasings are absent, not that no narration is.
+#:
+#: Case-sensitive in one place on purpose: `this PR` is narration, `this
+#: property` is not, and a case-insensitive `this pr` matches both.
+PROCESS_NARRATION: Final = re.compile(
+    r'an earlier (?:version|draft)|the first draft'
+    r'|[Rr]eview\b[^.]{0,25}?\b(?:found|caught|added|proved|showed)'
+    r'|found by mutation|round (?:\d+|one|two|three|four|five)\b'
```
darkclad: **🟡 `round` has no leading word boundary.**

`round (?:\d+|one|…)\b` matches inside "around", so "wraps around 2 lines" is flagged as narration. That was already true of `round \d`, but the scan now reads every string constant, so a false match is more likely. `\bround` fixes it, and "wraps around 2 lines" would fit in `test_the_patterns_leave_real_prose_alone`.

tests/core/test_equivalence.py
vjaganat90: Fixed — one "Those" left. That was mine, introduced when I reworded the sentence in the previous round.

tests/core/test_prose_boundary.py
vjaganat90: Taken, and you were right that it was live: the exception was written lowercase-only, so `This PR` at the start of a sentence — where it is usually written — still escaped. It is `[Tt]his PR` now, inside an otherwise case-insensitive pattern, with both casings as samples.

tests/core/test_prose_boundary.py
vjaganat90: Fixed with `\bround`. Broader than reported once the scan reads every string constant: `a workaround one` and `background 3` matched too. All three are now samples in the leave-real-prose-alone set.

darkclad (approved):
LGTM

PR #422, opened 2026-09-16
diagnostics: an undefined edge and a duplicate definition carry codes

vjaganat90:
Two failures that already existed, given codes; and the rule they enforce, documented.

## The change

```
ValueError: Error! No definition found for !&prod_ligand.gro!
ValueError: Error! Multiple definitions of &dup!
```

Both raised from library code. The design says the library reports and only the CLI adapter exits — a caller could not match on these, suppress them, or tell them from a bug. Nine such sites were converted with the diagnostics work; these two were missed.

- **`wic025`** — a reference nothing defines
- **`wic026`** — a name defined twice

Each message now says what to do about it: define the edge with `!&` on the producing step's `out:`, or declare it in the document's `inputs:` and bind it by name.

**No behaviour change in the compiler.** Both cases already failed; they now fail with a code. One thing downstream did change and had to be fixed: `wic026` was `ValueError: Error! Multiple definitions of &`, which the fuzz job's message arm accepted. Giving it a code moved it to the structured arm, where it was not in the tolerated set — turning a draw the job had always tolerated into a failure, and leaving the message entry unreachable so nothing pointed at the gap.

## The reference states the rule

New §4.1.2. `!* name` names an edge `!& name` defines. The definition must come **before** the reference — `!* e` above its `!& e` is `wic025` even in the same document, which the tutorial has always implied and the reference did not say — and there must be only one.

It also documents what `inputs:` is for, since that is the alternative to reaching for an undefined edge: a bare name asks the *includer* for a value, `!*` asks the *compilation* for an edge.

No `lang_version` bump — the grammar is unreleased, so this is a defect in it rather than a change to it.

## Why these paths had no tests

`compile_harness` and `hermetic` both pass `testing=True`, which lets `test_cwl_embedding_independence` recompile every subworkflow as if it were root. That branch absorbs an undefined edge into a workflow input, so the production error path was unreachable from the suite. The tests here compile with `testing=False` and are the first to exercise it.

## One test is load-bearing

`test_a_definition_nothing_consumes_is_not_reported` pins the claim these rules deliberately do **not** make. An `!&` with no `!*` is how a workflow names an artifact it produces, and fifteen corpus documents rely on it — "definitions and references must pair up" is the obvious next tightening and would reject all fifteen.

## Verification

Mutation-verified three ways: reverting either code to a different one fails two tests, disabling the unbound check fails three. `mypy` clean, `autopep8` clean, `pylint` 9.80, 1063 fast tests pass.

## What `wic025` does *not* cover, said plainly

The first draft of §4.1.2 claimed the rule holds over a compilation and that "the check runs at the root". It does not. `wic025` is raised by the root *invocation*: a reference inside an included subworkflow is absorbed into an input of that subworkflow one level down, its edge name is not carried up, and the root has nothing left to check.

A root containing a child that references an edge nothing defines **compiles**, and the reference surfaces as a generated workflow input that nothing produces. That is now what the reference says, and `test_an_obligation_no_includer_discharges_is_not_reported` pins it so the hole is visible instead of documented as closed.

Closing it means carrying the unresolved name up and checking the aggregate once at the root — deferred-obligation discharge in the typed IR. A third special case in the per-step loop that phase replaces is not the place for it.

## Tests that build a real parent

The previous subworkflow test compiled a child with `is_root=False`, which proves deferral is *allowed* and nothing about the obligation being *discharged*. Three cases replace it: a parent definition satisfying a child (asserting the child binds to the parent's output, not to a generated input); the ordering rule; and the escape above.

`compile_production` joins `hermetic.py` as the one production-mode helper, so `testing=False` is stated once rather than copied with fourteen arguments.


tests/core/hermetic.py
```diff
@@ -48,6 +48,26 @@ def compile_hermetic(yml: Yaml, name: str = 'oracle', *,
...
 
+def compile_production(yml: Yaml, name: str = 'binding', *,
+                       tools: Tools | None = None,
+                       is_root: bool = True) -> CompilerInfo:
+    """Compile as a real root compilation does: `testing=False`.
+
+    The one harness that does. `compile_hermetic` passes `testing=True`, which
+    is what lets `test_cwl_embedding_independence` recompile each subworkflow
+    as though it were root -- and that same branch absorbs an undefined edge
+    into a workflow input, so the production diagnostics are unreachable
```
darkclad: This is right for `wic025` and not for `wic026`. The `wic025` raise sits behind `if is_root and not testing:`, but the `wic026` raise has no `testing` guard at all — I compiled the duplicate-anchor document through both harnesses and got `['wic026']` either way.

So `compile_production` is required for `wic025` only; `compile_hermetic` reaches `wic026` just as well. Worth narrowing the wording to name `wic025`, so the next reader does not infer a guard that is not there. Same wording appears in `provocations.py`.

tests/core/provocations.py
```diff
@@ -40,6 +40,43 @@
 COMPILED: Final[dict[Code, Callable[[], object]]] = {}
 
 
+def _compile_as_root(yml: dict) -> None:
+    """Compile as a real root would: `testing=False`.
+
+    `compile_harness` and `hermetic` both pass `testing=True`, which is what
+    lets `test_cwl_embedding_independence` recompile every subworkflow as if it
+    were root. That branch absorbs an undefined edge into a workflow input, so
+    the production path these two codes live on is reachable no other way.
```
darkclad: Same as the note on `hermetic.py:59`: "reachable no other way" holds for `wic025` but not for `wic026`, which raises under `testing=True` too.

Separately — `_compile_as_root` is `hermetic.compile_production` with a different default name. Same `default_compilation_settings()`, same `testing=False` and `is_root=True`, same synthetic registry; I could not find a behavioural difference between them. Since `hermetic` is already imported across the suite, one of the two could probably go.

tests/core/test_fuzzy_compile.py
```diff
@@ -4,6 +4,8 @@
 import graphviz
 from hypothesis import given, settings, HealthCheck
 import networkx as nx
+from typing import Final
```
darkclad: `pylint` reports `C0411` here — the stdlib import `typing.Final` placed after the third-party block. It does not fail the gate (this PR scores 9.80 against `fail-under=9.25`), so it is purely cosmetic; moving the import up beside `import unittest` clears it.

docs/sophios_language_reference.md
```diff
@@ -282,6 +282,66 @@ guess at intent much later.
...
+
+Order is part of the rule. A reference is resolved against the definitions
+seen so far, so `!* e` written above the `!& e` that defines it is `wic025`
+even though the definition is in the same document.
+
+A document included as a subworkflow may reference an edge its includer has
+already defined; the includer's definitions are in scope when the child is
+compiled.
+
+**Not yet enforced.** A reference that *no* enclosing document defines is
```
darkclad, line 308: This is the second "specified but not fully wired" gap in the document, and the Implementation-status table at the top still lists only `!cwl` while its own prose says "Two parts of it". The mismatch predates this PR, but since this paragraph adds a third such caveat, a row recording that `wic025` covers the root document rather than the whole compilation would bring the table back in line with the prose that introduces it.

darkclad (approved):
LGTM — approving.

I built a Python 3.13 environment and ran the suite against this head (tree `a10a4bf`). `tests/core`: **1154 passed, 15 skipped, 28 failed** — and those 28 are byte-identical on master, all `wic015` from a stopped Docker daemon, as are 5 more in `tests/contrib/test_rest_api.py` from cwltool's `pwd` import on Windows. No regressions. mypy clean over `src/ examples/ tests/`, autopep8 clean, pylint 9.80 against the 9.25 threshold (master scores 9.72).

Two things I checked rather than assumed:

- `reference_relation(None, 'File')` returns `UNKNOWN`, not `DISJOINT`, so the `wic025` raise is genuinely reached rather than pre-empted by `wic023`.
- Deleting `Code.DUPLICATE_EDGE_DEF` from `TOLERATED_CODES` does fail `test_the_tolerated_set_covers_what_the_message_arm_used_to`, so that guard catches the drift it was written for.

`test_an_obligation_no_includer_discharges_is_not_reported` is the part I would single out. Pinning the hole the diagnostic does *not* close, with an assertion written to flip when the typed IR closes it, is the right way to ship a partial enforcement — the prose in the language reference and the test say the same thing, so neither can quietly drift from the code.

Follow-ups left inline. None of them blocks this.

tests/core/hermetic.py
vjaganat90: Fixed in 8a24192. The harness docstring now names only `wic025` as unreachable under `testing=True` and explicitly notes that other diagnostics may be reachable through either harness.

tests/core/provocations.py
vjaganat90: Fixed comprehensively in 8a24192. I removed `_compile_as_root` entirely: the `wic025` provocation now reuses `hermetic.compile_production`, while the `wic026` provocation deliberately uses `compile_hermetic` to preserve the distinction you verified.

tests/core/test_fuzzy_compile.py
vjaganat90: Fixed in 8a24192 by moving `Final` into the stdlib import group. Targeted pylint is clean for import ordering.

docs/sophios_language_reference.md
vjaganat90: Fixed in 8a24192. The implementation-status table now has an explicit `wic025` row recording root-only enforcement and the pending Spec 3 Link work; the prose below the table also points to §4.1.2.

darkclad (approved):
LGTM — approving `8a24192`.

All four follow-ups are addressed, and I checked each against the code rather than the commit message:

- `hermetic.py` — the docstring now scopes the claim to `wic025`, which is the one actually behind `if is_root and not testing:`.
- `provocations.py` — `_compile_as_root` is gone, and routing the `wic025` provocation through `compile_production` while deliberately leaving `wic026` on `compile_hermetic` encodes the distinction instead of just describing it. That reads better than what I suggested.
- `test_fuzzy_compile.py` — `C0411` is clear; targeted pylint is 10.00/10 across the three touched files.
- the language reference — the `wic025` row also makes the table agree with the "Two parts of it" prose above it, which had been out of step since before this PR.

Verification on this head: `tests/core` **1154 passed / 15 skipped / 28 failed**, the 28 identical to master and all `wic015` from a stopped Docker daemon on my machine. mypy clean over `src/ examples/ tests/`, autopep8 clean, pylint 9.80 against `fail-under=9.25`.

I also went looking for the same gap on the *other* code, since `wic025` is not in `TOLERATED_CODES`. It holds up: `'Error! No definition found for !&'` was never in the old `expected_messages`, and that raise already sat behind `is_root and not testing` before this PR, so it cannot reach the fuzz lane and there is nothing there to tolerate. The `wic026` half was the only one that needed moving between the arms, which is what this does.

One observation I do not attribute to this PR: `TestFuzzyCompile::test_fuzzy_compile` failed once in three full-suite runs here, then passed six times including four in isolation. `@given(wic_strategy)` runs without `derandomize`, so every run explores different data — and this commit only moves an import in that file, so the intermittency is the lane's own. Worth knowing if the weekly fuzz job has shown the same thing.

PR #423, opened 2026-09-16
Spec 3: the typed IR, and lowering a document onto it

vjaganat90:
Adds `sophios.ir`: the types a compiled workflow is made of, and a total `lower` from a parsed document onto them. Nothing imports it yet and the compiler is unchanged.

Stacked on #422.

## The types

Frozen, slotted dataclasses composed into a `WorkflowGraph`. Each names something the compiler already manipulates and currently spells as a `str` or threads through its call stack: `Namespace`, `StepId`, `Direction`, `PortId`, `PortType`, `Port`, `Binding`, `Edge`, `DeferredObligation`.

Two of them are worth singling out:

- **A step occurrence is not the tool it runs.** `StepId` is namespace, occurrence index and name. `docs/tutorials/append_twice.wic` invokes `append` twice and sequence-form `steps:` exists so it can — an IR keyed on the authored id rejects a shipped tutorial. The index is what the compiler already means by `{stem}__step__{i}__{key}`.
- **A port's direction is part of its identity.** CWL puts inputs and outputs in separate namespaces and tools declare the same name on both; without it the two compare equal, an edge between them reads as a self-loop, and a set of port ids loses one of every pair.
- **`PortId` is not a step id.** Today both are `str`, so nothing stops one being passed where the other is meant. Once they are different types the checker can stop it.
- **`PortType` is where the leak boundary becomes a type.** The declared CWL stays verbatim in `declared`, because passthrough is open and whatever was given must be emitted. Beside it sit the parts Sophios actually interprets — optionality, array depth, union alternatives. A type the compiler does not interpret has `declared` and nothing else.

The four mappings the compiler threads through its recursion (`explicit_edge_defs`, `explicit_edge_calls`, `input_mapping`, `output_mapping`) become fields of the graph, because that is what they always were: state every function had to be handed and could quietly disagree about. They are pairs, not dicts — a frozen graph holding a mutable mapping can have its invariants invalidated after the constructor checked them.

## Invariants at construction

They live in `__post_init__`, so a malformed graph cannot be *constructed* rather than merely detected. An edge naming a port no step declares would otherwise survive to emission as a dangling `source:` — which CWL accepts and a runner then fails on. A checker a caller must remember to run is a checker that is eventually not run.

A repeated step name is the one case handled twice on purpose: `WorkflowGraph` refuses it, but by then the only thing to point at is the whole graph, so lowering locates it first and the diagnostic carries the second step's position.

## Lowering

Total in the sense `parse` is — every document either lowers or produces diagnostics, and nothing raises. It reads the AST and nothing else: no registry, no filesystem, no config. A port's type is therefore what the document declared, and inference has not run; resolving a step's tool against the environment stays `Resolve`'s job.

An input bound with `!*` becomes an `Edge` when this document defines the name and a `DeferredObligation` when it does not — which is exactly what a subworkflow's reference to a parent's edge is. Discharging one belongs to `Link`, which can see the enclosing graph.

## Also here

`tests/core/test_settings_boundary.py` asserted that no `argparse.Namespace` travels past the CLI, but looked for the word `Namespace` in an annotation. That was already one near-miss from wrong — the comment records dodging sophios' own `Namespaces` by word-bounding — and the IR's `Namespace` collides with it head-on. It now resolves the module's imports and matches the spellings that actually denote the type, which also closes the `from argparse import Namespace as Ns` hole.

## Every authored binding is kept

Lowering only edges makes two documents differing solely in a literal produce equal graphs, so `Emit` would have to read the AST again — and a graph that cannot reproduce its own document is not the document's meaning. Each `in:` entry becomes a `Binding` carrying what was written *and* what it resolved to; `edges` and `obligations` are views over those, so there is one source of truth rather than three lists that can disagree.

## Edge resolution respects source order

Pre-scanning every definition made `!* e` written above its `!& e` resolve — which the compiler refuses and §4.1.2 (#422) forbids. Resolution now runs against the definitions seen so far: earlier definition is an edge, never defined here is a `DeferredObligation`, defined only later is `wic025`. A name defined twice used to keep the first and drop the second, leaving nothing for any phase to report; it is `wic026`.

## What the properties now rule out

Each of these was satisfiable by an implementation that is obviously wrong:

| Was | Now |
|---|---|
| totality — "a graph **or** diagnostics" | a second property requires every well-formed document to actually lower |
| determinism over three fields | the whole graph compared |
| seven hand-written malformed constructions | generated hostile constructions |
| edges checked against declared ports | every reference the graph holds — bindings, obligations, all four mappings |

Ten mutations verify them: revert an invariant or a resolution rule and a test fails.

`Namespace` no longer rejects a part containing `___`. That narrowed the language to suit the emitted spelling while leaving the other identity components free to contain it — both a restriction and incomplete.

Also closes the `*args`/`**kwargs` hole in the settings-boundary scan, which admitted `def f(*args: argparse.Namespace)`.


src/sophios/ir/lower.py
```diff
@@ -0,0 +1,160 @@
...
+    identities = _step_identities(document, here, diagnostics)
+    if identities is None:
+        return Lowered(None, diagnostics)
+
+    defined_anywhere = _edge_definitions(identities, document, diagnostics)
+    defined_so_far: dict[str, PortId] = {}
+    nodes: list[StepNode] = []
+
+    for step_id, step in zip(identities, document.steps, strict=True):
+        inputs = tuple(Port(PortId(step_id, Direction.INPUT, name), UNDECLARED, span=step.span)
```
darkclad, line 79: **Lowering still raises on five documents `parse` accepts.** The commit message says a step recovered without an id used to hit `StepId`'s constructor and is now reported — that part is fixed. But `PortId` (`types.py:104`) and `DeferredObligation` (`types.py:177`) carry the same non-empty-name guard, and both are reached from here without a catch:

```
empty input name:       *** RAISED ValueError: a port must have a name
empty output name:      *** RAISED ValueError: a port must have a name
empty out edge-def key: *** RAISED ValueError: a port must have a name
empty in + empty out:   *** RAISED ValueError: a port must have a name
empty edge-ref name:    *** RAISED ValueError: a deferred obligation must name what it awaits
```

Each from a document the parser accepts, e.g.:

```yaml
steps:
- id: s
  in:
    "": !* e
```

```yaml
steps:
- id: s
  in:
    f: !* ""
```

`_step_identities` is the shape that works: check the name, report it, and return. The same treatment over input names, `out:` entry names and edge-ref names would close this — and since all three are "a name the document left empty", they may deserve one diagnostic between them rather than three.

src/sophios/ir/lower.py
```diff
@@ -0,0 +1,160 @@
...
+
+    A repeated `id` is not an error: sequence-form `steps:` exists so that a
+    tool can be invoked twice, and `docs/tutorials/append_twice.wic` does. What
+    cannot be lowered is a step the parser recovered without a name, which is
+    reported rather than raised.
+    """
+    identities: list[StepId] = []
+    for index, step in enumerate(document.steps, start=1):
+        if not step.id:
+            if step.span is not None:
```
darkclad: When a step has no id *and* no span, this returns `Lowered(None, Diagnostics())` — neither a graph nor a diagnostic, which is precisely what the module docstring rules out:

```
>>> lower(Document(steps=(Step(id=''),)))
graph=False diags=0 errors=False ok=False
```

`ok` is False and nothing says why, which is worse for a caller than the exception this replaced. It would fail `test_every_parseable_document_lowers_or_diagnoses` directly if the generator could reach it.

Worth noting the guard may be dead: I could not get the parser to emit a spanless step, including through recovery — `- ` , `- foo`, `- [1,2]`, `- in: {}`, `- id:` and `- id: 3` all come back with a span. If a recovered step always has one, the condition can go and the `return` can be unconditional on the diagnostic. If it can ever be absent, this needs a spanless diagnostic rather than none.

src/sophios/ir/types.py
```diff
@@ -0,0 +1,324 @@
...
+    """
+
+    sink: PortId
+    name: str
+    type: PortType
+    span: SourceSpan | None = None
+
+    def __post_init__(self) -> None:
+        """Reject an obligation with nothing to satisfy."""
+        if not self.name:
```
darkclad, line 179: An empty edge name is fatal in reference position and fine in definition position:

```
empty edge-ref name (`f: !* ""`):  *** RAISED ValueError
empty edge-def name (`- f: !& ""`):  ok graph=True errors=False
```

The second puts `''` into `explicit_edge_defs` as a live producer name, where it will match any later `!* ""` that manages not to raise first. Whichever way this resolves, the two positions should agree — both a diagnostic, or both accepted.

src/sophios/ir/types.py
```diff
@@ -0,0 +1,324 @@
...
+    def __post_init__(self) -> None:
+        """Reject a graph naming a port no step declares, anywhere.
+
+        Checked over every reference the graph holds -- edges, obligations and
+        all four mappings -- rather than over the edges alone. An identity that
+        resolves to nothing is the same defect wherever it is stored, and it
+        reaches emission as a dangling `source:`, which CWL accepts and a runner
+        then fails on.
+        """
+        occurrences = [step.id for step in self.steps]
```
darkclad, line 280: The occurrence check is right, but nothing here relates a step's namespace to the graph's, so a graph will hold a step that sits somewhere else entirely:

```
>>> WorkflowGraph(Namespace(('a',)), steps=(StepNode(StepId(Namespace(('b',)), 1, 'mk')),))
accepted; graph.namespace == 'a'
```

Nothing depends on it yet because `StepId` carries its own namespace and `_references()` compares whole identities. Raised only because it is the one reference relationship in this class that is not checked, and `Link` composing graphs across namespaces is where that would start to matter.

src/sophios/ir/types.py
```diff
@@ -0,0 +1,324 @@
+"""The types a compiled workflow is made of.
+
+Each names something the compiler already manipulates and spells as a string or
+threads through a call stack, so the checker can verify what it means rather
+than only what it computes.
+
+Frozen and slotted, as `sophios.lang.nodes` is, and holding no mutable
+container: an invariant checked in `__post_init__` is worth having only if it
+cannot be invalidated afterwards.
```
darkclad: "holding no mutable container" is slightly wider than what holds. The four mappings are tuples now, which is the change that matters — but `passthrough` and `interpreted` are tuples of pairs whose *values* are `OpaqueCwl`, and that admits `list` and `dict`:

```
>>> hash(WorkflowGraph(Namespace(('a',)), passthrough=(('k', {'a': 1}),)))
TypeError: unhashable type: 'dict'
```

No invariant reads those payloads, so nothing can be invalidated through them and the reasoning in the docstring is untouched — it is the blanket phrasing that overshoots. "holds no mutable container any invariant depends on" would be exactly true, and would also explain why an `OpaqueCwl` payload is not a counterexample.

tests/core/test_lower.py
```diff
@@ -0,0 +1,275 @@
...
+    """Parse and lower one document, for the example-based claims below."""
+    document = parse(source, 'probe.wic').document
+    assert document is not None
+    return lower(document)
+
+
+@pytest.mark.fast
+@given(strat.documents())
+@COVERAGE
+def test_every_parseable_document_lowers_or_diagnoses(document: Document) -> None:
```
darkclad, line 50: The second property is a real strengthening — totality satisfiable by rejecting everything was a genuine gap. Worth recording what this pair still cannot see, since it is where the remaining raises live: `documents()` draws step ids from `STEMS`, binding names from `inputs_of(stem)` and output names from `outputs_of(stem)`, so **no name it draws is ever empty** — and every generated node carries `_SPAN`, so a spanless step is unreachable too.

That is the right generator for "a workflow the compiler can resolve", and it should stay that way. The empty-name cases belong beside the example-based claims below, as parametrized sources — the same place `test_a_malformed_graph_cannot_be_constructed` already covers the types.

darkclad (commented):
Review at `cea2e3b`, against #422 (`8c254a4`) rather than `master` — stacked, so this PR's own change is 5 files.

Installed and run: `pip install -e ".[test,mypy-types]"` on Python 3.12. `test_lower.py` + `test_settings_boundary.py` are **176 passed**; the fast lane is **1131 passed**; `mypy src/` is clean over 71 files; `pylint src/ examples/**/*.py tests/` is **9.79/10** against `fail-under=9.25` with nothing in `src/sophios/ir/`; `autopep8 --diff` is clean. The one fast-lane failure and all 36 non-fast failures reproduce identically on `master` — `cwltool` imports `pwd`, and the `cwltool`/`docker` binaries are absent on this box. No regressions.

`Direction` in the identity, `StepId` carrying the occurrence, `Binding` keeping what was written, and `_references()` checking all four mappings rather than the edges alone are each the right call, and the last one generalises the invariant instead of adding a case to it. Repeated step ids being *allowed* is the fix that matters most, since `append_twice.wic` ships.

**One thing blocks the totality claim.** `lower.py` says "nothing here raises", and `StepId`'s constructor is now reported rather than hit — but `PortId` and `DeferredObligation` carry the same kind of guard and are not. Five documents `parse` accepts still raise `ValueError`, all of them an empty name in some position. Details inline; the repro is four lines.

The rest are small: a `return None` that produces neither a graph nor a diagnostic, an asymmetry between `!& ""` and `!* ""`, and two claims that are slightly wider than what the code enforces.

src/sophios/ir/lower.py
vjaganat90: Confirmed all five, and the cause is worse than the symptom: I fixed the guard that was reported and not the other two that carry it. `PortId` and `DeferredObligation` are now checked before construction, in `_every_name_is_present`, which reports `wic027` for an input name, an `out:` entry, an edge definition and an edge reference. One code, as you suggested — they are one mistake in four positions.

src/sophios/ir/lower.py
vjaganat90: You are right that the guard is dead; I could not produce a spanless step through the parser either. It is gone and the diagnostic is unconditional. `Diagnostics.error` takes an optional span now, so a document assembled in memory still reports instead of returning a result that says nothing.

src/sophios/ir/types.py
vjaganat90: Both report `wic027` now. The definition side was the worse half — it put `""` into the edge table as a live producer, exactly as you describe.

src/sophios/ir/types.py
vjaganat90: Added. It was the one reference relationship in the class left unchecked, and `Link` composing graphs across namespaces is where it would have surfaced.

src/sophios/ir/types.py
vjaganat90: Reworded to "no mutable container that any invariant depends on", with the `OpaqueCwl` case named so the exception is explicit rather than an apparent oversight.

tests/core/test_lower.py
vjaganat90: Agreed — the generator should keep drawing resolvable workflows, so the empty-name cases are parametrized examples beside the construction table rather than a widened strategy. Worth recording that your point generalises: the property was green because nothing it drew could reach the branch, which is the same reason the totality claim passed while five documents raised.

darkclad (approved):
LGTM.

Re-reviewed at `3da5508`. Every one of the six comments is addressed, and `_every_name_is_present` is the right shape for it — reporting the class rather than the one position that happened to be reported already is what the previous fix was missing. One code across all four positions is the right call too.

Verified rather than read:

```
empty input name         graph=False errors=True  ['wic027']
empty output name        graph=False errors=True  ['wic027']
empty out edge-def key   graph=False errors=True  ['wic027']
empty edge-def name      graph=False errors=True  ['wic027']
empty edge-ref name      graph=False errors=True  ['wic027']
empty in + empty out     graph=False errors=True  ['wic027']
empty step id            graph=False errors=True  ['wic007']
step is null             graph=False errors=True  ['wic007']
step is scalar           graph=False errors=True  ['wic007']
no id key                graph=False errors=True  ['wic007']
self ref same step       graph=True  errors=True  ['wic025']
repeated step id         graph=True  errors=False []
numeric input name       graph=True  errors=False []
null input value         graph=True  errors=False []
dup input name           graph=True  errors=False []
dup output name          graph=True  errors=False []
empty steps              graph=True  errors=False []

RAISING CASES: 0
```

Seventeen documents `parse` returns, none of them raising. `!& ""` and `!* ""` agree now, the spanless branch that returned neither a graph nor a diagnostic is gone, and a graph refuses a step that sits in another namespace:

```
ValueError: StepId(namespace=Namespace(parts=('b',)), index=1, name='mk')
            sits in ('b',), not this graph's ('a',)
```

**Suite.** `pip install -e ".[test,mypy-types]"` on Python 3.12. `test_lower.py` + `test_settings_boundary.py`: **182 passed**. Fast lane: **1138 passed, 1 skipped**. `mypy src/`: clean over 71 files. `pylint src/ examples/**/*.py tests/`: **9.79/10** against `fail-under=9.25`, nothing in `src/sophios/ir/`. `autopep8 --diff`: clean.

The one fast-lane failure and all 36 non-fast failures reproduce identically on `master` and are this box, not this branch: `cwltool` imports `pwd`, and the `cwltool`/`docker` binaries are absent. The `cwl-utils` 0.44 `check_types()` arity error fails `mypy` on `master` at the same call site — worth someone's attention, but it belongs to neither PR.

Nothing imports `sophios.ir` yet, so this lands inert; the value is that the five phases coming next start from a data model that has already had its identity, direction and totality questions settled.

PR #424, opened 2026-09-17
Consolidate: private helpers lose documentation blocks nothing renders

vjaganat90:
Applies the docstring rule from #419–#423 to the rest of the repository. Stacked on #423.

## What the measurement said

I expected a large mechanical win and did not find one.

| | |
|---|---:|
| Total lines, `src/` + `tests/` | 31,172 |
| Docstring lines | 6,671 (21%) |
| Private functions carrying `Args:`/`Returns:` | **13** |
| `Raises:` blocks over a single `raise` | 6 |
| `Args:` entries that only restate the name and type | 5 |
| Structural duplication at `--min-similarity-lines=8` | **0** |

So the mechanically-removable class is roughly 130 lines out of 31,000. There is no code-duplication win at all.

## What shipped

**Eleven private helpers lose their generated-documentation blocks.** `docs/conf.py` sets `members`, `undoc-members` and `special-members: __init__`, and does *not* set `private-members` — so Sphinx renders every public function and every `__init__`, and skips `_private` ones entirely. Those blocks were rendered nowhere and restated the signature for a reader who has the signature.

Deliberately untouched: public functions, and `Step.__init__` / `Workflow.__init__`, which `special-members` puts in the Python API reference. Prose carrying something the signature does not is kept — `_write_to_disk` keeps the note that only the root workflow's inputs file is guaranteed complete, `_add_ranksame` keeps its graphviz reference.

**`test_generators.py`**, at 45.0% docstring, as a worked example of the judgement case. (Not the highest-ratio file in the tree — it ranks 6th; `tests/core/source_scan.py` is 55.8% and `src/sophios/lang/cwl.py` 52.9%, both small. The earlier "highest-ratio, 48%" here was wrong on both counts.)

## The result worth reading

That file gave up **9 lines** — 25 removed, 16 added, 340 → 331. (An earlier revision of this description said 13; that was not measured, and the commit message still carries the wrong figure.)

Every claim that names something a reader could break was kept: deleting `to_yml`'s desugaring call makes every mapping-form document raise `KeyError: 0`; replacing `workflows()`'s body with `st.none()` leaves the suite green; `!ii` constrains no literal to the type it binds, so a threshold demanding every draw compile would be flaky rather than stricter. What went was the framing around them — review-severity language, and the account of how each was found.

**The existing prose is largely earning its place.** It states domain facts the code cannot carry and names specific wrong edits, which is exactly the test. The 6,671 docstring lines are not 6,671 lines of bloat, and a blanket pass over them would remove signal rather than noise.

## What this deliberately does not do

It does not touch `equivalence.py`, `compiler.py`, `parser.py` or the rest of `src/`. Extrapolating from one file is weak evidence, and a docstring cut breaks no test — so nothing would catch it if I removed something load-bearing. A file-by-file pass is available; it should be a decision per file, not a sweep.

## The boundary scans stay on the lane that owns the invariant

These tests do not share one execution profile, so they no longer share one marker.

- `test_zone_boundary.py` remains in the blocking Spec 0 lane. A core/contrib import breach is a language-boundary failure and must fail the pull request.
- `test_leak_boundary.py` remains in the environment-dependent PR lane because it drives the compiler as well as inspecting source.
- The compiler-driving settings cases in `test_settings_boundary.py` remain on PR lanes.
- Only the repository-wide settings AST scans carry `sanity` and move to the named weekly step. The packaging lanes continue to exclude that marker.

The census still requires every test to be collected by at least one configured lane, so moving or renaming any selection remains mechanically checked.

## `test_prose_boundary` is deleted

290 lines out, 20 in. It carried two rules that were not the same kind of thing, and only one of them worked.

**Tracker identifiers survive**, as a `pygrep` hook beside the other pattern hooks in `.pre-commit-config.yaml`. They are a closed set of shapes — `P30`, `CE-11`, `CR-104`, `T2.4`, `#383` — so a pattern is the right tool, and every failure it had was a missing shape rather than a wrong method. Measured over all 148 tracked modules before shipping: zero matches, so it fires on new violations and not on the existing tree. A planted `# P30 says emission is canonical` is rejected with `file:line:content`.

**Narration is dropped.** No regex decides whether a sentence narrates its own history. Presenting a phrase list as a boundary is what let this report a clean tree through four consecutive reviews while the phrasings it existed to catch sat in the source — each round found the rule wrong, never the tree. A reviewer catches that; a pattern does not, and one that claims to is worse than none.

**What this trades away, stated plainly:** nothing in CI runs pre-commit, so this is enforced for anyone who has run `pre-commit install` and advisory for everyone else. That is the intended bargain — the identifiers are a house convention, not a correctness property, and they were already leaving the blocking lanes in the change above. It also puts the rule where a new contributor should meet it: at commit time with a one-line message, not twenty minutes into a lane.

## The lane census asks pytest instead of modelling it

**470 lines → 273.** The census parsed each workflow's pytest invocation and then reimplemented pytest's own selection — `-k` matching, `-m` marker expressions, path and node-id resolution — with five meta-tests holding the model to the tool. It lost that argument twice:

- it matched a test's name but not its module's, so renaming `test_tool_builder.py` silently dropped 3 tests from 3 lanes;
- it ignored markers and case, so `-k "not FAST"` read as *selecting* every `@pytest.mark.fast` test where pytest deselects all of them.

Each invocation's arguments now go to `pytest --collect-only`, and the node ids it reports are the answer. Asking pytest cannot disagree with pytest.

What's still assumed about the command line is one list: arguments that change what pytest *reports* rather than what it selects (`-vv` overrides the `-q` this parses; `--workers` wants a plugin a collection doesn't). Everything that decides selection passes through untouched.

The claim is simpler and stronger too. It was *"a test whose marker the packaging lane excludes must be named by a main lane"*; it is now **"every test is collected by something"** — and a lane naming a file that no longer exists fails it, which the model checked separately.

Thirty distinct invocations are collected sequentially because collection imports `test_setup`, which rewrites one shared generated schema; concurrent collectors made the census race with itself. The census remains `slow`. Mutation-verified three ways: a `sanity`-marked test in a file no lane names, a lane pointing at a missing file, and a collector that ignores its arguments each fail it. Measured stable across three consecutive runs.

`_marked_tests`, `_test_files` and `_tests_importing` stay — the Windows leg and the cwltool marker read the *source*, which isn't something pytest answers.

## Review follow-ups

- **`_prepare_compilation_state`** was the one private function in `src/` still carrying a field block after the pass removed eleven — a fair catch against a commit subject that reads as absolute. It is private in a module Sphinx *does* render, so the block reached no documentation either. What it said is worth keeping (which exceptions escape a two-hundred-line helper is not in the signature), so it is prose now rather than a deletion. **Zero** private functions in `src/` carry one.
- **`test_generators.py`'s `st.text()` sentence** had lost its referent: no call in `ast_strategies.py` passes bare text — every one names an alphabet, which was the point the sentence was making. It now says that and names `edge_names`, a line a reader can open.
- **The measurement in "the result worth reading" was wrong** — 9 lines, not 13 — and the ratio claim was wrong twice. Both corrected above. Since the whole argument of this PR rests on that number, it needed correcting rather than shipping.

Not taken here: `test_run_local_python_api_restores_environment_after_run` is marked `fast` without `needs_cwltool`, so `conftest.py` does not skip it off POSIX. Pre-existing, unrelated to this PR, and worth its own one-line change. Worth noting *why* the guard missed it: `test_every_cwltool_test_declares_that_it_needs_cwltool` looks for tests that import `cwltool`, and this one reaches it by monkeypatching `run_local.cwltool.main` instead.


darkclad (commented):
Review at `d5394a5`, scoped to the two commits this adds over #423 (`2e9e606`, `d5394a5`) -- 6 files, +25/-115, docstrings only, no executable line changed. #424 contains #423 exactly, with no divergence either way.

**The central claim checks out.** `docs/conf.py:61-66` sets `members`, `undoc-members` and `special-members: __init__` and does not set `private-members`, so autodoc skips `_private` members; there is no `autofunction`/`automethod`/`:private-members:` override anywhere under `docs/`, so none of the eleven was reachable by an explicit directive either. An AST audit of #423's head finds exactly eleven private non-dunder functions carrying `Args:`/`Returns:`, and this removes all eleven; the two survivors with field blocks are `Step.__init__` and `Workflow.__init__`, which is what the description says it kept. That also reconciles the census: 11 + 2 = 13 when counting `Args:`/`Returns:` only.

**Nothing load-bearing went.** Every removed entry either restates the signature or is duplicated in a public docstring Sphinx does render:

- `_write_to_disk`'s `Args:` block is verbatim on the public `write_to_disk` three lines above it (`input_output.py:50-55`), `relative_run_path` included.
- "virtual `.cwl` path for compiler bookkeeping" survives at `workflow.py:384` and `tool_builder.py:490`.
- `Step._get_input`'s `Raises: AttributeError` survives on `lookup_parameter` (`_workflow_runtime.py:126`), where the `raise` is, and on the public `get_output` and `set_input`.
- `_compile`'s note that `write_to_disk=True` emits under `autogenerated/` survives at `_workflow_runtime.py:461`.
- `GraphReps` as "a tuple of a GraphViz DiGraph and a networkx DiGraph" survives on the public `add_graph_edge` and `add_subgraphs`, and `GraphReps` is a `NamedTuple` with typed fields regardless.

**Lint and tests.** Every touched function keeps a summary line, so `missing-function-docstring` -- which pyproject does not disable -- still passes; `docparams` is not enabled, so a fully absent field block is not checked. No changed line exceeds `max-line-length=120`. `workflow.py` goes 993 -> 936, still over `max-module-lines=750` but moving the right way. The `test_prose_boundary.py` regexes are clean on all six files. In `test_generators.py` the rewritten docstring still matches the body: both surface forms and the bare-majority bar are asserted at lines 165-171, and the inline comment at line 156 that defers to "the ill-typed `!ii` literal the docstring describes" still has its referent.

**One number in the description does not survive checking.** "That file gave up **13 lines**" -- it gave up 9. `d5394a5` is `16 25` on `test_generators.py`: 25 removed, 16 added, net 9, and the file goes 340 -> 331 lines. Counting docstring lines only gives the same answer: 158 -> 149. The commit message repeats "Thirteen lines". Since "the result worth reading" is the measurement the whole argument rests on, it is worth correcting rather than shipping at 13. Related: the file is 46.5% docstring by my count rather than 48%, and it is the highest-ratio `tests/core/test_*.py` module but not the highest-ratio file in the tree -- `tests/core/source_scan.py` is 56% and `src/sophios/lang/cwl.py` 53%, both small.

**One private helper still carries a block.** `src/sophios/compiler.py:222` `_prepare_compilation_state` keeps a `Raises:` block. That is consistent with the stated scope -- the description excludes `compiler.py` deliberately, and the census counted `Args:`/`Returns:` only, so it was never in the eleven. But the commit subject reads as an absolute ("private helpers do not carry generated-documentation blocks") and this is a single block in a file the pass otherwise leaves alone. Either take it or narrow the subject.

**Ran the suite.** Set up Python 3.13 with the `test` and `mypy-types` extras and ran #424 and #423 side by side, so every result is differential.

- Collection: 1093 tests, no import errors.
- `pytest -m fast`: 981 passed, 3 skipped, 1 failed -- identical counts and the same test on both trees.
- The CI lane that covers the edited file (`test_hermeticity.py test_generators.py test_equivalence.py test_equivalences.py -m "not skip_pypi_ci"`): 94 passed, 1 skipped on both.
- Full `pytest tests/`: 1043 passed, 33 failed, 17 skipped -- and the two failure sets are identical, so this PR neither introduces nor fixes one. All 33 are environment: 26 are `wic015` (no docker on this box), 5 are the REST tests running workflows downstream of that, one is `[WinError 2]`, and one is `test_run_local_python_api_restores_environment_after_run`, which monkeypatches `run_local.cwltool.main` while `run_local.py:19-23` swallows that import on Windows because cwltool pulls in `pwd`. That last one is marked `fast` but not `needs_cwltool`, so `conftest.py` does not skip it off POSIX -- a pre-existing marker gap, visible identically on #423, and worth a separate one-line fix rather than anything here.
- `pylint src/ tests/`: 205 messages on both trees, differing only in the line numbers that moved; zero docstring messages. 9.80/10 against `fail-under=9.25`.
- `mypy src/ examples/ tests/`: clean on both, 144 files.

**And the stronger check.** Stripping every docstring and comparing the normalized ASTs, all six changed files are byte-identical between #423 and #424. Nothing in `src/` or `tests/` reads `__doc__`, so the only consumers of this text are Sphinx and the static prose scans -- and those scans pass. The change is inert by construction, not just green by observation.

**The stack, not the diff, is what gates this.** `master` moved to `c0c5dd6` while this was under review: #420 and #421 are squash-merged. This branch still carries #422 and #423 in full -- verified by content rather than SHA, since the diff from #422's head to this one touches none of #422's files. Merging this lands 19 files and +960/-135 against `master`: about +304 of that is #422 (the `wic025`/`wic026` codes, `compiler.py`, 60 lines of language reference) and about +630 is #423 (the new `src/sophios/ir/` package and `test_lower.py`). Neither has a review on it, and the two new diagnostic codes are contract a caller can match on.

So this review covers the docstring pass and nothing else. Reviewed and merged in stack order -- #422, then #423, then this -- each one is small. Merged from here, three PRs land on the strength of a review that read the docstrings.

tests/core/test_generators.py
```diff
@@ -1,12 +1,10 @@
...
-A property is only as strong as its generator. Repeated defects traced
-every High-severity finding to a blind spot: totality quantified over
-`st.text()`, which essentially never forms valid YAML with an alias; a
-document generator that produced only mapping-form `in:`-only documents made
-outputs, sequence steps and nested sidecars structurally invisible to every
-property it fed. A generator that silently stops producing a construct
-disables every property depending on it and nothing else notices — so
+A property is only as strong as its generator, and a generator that stops
+producing a construct disables every property depending on it with nothing
+else noticing. `st.text()` essentially never forms valid YAML with an alias;
```
darkclad: **⚪ The surviving `st.text()` sentence has lost its referent.**

The old text tied it to something concrete -- "totality quantified over `st.text()`" -- so a reader knew which property the example was about. On its own the claim now describes code that is not here: there is no unconstrained `st.text()` anywhere in `ast_strategies.py`, every call passes an alphabet.

The fact is worth keeping; it is the reason `edge_names` is `st.text('abcdefgh', min_size=1, max_size=4)` at `ast_strategies.py:120` rather than bare text. Saying that anchors it to a line a reader can open, which is the test the rest of this pass applies.

vjaganat90: Fixed — you were right that it described code that is not here. Every call in `ast_strategies.py` names an alphabet, so the sentence now says that and points at `edge_names`.

Also took the two other findings: `_prepare_compilation_state` was the last private field block in `src/` and is prose now, and the "13 lines" figure was wrong — it is 9 (25 removed, 16 added, 340 to 331), with the ratio 45.0% and 6th rather than "highest at 48%". Corrected in the description; the commit message still carries the old number, which would need a history rewrite to fix.

darkclad (commented):
Re-review at `00e52f9`, rebased onto `74caca6`.

**The earlier findings are all addressed.** The "13 lines" figure is corrected in the body. `_prepare_compilation_state` no longer carries a field block, and its new prose is accurate against the body: `ValueError` on missing `steps`, `TypeError` on `$namespaces`/`$schemas`. The `st.text()` sentence is anchored to `edge_names`. The two commits from the first round survived the rebase patch-identical (`git patch-id`).

**Checked, no issues:**

- `7025b01`: both slimmed docstrings match their bodies, and `_workflow_runtime` appears in no `automodule`, so nothing rendered is lost.
- `4a9daad`: the `pygrep` entry matches every shape the deleted scan did, and the `design_docs/` exemption for `Spec N` holds. Over 148 Python files it finds 0 hits, as stated. Planted `P30`, `CE-11`, `#412`, `Tasks 3-7` and `Spec 2` are caught; `wic006` and a `design_docs/`-cited `Spec 1` are not.
- Lint: pylint reports the same 207 messages as `master`; mypy is clean.
- Full suite, `master` against this head, each run with its own isolated `~/wic` config: exactly one test function fails here that passes on `master`. It is `test_no_test_is_collected_by_nothing`, the first inline comment. The other 40 failures are identical on both and environmental (no docker, POSIX-only cwltool).

**One blocker, one decision, two nits.** The new census races with itself. It has failed `Lint And Test` on both OS legs, and the current head has no completed `Lint And Test` run. Separately, the `sanity` move takes 27 compiler-driving tests off every PR under a label saying they compile nothing. Details inline.

tests/core/test_ci_coverage.py
```diff
@@ -30,149 +26,109 @@
...
-            continue
-        if not (REPO_ROOT / token.split('::', 1)[0]).exists():
-            raise ValueError(f'pytest invocation names a path that does not exist: {token!r}')
-        paths.append(token)
-    return paths
+def _covered() -> set[str]:
+    """Every test any configured invocation collects."""
+    argvs = sorted({tuple(argv) for lane in sorted(WORKFLOWS.glob('*.yml'))
+                    for argv in _invocations(lane)})
+    with ThreadPoolExecutor(max_workers=8) as pool:
```
darkclad: **🔴 The census races with itself and fails `Lint And Test` on both legs.**

Every collection runs with `cwd=REPO_ROOT`, and importing `tests/core/test_setup.py` calls `get_validator(..., write_to_disk=True)` at module level (`test_setup.py:89`). That truncates and rewrites `autogenerated/schemas/wic.json` (`wic_schema.py:653`), then reads it straight back. Eight concurrent collections that import it, through `test_examples.py`, `test_fuzzy_compile.py` or a full sweep, read each other's truncated file. `json.loads('')` raises, collection exits 2, and the assert at line 85 fails.

CI has already hit this. Run 35273025792 failed `PyTest Spec 0 Invariants` on both `ubuntu-latest` and `windows-latest` with `JSONDecodeError: Expecting value: line 1 column 1 (char 0)`, and each leg lost the race on a different invocation. The current head's run, 35288537859, was cancelled, so no green `Lint And Test` exists for it. Locally, on this head with the same config, it fails 3 of 3 at `max_workers=8` and passes at `max_workers=1` (74s against 12s).

Collecting sequentially is the smallest fix. Making `test_setup`'s import side-effect-free is the real one.

tests/core/test_leak_boundary.py
```diff
@@ -41,6 +41,11 @@
 from .ast_strategies import passthrough_keys, passthrough_values
 from .compile_harness import COMPILED, FAST, compile_cwl, compile_info
 
+#: A static scan over the repository, not a claim about a compilation.
+#: Collected by run_workflows_weekly.yml and by no lane that blocks a pull
+#: request -- these guard the codebase's shape, which changes slowly.
+pytestmark = pytest.mark.sanity
```
darkclad: **🟡 This file compiles workflows, so `sanity` takes compiler coverage off every PR.**

The comment above calls it "a static scan over the repository, not a claim about a compilation". The module docstring says the opposite: "checked against the real compiler... These run the actual `compile_workflow`". Its properties assert that passthrough keys survive compilation byte-identically, that unknown keys change nothing else, and that the residue validates under cwltool. The same holds for `test_settings_boundary.py:40`. `test_every_setting_reaches_the_compiler` and `test_the_workflow_path_reaches_the_compiler` drive `main._main()` with a real argv; they are the regression tests for CLI flags silently reaching the compiler as their defaults.

I asked pytest what `build_wheel.yml` and `lint_and_test.yml` select before and after. 180 test cases that run on every PR today run only in the weekly lane after this: 161 from settings, 14 from leak, 5 from zone. 27 of them drive the compiler: all 14 leak tests and 13 settings tests. The census can't flag this, because it counts the weekly lane as coverage.

The PR's rationale ("no compilation, no containers"; "a workflow change failing on a docstring") fits `test_zone_boundary` and the AST scans in the settings file. It doesn't fit these 27. Keep `test_leak_boundary.py` and the two compiler-driving settings tests on a PR lane, and mark only the static scans `sanity`: that gets the trade the PR describes without the coverage loss.

src/sophios/compiler.py
```diff
@@ -229,9 +229,11 @@ def _prepare_compilation_state(yaml_tree_ast: YamlTree,
                                testing: bool) -> _WorkflowSetup:
     """Normalizes the yml AST and initializes the mutable state used by compile_workflow_once's per-step loop.
 
-    Raises:
-        ValueError: If the yml AST is malformed (e.g. missing steps).
-        TypeError: If a top-level tag has the wrong type (e.g. $namespaces/$schemas)
+    Escapes as `ValueError` on a malformed AST (missing `steps`) and as
+    `TypeError` on a top-level tag of the wrong type (`$namespaces`,
+    `$schemas`). Stated in prose rather than a field block: this is private, so
```
darkclad: **⚪ The new sentence narrates the pass.**

"Stated in prose rather than a field block: this is private, so Sphinx renders nothing for it, and the field block was the last one left after the pass that removed the other eleven." The first sentence is the fact a caller needs. This one is the account of how the docstring came to be, which is what this PR removes everywhere else. With the narration regex gone, the PR body names review as the guard for this, so here is review. Dropping the sentence keeps the fact.

.github/workflows/lint_and_test.yml
```diff
@@ -229,7 +229,7 @@ jobs:
       # needs containers; both are static checks over the repository.
       - name: PyTest Spec 0 Invariants
         if: always()
-        run: cd sophios/ && pytest tests/core/test_zone_boundary.py tests/core/test_entry_points.py tests/core/test_ci_coverage.py
+        run: cd sophios/ && pytest tests/core/test_entry_points.py tests/core/test_ci_coverage.py
```
darkclad: **⚪ The comment above this step still describes a file it no longer runs.**

Lines 227-229 say the step covers "both Spec 0 mechanisms -- the core/contrib import boundary and entry-point resolution". `test_zone_boundary.py` is gone from the invocation, so the step now runs entry points and the census. Worth updating with the move, or the next reader will look for the import boundary here.

tests/core/test_ci_coverage.py
vjaganat90: Fixed in `fb51d60e`: collection is sequential. Importing `test_setup` writes one shared generated schema, so the census no longer lets collectors race over that file.

tests/core/test_leak_boundary.py
vjaganat90: Fixed in `fb51d60e`: `test_leak_boundary.py` is back on PR lanes; only the repository-wide settings AST scans carry `sanity`; the thirteen CLI-to-compiler settings cases remain unmarked and on PR lanes. `test_zone_boundary.py` is restored to blocking Spec 0 because a core/contrib breach must fail the build.

src/sophios/compiler.py
vjaganat90: Fixed in `fb51d60e`: removed the pass and Sphinx history. The docstring now states only the exceptions that escape.

.github/workflows/lint_and_test.yml
vjaganat90: Fixed in `fb51d60e`: restored `test_zone_boundary.py` to the Spec 0 invocation, so the comment and the blocking import-boundary gate agree.

vjaganat90:
Addressed the review findings in `fb51d60e`:

- serialized the lane census around its shared generated schema;
- kept core/contrib, leak-boundary, and compiler-driving settings checks blocking on pull requests;
- limited the weekly `sanity` move to repository-wide settings AST scans;
- removed the remaining review-history narration.

darkclad (approved):
LGTM

PR #425, opened 2026-09-17
Bug Fix: source inlining preserves the workflow call boundary

vjaganat90:
Stacked on #424.

## Problem

The source-level inliner removed a subworkflow call before it had discharged the call boundary. Its parameter substitution recognized only the obsolete `~formal` spelling, while the current Sophios language represents a workflow-input reference as the bare formal name. Inlining therefore deleted the child `inputs:` declaration and left an unbound name behind, producing `wic011` in the inline variants of the docking workflows.

That input failure concealed a second loss of meaning. A wrapper can export a child workflow output under an explicit Sophios edge name. Removing the wrapper without relocating that definition disconnects every downstream `!*` consumer and causes the compiler to manufacture required root inputs with no job values.

## Boundary rule

A workflow call is now removed only when the source transform can discharge the complete boundary:

- Authored `parentargs` and direct `wic: steps:` overrides are merged with the same precedence as compilation. Immediate-step overrides receive the same formal-to-actual substitution before they enter the child metadata, so a parent-authored override cannot reintroduce a deleted formal.
- The call must bind every declared child formal exactly once and may not bind undeclared formals.
- Exact bare formal references are replaced simultaneously in immediate child steps, explicit nested-call arguments, and child step metadata. Omitted nested arguments remain omitted, so the compiler retains authority to infer their source after the enclosing scope disappears. Tagged literals, explicit edge references, and raw CWL remain opaque, and substitution never crosses into a nested child's formal scope.
- Each anchored wrapper output must resolve through the child's declared `outputs[*].outputSource` to one unambiguous immediate producer. The edge definition moves to that producer; when the producer is another subworkflow, it moves onto that call and is discharged on the next inlining pass.
- Scatter, `scatterMethod`, `when`, unknown invocation semantics, incomplete interfaces, ambiguous outputs, and conflicting output anchors make the boundary non-inlineable. Descendants may still inline, but that call remains nested rather than being changed silently.

Discovery and mutation share one immutable inline plan. The inliner therefore cannot advertise a boundary using weaker rules than the rewrite that removes it.

## Result

The inline docking roots now preserve both sides of the call: the renamed formal inputs resolve to their actual values, and `pdb.pdbqt` / `box.pdb` remain definitions on the concrete producing steps. Their downstream bindings no longer become synthetic unvalued workflow inputs.

The partition oracle now generates deliberately different formal and actual names, so deleting a child interface without substitution is observable rather than hidden by identity bindings.

## Removed

This makes the legacy `~formal` substitution path, warning-based partial handling, and the unused tool-registry parameter obsolete. It also removes the dead compiled-CWL inliner and its path-rewriting helper; nothing called that path, and it expected dictionary-form steps while the compiler emits lists. The associated imports, stale compiler claim, false manufactured-document inventory entry, and stale test commentary are removed with it.


src/sophios/inlineing.py
```diff
@@ -16,29 +15,231 @@
...
+        effective_inputs = nested_call.get('in', {}) if isinstance(nested_call, dict) else {}
+        if not isinstance(nested_inputs, dict) or not isinstance(effective_inputs, dict):
+            continue
+        implicit = {name: copy.deepcopy(bindings[name])
+                    for name in nested_inputs
+                    if name in bindings and name not in effective_inputs}
+        if implicit:
+            parent_inputs = parentargs.setdefault('in', {})
+            if isinstance(parent_inputs, dict):
+                parent_inputs.update(implicit)
```
darkclad: **This changes which value an omitted nested argument receives.** The comment above says the compiler supplies an omitted argument as the same-named bare reference. That bare reference never reaches the compiler's bare-reference branch. `compile_workflow_once` snapshots `args_provided` (`compiler.py:724`) before it fills in `in[key] = key` (`:733`). An omitted key therefore goes to the `args_required` loop and on to `perform_edge_inference`. Inference takes the nearest upstream output of a compatible type, and if there is none it manufactures a `{step}___{key}` input. It never reads the same-named formal of the enclosing scope.

So materializing `bindings[name]` rebinds the argument:

```yaml
# root: inputs {actual: string}; calls child.wic with in: {name: actual}
# child.wic
inputs: {name: {type: string}}
steps:
  poly:
    in: {value: !ii 3}   # emits a string
  nested.wic: {}         # declares name: string; omitted here
```

```
                          nested.wic `name` is fed by
before inlining           child__step__1__poly/value    (inferred)
after inlining, base      oracle__step__1__poly/value   (still inferred)
after inlining, 26bc9c4   actual                        (materialized)
```

With no upstream producer, the un-inlined document compiles `name` to a manufactured root input. After this inline, `name` is bound to `actual` and the root interface loses that input.

Dropping lines 75–90 fixes both cases, at no measurable cost. With this block removed, inlining every corpus root gives results identical to `26bc9c4`, and `elm`, `vs_demo_2` and `vs_demo_3` still compile with the same inputs and isomorphic graphs. The only test that fails is `test_nested_child_keeps_its_formal_scope`, because its `== 'actual'` assertion pins this behavior. That test never compiles `tree.yml` before inlining. Comparing against that compile, as `test_equivalences` already does for splits, would have caught this.

src/sophios/inlineing.py
```diff
@@ -16,29 +15,231 @@
...
+    """Discharge a complete interface without crossing a nested formal scope."""
+    for index, step in enumerate(subtree['steps']):
+        step_key = utils.require_step_id(step)
+        _replace_input_references(step.get('in'), bindings)
+        parentargs = step.get('parentargs')
+        if isinstance(parentargs, dict):
+            _replace_input_references(parentargs.get('in'), bindings)
+        metadata = _step_metadata(subtree, index, step_key)
+        if metadata is not None:
+            _replace_input_references(metadata.get('in'), bindings)
```
darkclad, line 81: **A parent-authored override can undo this rewrite.** This line rewrites the child's own `wic: steps:`. `merge_yml_trees` runs before inlining (`main.py:57`), and afterwards the parent's `wic: steps: (i, child.wic): wic:` block is still in the parent. `inline_subworkflow` then merges that block back over the rewritten child tag, with the parent winning (`:371`).

A tool step doesn't notice, because `merge_yml_trees` already folded its metadata into the step and the compiler doesn't read it again. A nested *workflow* call is different: both `compile_workflow_once` and `_call_arguments` read it again, so the unrewritten formal comes back.

```yaml
# root
inputs: {actual: {type: string}}
steps:
  child.wic:
    in: {formal: actual}
wic:
  steps:
    (1, child.wic):
      wic:
        steps:
          (1, nested.wic): {in: {formal: formal}}   # child-scope reference
```

This document compiles as written. After one inline, the root carries `(1, nested.wic): in: {formal: formal}` and compiling fails with `wic011 … forget to use !ii before formal`, which is the error class this PR removes. Applying the same `bindings` to the `steps` entries of `parent_wic_tag` before the merge would close it. `test_child_step_metadata_cannot_reintroduce_a_deleted_formal` builds its tree without `merge_yml_trees`, so it cannot reach this path. This is not a regression, since the base fails on any bare formal. Non-blocking.

darkclad (commented):
Review at `26bc9c4` against #424 (`00e52f9`), not `master`. This PR is stacked, so its own change is one commit touching 11 files.

Verified: `test_source_inlining.py`, `test_equivalences.py`, `test_reference_compatibility.py` and `test_manufactured_documents.py` pass, except for the 4 `cwltool_agrees` cases, which also fail on `master` (`cwl-utils` 0.44 `check_types()` drift). The fast lane is **1096 passed**, against 1078 at the base, with the same single pre-existing failure. `mypy src/` is clean. Pylint on `inlineing.py` drops from three `too-many-locals` hits to one. `autopep8 --max-line-length 120` is clean.

**The headline claim holds against the corpus.** I used mm-workflows, image-workflows, biobb_adapters and `docs/tutorials`. Each root went through the CLI's own sequence (`read_ast_from_disk` → `merge_yml_trees` → `python_script_generate_cwl` → inline loop to a fixpoint) and was compiled with `testing=False`:

| root | base | `26bc9c4` |
|---|---|---|
| `elm`, `vs_demo_2`, `vs_demo_3` and both `_weekly` wrappers | `wic011` | compiles, with the same input count as the un-inlined root, 0 unvalued inputs and the same values |
| every other root with a call (20) | | identical to base |

The number of calls offered for inlining is identical at both revisions too, so the stricter boundary rule gives up no inlining anywhere in the corpus. `elm`, `vs_demo_2` and `vs_demo_3` are DiGraph-isomorphic to their un-inlined compile.

The two `_weekly` wrappers compile and bind the same values, but their graphs are not isomorphic. In the original, the explicit `pdb.pdbqt`/`box.pdb`/`binding_data.txt` edges are drawn from the wrapper node; after inlining they are drawn from the concrete producers. The anchors did land on the producers that each child's `outputSource` names. All five roots are on mm-workflows' `large_workflows` list, so `test_inline_subworkflows` covers none of them. Separately, `analysis`, `basic`, `docking_stability` and `setup_vac_min` fail `wic025` after inlining at both revisions. That is pre-existing and outside this PR.

Discovery and mutation sharing one immutable plan is the right structure, and so is the mutator re-proving that plan rather than trusting its caller. Deleting `inline_subworkflow_cwl` and `apply_args` is overdue. Making the partition oracle's formal names differ from the actual names is what lets the property see this class of bug at all.

**One thing blocks.** The implicit-capture materialization in `_bind_inputs` rebinds an omitted nested argument by name, but the compiler resolves it by inference. The result is a different edge, and it regresses a case the base inliner got right. Removing the materialization keeps every corpus result above. The second comment covers a parent `wic:` override that brings a rewritten formal back. That one is not a regression.

src/sophios/inlineing.py
vjaganat90: Fixed in `a00de48b`: removed implicit materialization. Omitted nested arguments stay omitted and retain compiler inference; the regression pins the preceding `poly/value` edge as the source before and after inlining.

src/sophios/inlineing.py
vjaganat90: Fixed in `a00de48b`: parent-authored immediate-step metadata is rewritten with the same formal-to-actual map before the parent and child `wic` metadata merge, so it cannot restore a deleted formal. The regression exercises the real `merge_yml_trees` sequence.

vjaganat90:
Addressed the review findings in `a00de48b`:

- omitted nested arguments remain omitted, preserving compiler inference across the removed boundary;
- merged parent-authored step metadata receives the same formal-to-actual substitution and cannot reintroduce a deleted formal.

darkclad (approved):
LGTM.

Re-reviewed at `a00de48`. Both comments are addressed. Omitted nested arguments stay omitted, so inference decides them on both sides of the removed boundary. Parent-authored overrides get the same formal-to-actual map before the merge, and it stops at the immediate steps, so it never enters a grandchild's scope. The rebase left `0189c50` unchanged from `26bc9c4`.

Checked with runs, not just by reading:

```
                                  before inlining                 after inlining
omitted arg, upstream producer    child__step__1__poly/value      oracle__step__1__poly/value
omitted arg, no producer          manufactured root input         manufactured root input (renamed)
parent override of grandchild     compiles                        compiles; fully inlined mk_file <- actual
```

The fast lane is 1097 passed against 1078 at `fb51d60`, with the same single pre-existing failure. Corpus results match round 1: `elm`, `vs_demo_2` and `vs_demo_3` and their `_weekly` wrappers compile once inlined, with the same input counts and no unvalued inputs, and every other root is unchanged. `mypy src/` and `autopep8` are clean.

PR #426, opened 2026-09-18
Consolidate: one error type out of the Python API, and codes in their own module

### LINKED ISSUE #111: Error messages in WIC Python API
*Moving https://github.com/PolusAI/polus-plugins/issues/467 to correct repo*

**Description**

When WIC fails to compile with WIC Python API, no helpful error message is logged back to the user.

**Proposal**

Running WIC on the command line directly provides better feedback so let's look integrating those.

**Additional context**

Those problems are recurrent when using the WIC Python API to develop workflows.

vjaganat90:
Closes #111. Stacked on #425. Its diagnostic contract is independent of the inlining change, but the branch now carries the complete #424 → #425 ancestry.

## What #111 actually is now

The issue (2023) says the Python API logs "no helpful error message" and that the CLI does better. The first half has quietly stopped being true — the diagnostics work landed real messages:

```
error [wic025] '!* absent' names an edge nothing defines. Define it with '!& absent'
on the producing step's out:, or declare it in this document's inputs: and bind it by name.
```

What remained is structural. The Python API had **two unrelated error vocabularies**:

| | |
|---|---|
| `api/python/_errors.py` | four bare `Exception` subclasses, no codes, docstring reads *"Internal exception types"*, exported nowhere |
| `lang/diagnostics.SophiosError` | coded, structured, what the CLI catches and prints |

`Workflow.compile()` could raise either. Catching the second meant importing `sophios.lang.diagnostics` — a path that reads like somewhere a user should not be.

## One type out

The four are `SophiosError` subclasses now, so one `except` catches everything and `.diagnostics` carries the codes whichever failed. Both are re-exported from `sophios.api.python.workflow` — where `compile()` lives, and therefore where a caller looks.

```python
from sophios.api.python.workflow import Workflow, SophiosError

try:
    compiled = Workflow(steps, "wf").compile()
except SophiosError as e:
    for d in e.diagnostics:
        print(d.code, d.message)
```

The package root still exposes modules only; that boundary is unchanged.

## `Code` → `SophiosErrorCode`, in its own module

`Code` was too general a name for a symbol that is contract, and it lived in with the reporting machinery. It is now `SophiosErrorCode` in `src/sophios/lang/error_codes.py` — a caller matching on a code should not have to import the plumbing around it.

198 references across 20 files. The `wic0NN` *values* are unchanged, so nothing documented in the language reference moves.

## Two ranges, one enum

| Range | Means | Documented in |
|---|---|---|
| `wic0NN` | the document said something the language does not accept | the language reference |
| `api0NN` | the document is fine; the **call** was not | here |

One type, because a caller wants one thing to match on. Separate ranges, because a `wic0NN` in an API position sends a reader to a section of the reference that does not describe their problem. `SophiosErrorCode.is_language` makes that split routable rather than a string-prefix convention.

Four new codes — `api001` `INVALID_INPUT_VALUE`, `api002` `INVALID_STEP`, `api003` `INVALID_LINK`, `api004` `INVALID_TOOL` — with four provocations, as `test_every_code_has_a_registered_provocation` requires.

## Verification

Mutation-verified three ways: make `ApiError` stop subclassing `SophiosError` (12 tests fail), give an error the wrong code (2 fail), make `is_language` stop distinguishing (4 fail).

`mypy`, `autopep8` clean; `pylint` 9.81 against the 9.25 floor; 1149 fast tests pass.

## One judgement call

`InvalidCLTError` previously subclassed `ValueError` and now subclasses `SophiosError`. Anyone catching `ValueError` around a tool load would stop catching it. Nothing in the repo does, and the unification is the point of the change — but it is a behaviour change rather than a pure addition, so it is called out rather than buried.


src/sophios/api/python/_errors.py
```diff
@@ -1,17 +1,51 @@
...
+    Subclasses name the code; the message is whatever the raise site says, in
+    as many lines as it needs -- the same shape as `SophiosError.error`.
+    """
 
+    code: ClassVar[SophiosErrorCode]
 
-class InvalidCLTError(ValueError):
-    pass
+    def __init__(self, *messages: str) -> None:
+        super().__init__(Diagnostic(Severity.ERROR, self.code, message) for message in messages)
```
darkclad, line 27: **An `ApiError` raised with no message no longer constructs.** `SophiosError.__init__` rejects an empty diagnostics list (`diagnostics.py:123`), and this builds one diagnostic per message. So `InvalidInputValueError()` raises `ValueError('SophiosError requires at least one diagnostic')` from inside its own constructor. `_utils.py:170` does exactly that, and a user reaches it through the public API: a `config_path` YAML that binds `file: {class: File}` with no `location` or `path`, passed to `Step(clt_path=cwl_adapters/append.cwl, config_path=...)`.

| | base `a00de48` | `dbb553f` |
|---|---|---|
| raised | `InvalidInputValueError` | `builtins.ValueError` |
| caught by `except SophiosError` | n/a | **no** |

That breaks the one-`except` contract this PR sets up, on a raise site that already exists. There are two fixes: give `_utils.py:170` a message (`f"{data['class']} value has no location or path"`), or have `ApiError` fall back to a default message when it is given none. The second also covers the next bare raise, but it makes an error with nothing to say representable again. That is what `SophiosError`'s docstring rules out, so the first fix plus a test is probably the better one.

tests/core/provocations.py
```diff
@@ -212,8 +212,42 @@ def _provoke_empty_name() -> None:
     raise SophiosError(tuple(diagnostics))
 
 
+def _provoke_invalid_input_value() -> None:
```
darkclad, line 215: **These four provocations raise the exception directly, so they prove the class carries its code, not that anything can raise it.** Delete every `raise InvalidLinkError` in `src/` and the registry still passes. The registry exists to catch exactly that, and a real `api001` provocation would have caught the empty-message constructor above. The following call paths each raise their code through the public API (checked at `dbb553f`):

- `api001`: `Step(clt_path=<adapters>/append.cwl, config_path=<yml with file: {class: File}>)`. That is the no-location case, which fails today, or `location: /no/such/file`.
- `api002`: `Workflow([append], "wf").compile()` with `append` linked to a step outside the workflow. This is already what `test_every_api_failure_is_catchable_as_one_type` builds.
- `api003`: `wf.outputs.out = 3` on any workflow.
- `api004`: `Step(clt_path="no_such_tool.cwl")`.

A small point that goes with it: `InvalidLinkError`'s docstring ("an output bound to a source that is not one of the workflow's steps") is one of its four raise sites. The other three are an incompatible type, a workflow output bound as an input, and an output bound to something that is not a port.

darkclad (commented):
Review at `dbb553f` against #425 (`a00de48`), not `master`. This PR is stacked, so its own change is one commit touching 25 files. Most of it is the mechanical `Code` → `SophiosErrorCode` rename, and no `Code` reference is left anywhere in the tree.

Verified: the fast lane is **1104 passed**, against 1093 at the base. Both have the same single pre-existing failure (`test_run_local_python_api_restores_environment_after_run`, Windows). The provocation meta-tests pass (32). `mypy src/` is clean across 72 files, and `autopep8 --max-line-length 120` is clean. The only pylint hit in the touched API modules is the pre-existing `too-many-locals` on `Step.__init__`.

The design is right. One exception type out of the API, `.diagnostics` on every failure, and a separate `api0NN` range, so that an API failure never points a reader at the language reference. That closes what is left of #111. `is_language` is a better routing key than a string prefix, and moving the enum out of the reporting module is the right cut.

**One thing blocks.** `ApiError` inherits `SophiosError`'s at-least-one-diagnostic invariant, but one existing raise site passes no message. That site is reachable from `Step(config_path=...)`, and it now escapes `except SophiosError` as a bare `ValueError`. The second comment covers why the registry didn't catch it: the four `api0NN` provocations construct the exception directly instead of provoking it. The third covers a module comment that promises more than the module does.

Small and not worth a thread: 15 of the renamed import sites import `SophiosErrorCode` from `sophios.lang.diagnostics`, the module the PR moved it out of. That works through the re-import, but it is the very dependency `error_codes.py`'s docstring says a caller should not need.

src/sophios/api/python/workflow.py
```diff
@@ -56,10 +61,19 @@
 
 
 StrPath = str | Path
+#: Re-exported so a caller can `except SophiosError` without importing
+#: `sophios.lang.diagnostics`. Everything this module raises is one of these,
```
darkclad: **"Everything this module raises is one of these" is not true of this module.** `Step(source=..., clt_path=...)` raises `TypeError`, `to_wic_yaml(inline_subworkflows=False)` on a nested workflow raises `ValueError`, and scatter and `when` misuse raise `ValueError` from `_workflow_runtime.py`. Those are call errors in exactly the `api0NN` sense, so converting them is a reasonable follow-up, not something this PR has to do. The comment should claim only what holds: what `compile()` reports, or the four `Invalid*Error` types. A reader will take this comment as the contract, and write `except SophiosError` as their only handler.

src/sophios/api/python/_errors.py
vjaganat90: Fixed in 10b4f045. The missing-location path now raises `InvalidInputValueError("File value has no location or path")`, while an invalid path value gets its own precise message. The broad `except Exception` is narrowed to the `Path(...)` conversion errors. A regression reaches this through the public `Step(clt_path=..., config_path=...)` constructor and asserts the `SophiosError` family, `api001`, and message.

tests/core/provocations.py
vjaganat90: Fixed comprehensively in 10b4f045. All four direct exception raises are gone. The registry now provokes `api001` through a malformed `Step` config, `api002` by compiling a workflow linked to an external step, `api003` by binding a workflow output to a non-port, and `api004` by loading a nonexistent tool. Deleting the production raise sites now makes the registry fail. I also broadened `InvalidLinkError`’s docstring to cover invalid source, ownership, and type relationships.

src/sophios/api/python/workflow.py
vjaganat90: Fixed in 10b4f045. The export comment and `_errors.py` module contract now explicitly cover structured compile/validation failures only and state that ordinary Python call-contract mistakes remain built-in exceptions. They no longer promise that every operation in the module is caught by `except SophiosError`.

vjaganat90:
Also addressed the import-pollution point from the review summary: every internal consumer imports `SophiosErrorCode` from `sophios.lang.error_codes`, and commit 909a7062 removes the compatibility symbol from `sophios.lang.diagnostics` altogether. This is code organization only; no import-layout or source-scan meta-tests remain. Targeted result for the functional review fixes: 145 passed; mypy clean across all 19 touched files; no full suite run.

darkclad (approved):
LGTM.

Re-reviewed at `909a706`, now on `master` (`266c19b`) after #425 merged. `a81bbb2` is an unchanged rebase of the `dbb553f` I reviewed (`git range-diff` shows `=`), so the new work is `10b4f04` and `909a706`. All three comments are addressed, and so is the import point from the summary.

Verified rather than read:

- **Empty-message constructor.** The config `File` with no `location` now raises `InvalidInputValueError` carrying `api001` and a real message, through the public `Step(config_path=...)`. Narrowing the `except` to the `Path(...)` conversion is better than the fix I suggested.
- **Provocations.** I mutated each production raise site to a plain `ValueError`, one at a time. All four `api0NN` provocations fail (`api004` covers both `InvalidCLTError` sites). The registry now proves the codes are reachable, not just that the classes carry them.
- **Imports.** No `SophiosErrorCode` import through `sophios.lang.diagnostics` is left anywhere in the tree, and `diagnostics` no longer re-exports it.

**Suite.** Fast lane: **1105 passed**, against 1093 on `master`, with the same single Windows `run_local` failure on both. The provocation and new API tests pass (40). `mypy src/` is clean over 72 files, and `autopep8 --max-line-length 120` is clean. Pylint on the touched modules is 9.97; its one note (`len()` truthiness in `SophiosError.__init__`) predates this PR.

PR #427, opened 2026-09-18
Spec 3: emit canonical CWL from the workflow graph

vjaganat90:
## Stack position

Second descendant of merged #426, stacked on bug-fix #430. This PR remains based on `master`; commit ancestry supplies the stack order.

## Semantic change

This establishes `WorkflowGraph` as a sufficient source for terminal compilation artifacts. The graph now carries structured resolved run descriptors, complete process and workflow port declarations, workflow inputs and outputs, job bindings, child workflow graphs, language and CWL versions, requirements, namespace/schema declarations, and preserved opaque CWL residue.

`emit(graph)` is a pure terminal projection. It performs no discovery, resolution, filesystem access, configuration lookup, or compiler-state lookup. Lower, Link, and Infer continue to treat opaque CWL as semantically unavailable; Emit may traverse it only to transport an owned copy byte-for-byte. The job-input projection is likewise graph-derived.

The live legacy compiler now stops before dictionary finalization, converts its separate terminal semantic components into an immutable graph, and calls the new emitter. It never gives Emit a completed CWL document and never hides one in `OpaqueCwl`. The old assembly is independently selectable only as the migration oracle.

## Closure

The old and new terminal paths are required to agree at `IDENTICAL` strength for every generated workflow artifact, including nested workflows. Determinism is stated over one graph and across interpreter hash seeds. Emitted workflows must validate as CWL v1.2. A hand-built graph emits without the bridge, and the dependency boundary has both generated coverage and a planted-breach discrimination control.

This closes the emission phase while leaving the legacy front and middle intact for the next stacked extractions. The temporary terminal bridge is intentionally visible and is removed when the legacy compiler is retired.


src/sophios/legacy_graph.py
```diff
@@ -0,0 +1,167 @@
...
+        passthrough=tuple((key, deepcopy(value)) for key, value in top.items()
+                          if key not in _WORKFLOW_FIELDS),
+        name=state.name,
+        lang_version=state.lang_version,
+        cwl_version=str(top['cwlVersion']),
+        workflow_inputs=_workflow_ports(top.get('inputs', {}), outputs=False),
+        workflow_outputs=_workflow_ports(top.get('outputs', {}), outputs=True),
+        job_bindings=tuple(JobBinding(key, deepcopy(value))
+                           for key, value in state.job_inputs.items()),
+        requirements=tuple((key, deepcopy(value)) for key, value in requirements.items()),
```
darkclad, line 96: This breaks any workflow that uses the list form of `requirements:`, which is valid CWL (`requirements: [{class: ResourceRequirement, coresMin: 1}]`). `requirements.items()` raises `AttributeError: 'list' object has no attribute 'items'`, and because `graph_from_legacy_state` runs on every compile, the `legacy_emission=True` arm crashes too. `maybe_add_requirements` only turns the value into a dict when it adds a requirement, so a list survives to this line whenever a workflow needs no scatter, `when`, `valueFrom` or subworkflow requirements.

Repro through the CLI, where the base compiles and the head fails:

```yaml
requirements:
  - class: ResourceRequirement
    coresMin: 1
steps:
  - id: touch
    in:
      filename: !ii empty.txt
```

`b434daa` writes `autogenerated/listreq.cwl` with the list unchanged. `92515ae` reports `Failed to compile` with this `AttributeError`. `compile_hermetic` shows the same split. A bare `requirements:` (which parses to `None`) also fails on this line, where the base emitted `requirements: null`.

The bridge shouldn't decide what shape `requirements` has. One fix is to carry a non-mapping value opaquely, like other unmodelled residue, and let `emit` write it back as it came. Please also add a list-form case to the IDENTICAL differential, as an explicit example or by extending `ast_strategies.workflows`. The docstring for `test_emit.py` already says generated workflows inherit that strategy's exclusions, and this is one of them.

src/sophios/ir/emit.py
```diff
@@ -0,0 +1,95 @@
...
+    known.update({name: deepcopy(value) for name, value in graph.passthrough})
+    return {name: known[name] for name in graph.field_order if name in known}
+
+
+def emit_job_inputs(graph: WorkflowGraph) -> WorkflowInputsFile:
+    """Project the concrete job input document carried by ``graph``."""
+    return {binding.name: deepcopy(binding.value) for binding in graph.job_bindings}
+
+
+def emit_visualization(graph: WorkflowGraph) -> dict[str, Any]:
```
darkclad: Non-blocking. `emit_visualization` is exported from `sophios.ir`, but nothing calls it and no test covers it. On any compiled graph it also returns `edges=()`, because `graph_from_legacy_state` never fills in `WorkflowGraph.edges`. So the claim in the PR description that visualization is "graph-derived" isn't met yet. Either add a test on a graph that has edges, or leave this function out until the Link extraction adds edges to the graph. That way no caller starts depending on an empty edge list.

darkclad (commented):
Review at `92515ae` against #430 (`b434daa`), so it covers this PR's own two commits (`63e4a17`, `92515ae`), not the #430 fix.

The design is right. Emit only reads the graph, and the graph is built from the separate pieces the compiler has before it assembles a CWL dict, not from a finished document. `legacy_emit` stays as a switchable oracle, the hand-built graph test shows the emitter works without the bridge, and the boundary guard is tested against a planted breach. Those are the right checks for this stage.

Verified: the fast lane is **1123 passed**, against 1105 passed at the base. Both have the same single pre-existing Windows failure (`test_run_local_python_api_restores_environment_after_run`). The Spec 3 lane (`test_lower.py test_emit.py`, slow and Hypothesis tests included) gives 29 passed, with the two cwltool validator tests skipped because they are in `WINDOWS_EXCLUDED`. I also compiled every `docs/tutorials/*.wic` through the CLI on base and head. All 45 generated files are byte-identical, and `fail.wic` fails on both. `mypy src/` is clean over 74 files, and `autopep8 --max-line-length 120` is clean. The only pylint hits in the touched modules are the existing size warnings on `compile_workflow_once`. `tests/core/benchmarks.py` timings are unchanged.

**One thing blocks.** The bridge assumes `requirements` is a mapping, so a workflow using CWL's list form of `requirements:` no longer compiles. The base compiles it and passes the list through unchanged. Details are inline. The generated workflows never produce that form, which is why the IDENTICAL differential didn't catch it.

src/sophios/legacy_graph.py
vjaganat90: Fixed in `5719c88b`. The bridge models only the mapping form and carries any other shape as residue, which `emit` already writes back over the modelled value — so the list survives, and so does a bare `requirements:` that parses to `None`. The bridge no longer decides the shape.

Pinned both spellings as explicit examples rather than extending the generator: `ast_strategies.workflows` has no `requirements:` dimension at all, so adding one would be a new axis across every property in the chain rather than a widening of an existing one. The test says that is why it is pinned.

src/sophios/ir/emit.py
vjaganat90: Removed in `c791f0e8`. Nothing calls it anywhere through to the head of the chain, so I would rather it return with Link, when there are edges to project. I also corrected the description, which claimed the visualization projection was graph-derived.

darkclad (approved):
LGTM.

Re-reviewed at `c791f0e`, now on `master` (`92eb2ed`) after #430 merged. `57889f6` and `1786dca` are unchanged rebases of the `63e4a17` and `92515ae` I reviewed (`git range-diff` shows `=`), so the new work is `5719c88` and `c791f0e`. Both comments are addressed.

Verified rather than read:

- **`requirements:` shape.** My CLI repro now compiles and writes the list back unchanged. In-memory probes of dict, list and bare `requirements:` match `master` exactly, and so do list `hints`, `$schemas` and a key placed before `steps`. The new pinned test fails in both cases against the old bridge (`92515ae`) and passes at the head, so it catches the bug. Pinning it instead of adding a new generator axis is reasonable, and the test's docstring explains why.
- **`emit_visualization`.** It's gone, along with its export, and the description no longer makes the claim.

**Suite.** Fast lane: **1125 passed**, against 1105 on `master`, with the same single Windows `run_local` failure on both. The Spec 3 lane (`test_lower.py test_emit.py`) gives 31 passed; the 2 skips are the cwltool validator tests in `WINDOWS_EXCLUDED`. All 45 files generated from `docs/tutorials` are byte-identical to `master`. `mypy src/` is clean over 74 files, pylint gives `emit.py` and `legacy_graph.py` 10.00, and `autopep8 --max-line-length 120` is clean.

PR #428, opened 2026-09-18
Spec 3: resolve registries and make Parse the front door

vjaganat90:
## Stack position

Second descendant of #426, stacked on #427. This PR remains based on `master`; ancestry supplies the dependency order.

## Semantic change

This establishes the typed front end:

`parse(source) → resolve(Document, RegistrySnapshot) → lower(ResolvedDocument)`

There is no YAML, mapping, or legacy-loader adapter between those phases. The temporary adapter begins only after Lower, where the still-legacy Link and Infer implementation currently takes over.

Nothing in `src/` calls `front_end` yet. The CLI and the Python API still enter through `wic_loader` and the legacy compiler, so in this PR the typed chain runs only under test. The stacked PRs below replace its consumers; the retirement PR makes it the live path.

`RegistrySnapshot` is an owned, deterministic content snapshot of tool definitions and workflow sources. Resolve performs no discovery or filesystem access. It resolves process identity, run descriptors, complete input/output interfaces, formats, defaults and optionality, nested workflow source, language version, step namespaces, and implementation precedence. Generated-tool identities are deterministic but not yet resolvable: the legacy side registers a generated CWL tool under a `python_script_<uuid4>` name, while Resolve keys on a content digest, so a `python_script` step has no definition in any snapshot. Registering it belongs with the front door that generates it. Registry iteration order has no semantic effect.

Lower now accepts the resolved form and produces complete step ports, declarations, authored bindings, child graphs, workflow interfaces, and emission descriptors directly. The syntax-only Lower entry remains temporarily for the already-landed foundation contracts; the canonical chain uses `ResolvedDocument`.

The local `!cwl` form is now usable end to end. It passes the reference through unresolved without requiring `--allow_raw_cwl`, so one explicit escape does not weaken every other binding. The language reference and generators now reflect that implemented surface.

## Closure

The typed path is required to remain `IDENTICAL` to legacy output after the post-Lower handoff. That property is a source round-trip: the bridge rebuilds the step list from the parsed source, so resolution itself is compared separately, per step and recursively, against an independent model of the registry. Separate properties establish registry-only dependence, registry-order independence, direct typed phase wiring, complete interfaces, nested source resolution, sidecar namespace precedence, implementation precedence, generated identity stability, and the raw-CWL boundary.

The corpus harness no longer discovers tool plugins or writes generated schemas during test collection. Environment-dependent registries are loaded explicitly at execution; workflow path inventory remains lightweight source discovery for parametrization.

The post-Lower bridge is intentionally temporary. The next two stacked PRs replace its Link and Infer consumers, after which the retirement PR removes it with the legacy compiler.


tests/core/test_setup.py
```diff
@@ -150,14 +184,15 @@ def wic_yaml_filter_implementations_or_steps(yml: Yaml) -> bool:
...
-# NOTE: The CLI version of mypy and the VSCode version of mypy disagree on the
-# following line. The "type: ignore" comment is NOT unused.
-wic_strategy = wic_strategy.filter(wic_yaml_filter_blank_steps)
-wic_strategy = wic_strategy.filter(wic_yaml_filter_implementations_or_steps)
-wic_strategy = wic_strategy.filter(wic_yaml_filter_top_level_types)
+def wic_strategy(registry: CorpusRegistry) -> SearchStrategy:
+    """Construct the schema strategy from an explicitly loaded registry."""
+    stems = sophios.utils.flatten([list(paths) for paths in registry.workflows.values()])
+    schema = sophios.schemas.wic_schema.wic_main_schema(
+        registry.tools, stems, {}, hypothesis=True)
```
darkclad: **Passing `{}` here breaks `test_fuzzy_compile`.** `wic_main_schema(hypothesis=True)` inlines each tool and workflow schema from `schema_store` because hypothesis-jsonschema won't resolve an external `$ref` (see the NOTE at `wic_schema.py:408`). With an empty store, every step becomes `{'$ref': 'tools/<stem>.json'}`, and the first draw fails:

```
FAILED tests/core/test_fuzzy_compile.py::TestFuzzyCompile::test_fuzzy_compile
_RefResolutionError: hypothesis-jsonschema does not fetch remote references (uri='tools/toString.json')
```

That is from running the test at `473cc6c`. I also built the schema directly from this repo's 22 adapters: with `{}` it has 34 unresolved refs and draws nothing, and with the store `get_validator` fills it has none and draws normally. The base passed the store it had loaded or built. `test_fuzzy_compile` only runs in `fuzzy_compile_weekly.yml`, so this PR's checks won't catch it.

This also hides a real finding. At `92515ae` the same test generates and fails in 2m41s with `AttributeError: 'NoneType' object has no attribute 'items'` at `legacy_graph.py:89`, the `requirements.items()` from #427 (`63e4a17`), which a generated `requirements: null` reaches. That line is unchanged here, so the fuzz job will report it again once it can draw.

The fix is for `CorpusRegistry` to keep the `schema_store` that `load_test_registry` already builds, and for `wic_strategy` to pass `registry.schema_store`. Separately, `test_fuzzy_compile` now rebuilds the schema and calls `hj.from_schema` inside every example, through `st.data()`. Caching the strategy next to the registry would keep the weekly budget where it was.

src/sophios/ir/pipeline.py
```diff
@@ -0,0 +1,75 @@
...
+
+def legacy_after_lower(document: ResolvedDocument, graph: WorkflowGraph) -> Yaml:
+    """Temporarily adapt typed Lower output into the surviving Link/Infer path.
+
+    This is intentionally after Lower and visibly named legacy.  It is deleted
+    once Link and Infer consume ``WorkflowGraph`` directly.
+    """
+    if graph.name != document.name or len(graph.steps) != len(document.steps):
+        raise ValueError('the post-Lower bridge received a graph from another document')
+    raw = utils_cwl.desugar_into_canonical_normal_form(to_json(document.source))
```
darkclad, line 54: **The bridge rebuilds the step list from `document.source`, the parsed `Document`, so the `IDENTICAL` differentials only see Parse and `to_json`.** What Resolve found (process, `run_path`, interfaces) never reaches the compiled output. Neither does anything Lower built (ports, declarations, bindings, emission). For a flat workflow, `graph` is only checked for its name and step count. Resolve's output is used only for `python_script` ids, which the generator never draws, and for finding a child graph.

I checked this with a mutation. I made every step resolve to `registry.tools[0]`, the wrong process for almost every step. `test_resolution_is_identical_to_legacy_lookup`, `test_parse_front_door_is_identical_to_the_legacy_input`, `test_parse_resolve_lower_are_directly_typed` and `test_resolution_depends_only_on_the_registry` all still pass. Three pinned single-step examples fail, and so does `test_registry_order_cannot_change_resolution`, but only because reversing the snapshot changes which tool is first.

So "the typed path is required to remain `IDENTICAL` to legacy resolution" doesn't hold as written: the differential shows that Parse → `to_json` → desugar round-trips the corpus. That is worth having, but it is a Parse property. Two changes would make the claim true: compare Resolve's `run_path` and ports per step against what the legacy compiler looked up, in the generated property rather than in one pinned example, and build the bridged steps from the resolved/lowered values instead of `document.source`. At a minimum, the two tests that are named `identical` should say they test the source round-trip. They are also the same property twice (one filters `_scalar_literals_fit`, the other returns early), so one of them can go.

docs/sophios_language_reference.md
```diff
@@ -218,7 +214,7 @@ A step input is exactly one of these. There is no fifth form.
 |---|---|---|
 | Inline literal | `f: !ii empty.txt` | A literal value. Never an edge. |
 | Edge reference | `f: !* name` | Consumes an edge defined elsewhere |
-| Raw CWL reference | `f: !cwl step/out` | Opaque to Sophios; passed through unresolved. **Not yet usable — awaits the Spec 3 compiler migration** |
+| Raw CWL reference | `f: !cwl step/out` | Opaque to Sophios; passed through unresolved. |
```
darkclad: **The example on this row compiles to a dangling reference.** Authored step names don't survive into the CWL: `echo` is emitted as `<wf>__step__1__echo`. So `!cwl step/out` points at a step that doesn't exist. At `473cc6c`, with `echo` followed by `cat` bound as `file: !cwl echo/stdout`, the CLI emits:

```yaml
- id: stepout__step__1__echo
  ...
- id: stepout__step__2__cat
  in:
    file: echo/stdout
```

The only spelling that works is the mangled id. That id depends on the file name, and it changes when the workflow is embedded or inlined. In practice, `!cwl` now works for workflow inputs (`!cwl greeting` is fine) and not for step outputs. With the status table now saying "Usable in a compiled workflow: Yes", this example should be a workflow input. It should also say that step ids are renamed on emission, so a step-output reference has to use the emitted id or `!*`.

src/sophios/ir/resolve.py
```diff
@@ -0,0 +1,311 @@
...
+# pylint: disable-next=too-many-arguments,too-many-positional-arguments,too-many-locals
+def _resolve_process(step: Step, sidecar: WicSidecar | None, registry: RegistrySnapshot,
+                     version: str, trail: tuple[RegistryKey, ...],
+                     diagnostics: Diagnostics) -> ResolvedProcess | None:
+    sidecar_entries = dict(sidecar.entries) if sidecar is not None else {}
+    namespace = str(sidecar_entries.get('namespace', 'global'))
+    interpreted = dict(step.interpreted)
+    run = interpreted.get('run')
+    authored_name = _stem(run) if isinstance(run, str) else _stem(step.id)
+    generated = step.id == 'python_script'
```
darkclad, line 182: **Every `python_script` step fails to resolve.** The key becomes `python_script_<digest of the bindings>`. Nothing ever adds a tool under that name to a `RegistrySnapshot`: the legacy side generates the CWL in `ast.python_script_generate_cwl` under a `python_script_<uuid4>` name. So `front_end('steps:\n- id: python_script\n  in:\n    script: !ii a.py\n', ...)` returns no graph and `wic013 process global/python_script_005da6aa8cc3b88c is absent from the supplied registry`. That message points the user at their registry, not at a generated tool.

This doesn't block, because nothing in production calls `front_end` yet. But the description lists generated-tool identities as resolved, and `test_generated_process_identity_is_content_deterministic` only hashes the id. It never resolves a step. Either say in the description that generated tools are for a later PR, or add the step that registers the generated definition under this key, with one test that resolves a `python_script` step end to end.

darkclad (commented):
Review at `473cc6c`. This PR is stacked on #427, so I reviewed the four commits after `92515ae` (#427's head): `4a927b7`, `fb5d7dd`, `0d9bca3` and `473cc6c`.

Verified: the fast lane is **1151 passed**, against 1123 at `92515ae`. Both have the same single pre-existing Windows failure (`test_run_local_python_api_restores_environment_after_run`). `test_resolve.py`, `test_lower.py`, `test_emit.py` and `test_hermeticity.py` pass together (65 passed, 10 skipped). `mypy src/` is clean over 77 files, and `autopep8 --max-line-length 120` is clean. Pylint on the new `ir` modules, `test_resolve.py` and `test_setup.py` rates 9.98, and its one hit (`_resolve_version`) predates this PR.

The cleanup of test collection is right: importing `test_setup` no longer discovers plugins or writes under `autogenerated/`, and the probe test proves it. `!cwl` also really works now through the CLI: `message: !cwl greeting` compiles to `message: greeting` without `--allow_raw_cwl`.

**One thing blocks.** `wic_strategy` now builds its schema from an empty store, and `test_fuzzy_compile` can no longer draw a single example (first comment). It only runs in the weekly job, so this PR's CI won't show it. The second comment covers why the two `IDENTICAL` differentials don't test Resolve or Lower. The third covers the `!cwl step/out` form that the reference now marks as usable, and the fourth covers `python_script`, which the typed chain can't resolve yet.

One point that goes with the second comment: the title and the body say Parse is now the front door, and "the live typed front end". Nothing in `src/` calls `front_end` or `legacy_after_lower`. The CLI and the Python API still enter through `wic_loader` and the legacy compiler, so today the typed chain runs only in `test_resolve.py`. That fits a stacked series, but the description should say so. A reader of this PR will otherwise assume `sophios` parses with `sophios.lang` now.

tests/core/test_setup.py
vjaganat90: Fixed in `31b24507`. `CorpusRegistry` now carries the store it already built, and the strategy is built from that rather than `{}`. I reproduced your exact failure first and kept it as the mutation check: with `{}` the first draw raises `_RefResolutionError` on `tools/echo.json`; with the registry store it draws. The strategy is also built once now instead of once per `st.data()` example.

The `requirements: null` crash you found behind it is fixed on #427 and is in `master`, so unmasking this costs nothing.

src/sophios/ir/pipeline.py
vjaganat90: You are right, and the mutation reproduces: pointing every lookup at `self.tools[0]` left the property green.

`892f0429` does three of the four things. The surviving `IDENTICAL` test is renamed for what it measures — the source round-trip — and says in its docstring why a wrong resolver survives it. Its duplicate is gone. And a new generated property compares run path and both interfaces per step, recursively through child workflows, against an independent model of the registry; that one fails under your mutation.

I deliberately did not rebuild the bridge from the resolved values. It is deleted in the retirement PR, and comparing Resolve's output directly is the evidence the claim needed — rebuilding the bridge would move where the comparison happens without widening it. Say the word if you would rather have both.

One near-miss worth recording: my first version of that property compared `process.key.name` against the run path, which is self-consistent under your mutation and passed it. It compares the authored step id now.

docs/sophios_language_reference.md
vjaganat90: Fixed in `d9a3d4d7`. The row's example is a workflow input now, and the text says a step id does not survive emission — `<workflow>__step__<n>__<id>` — so a step output needs `!*` rather than `!cwl` with a name that changes when the workflow is embedded or inlined.

src/sophios/ir/resolve.py
vjaganat90: Described rather than fixed. The body now says generated identities are deterministic but not resolvable, and why: the legacy side registers the generated tool under `python_script_<uuid4>` while Resolve keys on a content digest, so no snapshot holds it.

I would rather not paper over that with a registration step here. It is the same seam as the rest of the loader — `read_ast_from_disk` generates the CWL, writes it under `autogenerated/`, and hands the compiler a tree — and it belongs with whatever replaces that front door, not bolted onto Resolve.

darkclad (approved):
LGTM.

Re-reviewed at `d9a3d4d`, now on `master` (`c5e6e08`) after #427 merged. The rebase changed only `ir/__init__.py` in `45ca128`, where `emit_visualization` dropped out with #427's consolidation. The other three commits are unchanged (`git range-diff` shows `=`), so the new work is `31b2450`, `892f042` and `d9a3d4d`. All four comments are addressed, three by code and one by the description.

Verified rather than read:

- **Fuzz strategy.** `test_fuzzy_compile` now passes at `d9a3d4d` (2 passed, 1m58s). At `473cc6c` it failed on its first draw.
- **Resolution check.** I re-ran my mutation (every step resolves to `registry.tools[0]`). `test_every_resolved_process_is_the_one_the_registry_holds` now fails, so the generated suite catches a wrong resolver, not just the pinned examples. Comparing the authored step id rather than `process.key.name` was the right catch. I don't need the bridge rebuilt as well: the direct comparison is the evidence the claim needed, and the bridge goes away with the retirement PR.
- **`!cwl` docs.** The row's example is a workflow input now, and the new paragraph says why a step output needs `!*`.
- **Description.** It now says that nothing in `src/` calls `front_end` yet, and that generated identities are deterministic but not resolvable.

**Suite.** Fast lane: **1153 passed**, against 1125 on `master`, with the same single Windows `run_local` failure on both. `mypy src/` is clean over 77 files, and `autopep8 --max-line-length 120` is clean. Pylint on `test_resolve.py` and `test_setup.py` is 9.96, and its one hit (`_resolve_version`) predates this PR.

**A follow-up that is not this PR's.** One of my clean runs of `test_parse_resolve_lower_are_directly_typed` failed inside the generator, not in anything this PR touches. `lang/render.py:240` (`_spell_scalar`) re-parses a constructed string literal with `yaml.safe_load`, which raises for strings such as `@`, `*`, `%`, `&`, `,`, `:`, `...`, `{` and `[`. It is the same on `master`: `render(Document(steps=(Step('echo', inputs=(('message', InlineLiteral('@', span)),)),)))` raises `ScannerError`. A parsed document is not affected, because it keeps its source spelling. But every `strat.workflows()` property can hit it whenever a literal like that is drawn, so it is worth its own fix: return `None` when the re-parse raises, so the caller falls back to the desugared form.

PR #429, opened 2026-09-18
Spec 3: link workflow graphs and discharge obligations

vjaganat90:
## Purpose

This is the third of five remaining Spec 3 changes, stacked on #428. It moves workflow composition and explicit-reference resolution out of the recursive legacy compiler and into a pure typed `Link` phase:

`Parse → Resolve → Lower → Link`

The phase receives only a `WorkflowGraph` and returns either a linked graph or structured diagnostics. Registry lookup, filesystem discovery, emission, and inference are outside this boundary.

## Semantic boundary

Link owns graph composition and authored wiring. It recursively attaches resolved child graphs to workflow-call descriptors, projects call inputs onto the child's concrete input ports, redirects workflow-call outputs to their concrete producers, and places cross-scope edges at their lowest common ancestor. Structured step and port identities remain injective across the composed graph, so namespace spelling is not used as identity.

Every `DeferredObligation` must be discharged by a visible explicit definition. An obligation that reaches the root unresolved is `wic025`; it is not promoted into an accidental workflow input. Authored call metadata also cannot recreate a formal removed from the resolved process interface.

Reference compatibility remains a Sophios language decision. Link applies the versioned `reference_relation` to local aliases, workflow-input references, workflow-call bindings, and cross-scope obligations, including producing and consuming scatter lifts. It rejects only a proven `DISJOINT` relation as `wic023`; `UNKNOWN` is deliberately preserved for final CWL conformance. Production code does not delegate this judgment to `cwltool.check_types()`.

## Stack role

This closes P52–P54 at `UP_TO_EMBEDDING`. The only tolerated differential is the run-path representation introduced by resolved process embedding; renaming is not accepted. The temporary handoff writes already-linked local sources into the legacy tail, which remains solely as the differential oracle and temporary Infer implementation. The next PR replaces that final semantic dependency with typed fixed-point inference.

Depends on #428. All PRs remain based on `master`; commit ancestry defines their order.


src/sophios/ir/link.py
```diff
@@ -0,0 +1,258 @@
...
+
+
+def link(graph: WorkflowGraph) -> Linked:
+    """Compose ``graph`` and discharge every obligation it can prove."""
+    diagnostics = Diagnostics()
+    attached = _attach_children(graph)
+    normalized = _normalize_explicit_edges(attached, attached, diagnostics)
+    attached = _attach_children(normalized)
+    _check_workflow_inputs(attached, diagnostics)
+    definitions = _definitions(attached)
```
darkclad, line 39: **A cross-scope `!*` resolves to a definition that comes after it, while the same order inside one scope is rejected.** `_definitions` collects every `!&` in the composed tree with no regard to position, so an obligation in a call is discharged by a root step authored after that call. Lower applies `defined_so_far` to a local `!*`, so the flat form of the same program is `wic025`. Checked at `c5975f2` with `count` in `child.wic` as `in: {file: !* shared}`:

| root | typed (this PR) | legacy |
|---|---|---|
| `mk_file` (`!& shared`), then `child.wic` | links `mk_file -> count` | links |
| `child.wic`, then `mk_file` (`!& shared`) | **links** `mk_file -> count` | promotes `root__step__1__child.wic___child__step__1__count___file` to a root input |
| flat: `count` (`!* shared`), then `mk_file` | `wic025` at `front_end` | n/a |

The second row is the promotion this PR sets out to remove, so changing it is intended. The PR description says an obligation left at the root is `wic025`, though, and that is what this row should report if order matters. The alternative is that order does not matter across a call, and then the flat rule and this one disagree about the same language construct. Either answer works, but it should be one rule, stated once and pinned by a test. `test_link_is_equivalent_up_to_embedding` did not catch this because legacy and typed disagree here, so the generator evidently never produces a call that precedes its producer.

src/sophios/ir/types.py
```diff
@@ -310,14 +436,30 @@ def _references(self) -> tuple[tuple[str, PortId], ...]:
     @property
     def edges(self) -> tuple[Edge, ...]:
         """Every resolved edge, derived from the bindings that produced them."""
-        return tuple(b.resolution for s in self.steps for b in s.bindings
-                     if isinstance(b.resolution, Edge))
+        local = tuple(b.resolution for s in self.steps for b in s.bindings
+                      if isinstance(b.resolution, Edge))
+        return local + self.composition_edges + tuple(
```
darkclad: **`edges` is now recursive, which changes what `emit_visualization` draws.** `emit.py:60` builds each level's `edges` from `graph.edges` and then recurses into `children`, so a child's edges now appear at its own level and again at every ancestor. At the parent level the endpoints are child step indices that do not name any of that level's `nodes`. Probe: a root whose only step is `child.wic`, where the child is `mk_file -> count`:

| | `473cc6c` | `c5975f2` |
|---|---|---|
| root `edges` | `()` | `((1, 'file', 2, 'file'),)` |
| child `edges` | `((1, 'file', 2, 'file'),)` | `((1, 'file', 2, 'file'),)` |

The root has one node, so index 2 does not exist at that level. `emit_visualization` has no test with a child, which is why the suite stays green. `link()` itself only needs the recursive view through `_place_edges`. One fix is a separate property for the tree-wide view (`all_edges`, next to `all_steps`), with `edges` left as local bindings plus this level's `composition_edges`. `legacy_after_link` already filters `graph.edges` back down to one namespace, so it would get simpler too.

src/sophios/ir/lower.py
```diff
@@ -100,6 +118,203 @@ def lower(document: Document, namespace: Namespace | None = None) -> Lowered:
...
+def _resolved_step_node(workflow_name: str, identity: StepId, resolved: ResolvedStep,
+                        defined_so_far: dict[str, PortId], defined_anywhere: dict[str, PortId],
+                        diagnostics: Diagnostics) -> StepNode:
+    source = resolved.source
+    declared_inputs = {port.name: port.declaration for port in resolved.process.inputs}
+    declared_outputs = {port.name: port.declaration for port in resolved.process.outputs}
+    for name, _ in source.inputs:
+        if name not in declared_inputs:
+            diagnostics.error(
+                SophiosErrorCode.SUBWORKFLOW_INVALID,
```
darkclad: **`wic013 SUBWORKFLOW_INVALID` now also fires on plain tool steps.** Rejecting the binding is right: legacy raises `ValueError: Provided input ghost not found in interface of step mk_file` for the same input. But this path handles every resolved step, so `front_end('steps:
- id: mk_file
  in: {name: !ii a, ghost: !ii x}
')` now reports a subworkflow error for a CLT. #428 accepted that step. Either pick a code that fits both cases or branch on `resolved.process.child`. Separately, `test_unknown_call_argument_cannot_restore_a_deleted_formal` asserts only `has_errors`, so whichever code is chosen should be pinned there.

darkclad (commented):
Review at `c5975f2` against #428 (`473cc6c`), not `master`. This PR's own change is three commits: `99ae5c8`, `7bf5d39` and `c5975f2`, 631 lines over 10 files. #427 and #428 are reviewed on their own PRs.

Verified: the fast lane is **1171 passed**, against 1155 at the base. Both have the same single pre-existing failure (`test_run_local_python_api_restores_environment_after_run`, Windows). All 11 `test_link.py` tests pass, including both `skip_pypi_ci` properties (`UP_TO_EMBEDDING` differential and composed-namespace injectivity). `mypy src/` is clean, and `autopep8 --max-line-length 120` is clean on the touched files. The one pylint note is `too-many-locals` (16/15) on `legacy_after_link`.

The design is right. Link takes only a `WorkflowGraph`, owns composition, places edges at the lowest common ancestor, and rejects only a proven `DISJOINT`. `UNKNOWN` goes on to CWL. Ending the silent promotion of an unresolved obligation to a root input is the most useful change in the stack so far.

**One thing needs an answer before approval.** Link ignores authored order when it discharges a cross-scope obligation, but Lower enforces it within a scope. So the same program is `wic025` when flat and links cleanly once it is split across a call. The first comment has the probe. The second comment covers a side effect of making `WorkflowGraph.edges` recursive: `emit_visualization` (from #427) now lists every child edge again at each ancestor level. The third is a smaller point about which code the new undeclared-binding check reports.

src/sophios/ir/link.py
vjaganat90: Order matters, and the reference already said so — §4.1.2: "a reference is resolved against the definitions seen so far" and "a document included as a subworkflow may reference an edge its includer has **already defined**". Link was the one place not applying it. Fixed in `358a2a4f`: an obligation is discharged only by a definition whose position precedes it, comparing the step indices on the path down to each, so an enclosing step precedes everything inside a later call and follows everything inside an earlier one. Your row 2 is `wic025` now, which is what the description claimed. Both rows of your table are pinned.

I did not touch the implementation-status table. Nothing in `src/` calls `link` yet either, so "usable in a compiled workflow" stays **Partial** until the retirement PR — same situation as `front_end` on #428.

src/sophios/ir/types.py
vjaganat90: Fixed in `ddd82cb1`, and the symptom is gone twice over — `emit_visualization` was removed on #427, so nothing draws it now.

`edges` is local: its own bindings plus what Link placed at that level. The recursion existed only for `legacy_after_link`, which filtered it straight back down to one namespace, which is exactly why nothing failed.

I did not add `all_edges`. With `edges` local the whole suite passes, so nothing actually wants the tree-wide view, and an unused property is what let the empty-edge-list claim stand on #427 in the first place. Easy to add the moment something needs it.

src/sophios/ir/lower.py
vjaganat90: Fixed in `0eb9f7b8` with a new code rather than a branch. `wic028 UNDECLARED_PORT`: one code across both positions, because naming an input the process lacks and naming an output it lacks are one mistake, and one code across the process kinds, because a CommandLineTool has an interface just as a subworkflow does. Branching on `resolved.process.child` would have needed a second code for the tool case anyway, and the two cases are not different mistakes.

Its registered provocation uses a tool step — the case the old code did not fit — and `test_unknown_call_argument_cannot_restore_a_deleted_formal` pins the code instead of just `has_errors`.

darkclad (approved):
LGTM.

Re-reviewed at `5ed8bf9`, now stacked on #428 (`d9a3d4d`) after #427 merged. `47420f6`, `0d755a7` and `6fad970` are unchanged rebases of the three commits I reviewed (`git range-diff` shows `=`), so the new work is `0eb9f7b`, `358a2a4`, `ddd82cb` and `5ed8bf9`. All three comments are addressed.

Verified with runs, not just by reading:

```
case                                   expected   got
root def, then call                    links      links
call, then root def                    wic025     wic025
flat use before def                    wic025     wic025
sibling producer call, consumer call   links      links
sibling consumer call, producer call   wic025     wic025
child def, then grandchild call        links      links
grandchild call, then child def        wic025     wic025
tool step binds an undeclared input    wic028     wic028
```

- **Order.** One rule now holds on both sides of a call, including sibling calls and a def inside an intermediate call. Comparing path indices lexicographically is the right encoding of §4.1.2.
- **`edges`.** It is local again. The composed `mk_file -> count` edge sits at the root, the child owns none, and the bridge still passes the `UP_TO_EMBEDDING` differential. Leaving out `all_edges` until something needs it is fine.
- **`wic028`.** A tool step now reports `UNDECLARED_PORT`, the provocation uses a tool step, and the call-argument test pins the code. `5ed8bf9` puts it in the fuzz job's tolerated set next to `wic013`, which it split from, and `test_fuzzy_compile.py` passes.

**Suite.** Fast lane: **1175 passed**, against 1157 on #428, with the same single Windows `run_local` failure on both. All 13 `test_link.py` tests pass, including both `skip_pypi_ci` properties. `mypy src/` is clean over 78 files, and `autopep8 --max-line-length 120` is clean on the touched files.

PR #430, opened 2026-09-18
fix: create generated CWL output directory

vjaganat90:
## Stack position

First descendant of merged #426 and the bug-fix base for #427 through #431. All PRs remain based on `master`; commit ancestry supplies the stack order.

## Problem

The scheduled Benchmarks job starts from a clean checkout and `sophios --generate_schemas` reaches `python_script_generate_cwl()` before any caller creates `autogenerated/`. The writer then crashes with `FileNotFoundError` while opening its generated CWL path.

Failed job: https://github.com/vjaganat90/sophios/actions/runs/35297287446/job/105452344572

## Fix

Make the generated-CWL writer own creation of its parent directory. Keep the registered tool path as the existing string representation.

## Targeted validation

- `sophios --generate_schemas` from a clean output state
- Direct Python-script CWL generation with `autogenerated/` absent; verified file creation, tool registration, and step rewrite
- `pre-commit run --files src/sophios/ast.py`


darkclad (approved):
LGTM

PR #431, opened 2026-09-18
Spec 3: infer graph edges to a fixed point

vjaganat90:
## Purpose

This is the fourth of five remaining Spec 3 changes, stacked on #429. It extracts the final semantic search from the recursive compiler into a pure typed phase:

`Parse → Resolve → Lower → Link → Infer`

Infer receives a linked `WorkflowGraph`, an immutable `InferencePolicy`, and an immutable converter catalog. It returns a new graph or structured diagnostics. It does not parse source, resolve registries, traverse the filesystem, emit CWL, or mutate process-wide policy.

## Candidate semantics

The phase preserves the established inferred-edge contract rather than borrowing the stricter explicit-reference judge. Candidate selection retains backward step order, reverse output declaration order, the existing union behavior of `types_match()`, exact format handling, `_log_` exclusion, optional naming conventions, per-output break rules, defaults by presence including falsy values, and producing/consuming scatter lifts. An unmatched required input becomes an explicit workflow-boundary input; nested boundary inputs propagate through workflow calls without re-running an earlier phase.

This separation is intentional: Link has rejection authority and diagnoses only proven-disjoint authored references. Infer is a convenience search whose historical permissiveness is part of workflow behavior. Conflating the two would either reject valid-but-opaque references or silently change which producer inference chooses.

## Fixed point and insertion

Speculative converter insertion is now graph-local. A deterministic catalog exposes only the existing `insert_steps_automatically_*` whitelist. Infer inserts the lexicographically first eligible converter, then repeats only Infer over the already parsed, resolved, lowered, and linked graph. Parse, Resolve, Lower, and Link do not repeat. The iteration limit is policy data and exhaustion is exactly `wic022`.

Both compiler inference-rule mappings and naming-convention rewrites now travel through explicit options. The CLI and `cwl_subinterpreter` no longer write compiler or inference module globals, so independent compilations cannot contaminate one another.

## Equivalence and performance

P55 and P56 hold at `IDENTICAL`, including canonical mapping order. The generated differential is paired with an independent test-side candidate model; planted cases cover ambiguity, reverse ordering, falsy defaults, scatter, two converter insertions, and limit exhaustion. The temporary post-Infer bridge supplies already-selected sources to the legacy finalizer, leaving that path as an emission oracle rather than a second candidate judge.

The existing synthetic benchmark medians on the same runner remained in the same sub-millisecond/noise band: speculative insertion 0.000727s → 0.000853s, deep nesting 0.001187s → 0.001362s, and wide inference 0.012479s → 0.012946s. No threshold or optimization claim is introduced.

Depends on #429. All PRs remain based on `master`; commit ancestry defines their order.


darkclad (commented):
Review at `711c097` against `master` (`4ae925b`), now that #429 has merged. The change is two commits, `fb66a1c` and `711c097`, 834 lines over 18 files.

Verified: the fast lane is **1187 passed**, against 1171 at the base, with the same single pre-existing Windows failure on both (`test_run_local_python_api_restores_environment_after_run`). The PR-specific set (`test_infer_phase.py`, `test_link.py`, `test_settings_boundary.py`, `test_hermeticity.py`, `test_manufactured_documents.py`) is 228 passed, 8 skipped. `mypy src/` is clean over 79 files, pylint is 10.00 on `infer.py`, `pipeline.py` and `lower.py`, and autopep8 is clean on the touched files.

The structure is right. Infer is pure, the policy and catalog are immutable, and only Infer repeats after an insertion. Moving `inference_rules` and `renaming_conventions` out of module globals into `CompilerOptions` removes real leakage between compilations. The Python API and REST paths got `{}` and `[]` from those globals before and still do, so nothing changes for them.

**Two divergences from legacy need fixing before approval.** Both contradict the claim that candidate selection is preserved, and the current differential reaches neither. The inline comments have the probes: a workflow call contributes no candidates, and converter insertion ignores `break`.

src/sophios/ir/infer.py
```diff
@@ -0,0 +1,405 @@
...
+
+
+def _renumber_emission(name: str, index: int, step: StepNode) -> StepNode:
+    if step.emission is None:
+        return step
+    return replace(step, emission=replace(
+        step.emission, id=f'{name}__step__{index}__{step.id.name}'))
+
+
+def _propagate_child_inputs(graph: WorkflowGraph) -> WorkflowGraph:
```
darkclad: **A workflow call contributes no candidates, so an input after a call never infers from it.** The typed graph gives a call step inputs (this function adds the child's `workflow_inputs`) but no outputs, and nothing does the same for the child's outputs. `_candidate` and `_insertion_candidate` both read `producer.outputs`, so neither can see a call. Checked at `711c097` with `sub.wic` holding `mk_file` (`name: !ii x`):

| root | legacy | typed Infer + `legacy_after_infer` |
|---|---|---|
| `sub.wic`, then `count` | `count.file` <- `oracle__step__1__sub.wic/sub__step__1__mk_file___file` | `count.file` becomes root input `oracle__step__2__count___file` |
| `sub.wic` (child `mk_2`), then `use_2`, insertion on | inserts `insert_steps_automatically_conv_2` between them | inserts nothing; `use_2.file` becomes a root input |

The first row fails `assert_compilations_equivalent(..., Strength.IDENTICAL)` at `.steps[1].in.file`. `test_infer_is_identical_to_the_legacy_candidate_search` cannot catch it: in 1,500 examples from `strat.workflows()`, no tool step ever came after a `.wic` call. The differential needs a planted case with a call followed by a consumer. If the generator stays as it is, the `BLIND SPOTS` note in `test_infer_phase.py` should also say so.

src/sophios/ir/infer.py
```diff
@@ -0,0 +1,405 @@
...
+             if port.id.port.split('___')[-1].replace('output_', '') == wanted]
+    return named[0] if named else matches[0]
+
+
+def _insertion_candidate(steps: list[StepNode], position: int, sink: Port,
+                         catalog: InsertionCatalog) -> Insertion | None:
+    sink_formats = _formats(sink.declaration)
+    if not sink_formats:
+        return None
+    previous_formats = {
```
darkclad, line 223: **Insertion looks past a `break` rule, where legacy does not.** `previous_formats` takes every output of every earlier step. Legacy builds `out_formats` from `attempted_matches_all`, which stops at the step where `break` fires. Probe: `[mk_1, mk_2, use_1]` from `_insertion_registry()` with `wic: {steps: {'(2, mk_2)': {wic: {inference: {file: break}}}}}`, insertion on. Legacy emits `mk_1, mk_2, use_1`, with `use_1.file` as a root input. Typed Infer inserts `insert_steps_automatically_conv_1` before `use_1`. `_candidate` already tracks where the break stops the search, so the insertion search could reuse the outputs it actually visited. A planted case next to `test_converter_insertion_reaches_the_same_fixed_point` would pin it.

src/sophios/ir/infer.py
vjaganat90: Fixed in commit 06252df5. Infer now materializes a workflow call’s child output interface on the call step before candidate selection, including the child’s authored outputs and namespaced step outputs with effective scatter types. Both ordinary candidate search and converter search can therefore consume outputs produced inside a preceding call. The regression coverage includes a child feeding a later tool directly and a child feeding an inserted converter.

src/sophios/ir/infer.py
vjaganat90: Fixed in commit 06252df5. Candidate search now returns the exact output ports it attempted before matching or stopping at a break rule; converter selection consumes only that frontier instead of rescanning every earlier step. The planted mk_1 → mk_2 (break) → use_1 case now leaves use_1.file as a workflow input, matching legacy behavior.

vjaganat90:
Addressed both candidate-selection divergences in commit 06252df5. Workflow-call steps now expose the complete child output interface to parent inference, including effective scatter types, so direct edges and converter insertion can consume child-produced outputs. Converter selection now receives the exact output frontier visited by candidate search, preserving local break semantics instead of looking behind the break. Planted regressions cover direct workflow-call inference, workflow-call converter insertion, and break-bounded converter search.

src/sophios/ir/infer.py
```diff
@@ -0,0 +1,431 @@
...
+                inferred_edges.append(Edge(source, port.id, port.span))
+                bound.add(port.id)
+                continue
+            insertion = _insertion_candidate(attempted, port, catalog)
+            if policy.insert_steps_automatically and insertion is not None:
+                return _insert(graph, position, insertion, policy), True
+            current_step = steps[position]
+            input_name = _input_name(current_step, port)
+            if input_name not in {item.name for item in workflow_inputs}:
+                declaration = port.declaration or port_declaration(port.type.declared)
```
darkclad, line 155: **A promoted boundary input keeps a scalar `format`; legacy always emits a list.** `inference.py:246-248` does `in_formats = declared_formats(...)` and then `in_dict['format'] = in_formats`, and `:386` puts that dict straight into `inputs_workflow`, so legacy normalizes one authored format to `['edam:format_1']`. This path carries `port.declaration` through unchanged, so the same input stays `'edam:format_1'`. Checked at `06252df`, one step whose only required input is an unmatched `File`:

| input declares | legacy | typed Infer + `legacy_after_infer` |
|---|---|---|
| `format: edam:format_1` | `format: ['edam:format_1']` | `format: 'edam:format_1'` |
| `format: [edam:format_1, edam:format_2]` | same list | same list |
| no format | no format | no format |

So it is only the scalar spelling. `assert_compilations_equivalent(..., Strength.IDENTICAL)` fails at `.inputs.<name>.format` for the first row, which is the contract the rest of this file asserts. Nothing in the suite reaches it: `SYNTHETIC_TOOLS` declares no formats, and in the converter tests every formatted input is either matched or bound by an insertion. In the real corpus a formatted `File` input with no producer is ordinary, so this would change emitted CWL once Infer replaces the legacy search. A planted case with an unmatched formatted input would pin whichever spelling you choose.

darkclad (commented):
Re-reviewed at `06252df`. `fb66a1c` and `711c097` are unchanged (`git range-diff` shows `=`), so the new work is `06252df` alone. Both comments are addressed.

Verified with runs, not by reading:

```
case                                        legacy vs typed
nested call, then consumer                  IDENTICAL
call with two producers, then consumer      IDENTICAL
root producer, call producer, consumer      IDENTICAL
consumer before the call                    IDENTICAL
scattered producer inside a call            IDENTICAL
two consumers after one call                IDENTICAL
nested call feeds an inserted converter     IDENTICAL
insertion inside a child, consumer at root  IDENTICAL
break case, insertion on                    IDENTICAL
break case, insertion off                   IDENTICAL
```

- **Workflow-call candidates.** `_propagate_child_interface` exports the child's authored outputs and its namespaced step outputs, so a call is a producer on both paths. The namespacing matches legacy through two levels of nesting (`mid__step__1__inner.wic___inner__step__1__mk_file___file`), and ambiguity inside a child still resolves to the most recent step.
- **Break.** Threading the attempted frontier out of `_candidate` is the right shape: one search decides what was visited, and insertion cannot see past it. The planted `mk_1 -> mk_2 (break) -> use_1` case leaves `use_1.file` a workflow input on both sides.

**Suite.** Fast lane: **1190 passed**, against 1171 on `master`, with the same single pre-existing Windows `run_local` failure on both. The Spec 3 phase files plus `test_settings_boundary.py`, `test_hermeticity.py` and `test_manufactured_documents.py` are 275 passed, 10 skipped. `mypy src/` is clean over 79 files, pylint is 10.00 on `infer.py`, and autopep8 is clean on the touched files.

One smaller divergence is left, in the inline comment: a promoted boundary input keeps a scalar `format` where legacy emits a one-element list. It predates `06252df` and does not change which edges are chosen.

src/sophios/ir/infer.py
darkclad: **Fixed** `_propagate_child_interface` gives a call step the child's output interface, and I checked it at `06252df` beyond the planted cases: nested calls name ports the way legacy does two levels down, a child with two producers still resolves to the most recent, a scattered producer inside a call behaves, and a converter fed through a nested call inserts in the same place. All IDENTICAL against legacy.

src/sophios/ir/infer.py
darkclad: **Fixed** Returning the attempted frontier from `_candidate` keeps one search in charge of what insertion may see. Verified at `06252df` that `mk_1 -> mk_2 (break) -> use_1` leaves `use_1.file` a workflow input, with insertion both on and off, matching legacy exactly.

src/sophios/ir/infer.py
vjaganat90: Fixed in commit 3179d09d. When Infer synthesizes a workflow-boundary input, a literal scalar format IRI is canonicalized to the one-element list spelling used by legacy inference; existing lists and absent formats remain unchanged. CWL v1.2 format expressions are opaque rather than literal sets, so strings containing either expression form remain scalar through both the graph and the temporary handoff. Regressions pin IDENTICAL output for the literal IRI and preserve both parameter-reference and JavaScript expression forms.

vjaganat90:
Narrowed the boundary-format fix in commit 3179d09d to the CWL v1.2 distinction: only literal scalar format IRIs canonicalize to singleton lists. Existing arrays, absent values, and both CWL expression forms remain unchanged and opaque.

src/sophios/ir/infer.py
```diff
@@ -0,0 +1,440 @@
...
+
+
+def _formats(declaration: PortDeclaration | None) -> tuple[Any, ...]:
+    if declaration is None or not declaration.has_format:
+        return ()
+    value = declaration.format
+    return tuple(value) if isinstance(value, list) else (value,)
+
+
+def _canonical_boundary_format(value: Any) -> Any:
```
darkclad: **The expression carve-out is a deliberate divergence from legacy, and nothing in the suite says so.** `declared_formats` (`inference.py:31-43`) wraps *any* string, expression included, so legacy's promoted boundary input gets the singleton list either way. Checked at `3179d09`, one step whose only required input is an unmatched `File`:

| input declares | legacy | typed Infer + `legacy_after_infer` |
|---|---|---|
| `edam:format_1` | `['edam:format_1']` | `['edam:format_1']` |
| `[edam:format_1, edam:format_2]` | same list | same list |
| `['edam:format_1']` | same list | same list |
| no format | absent | absent |
| `$(inputs.source.format)` | `['$(inputs.source.format)']` | `'$(inputs.source.format)'` |
| `${ return inputs.source.format; }` | singleton list | scalar |
| `edam:format_$(x)` | singleton list | scalar |

I agree with the reading: in CWL v1.2 an element of a `format` array is an IRI, so wrapping an expression in one states something the author did not write, and legacy is the side that is wrong. The issue is only that `test_promoted_input_preserves_a_cwl_format_expression` asserts the typed spelling alone and never compares against legacy, so the file now contains one case that is intentionally *not* IDENTICAL without saying so. A sentence in that test's docstring, or in the `BLIND SPOTS` note, would keep the next reader from reading the omission as an oversight. The last row is worth a thought too: the marker test is a substring check, so a literal IRI that merely contains `$(` is treated as opaque.

Minor, while you are there: the same `('$(', '${')` test now exists in `_canonical_boundary_format` and again in the `wic_inferred_input` arm of `compile_workflow_once` (`compiler.py:1028-1031`). The bridge is temporary, so this is fine, but a shared helper would stop the two from drifting apart while it lives.

darkclad (approved):
LGTM at `3179d09`.

`fb66a1c`, `711c097` and `06252df` came through unchanged (`git range-diff` shows `=`), so this round is `3179d09` alone. The scalar-IRI case I raised is fixed, and all ten cases from round 2 still hold:

```
case                                        legacy vs typed
nested call, then consumer                  IDENTICAL
call with two producers, then consumer      IDENTICAL
root producer, call producer, consumer      IDENTICAL
consumer before the call                    IDENTICAL
scattered producer inside a call            IDENTICAL
two consumers after one call                IDENTICAL
nested call feeds an inserted converter     IDENTICAL
insertion inside a child, consumer at root  IDENTICAL
break case, insertion on                    IDENTICAL
break case, insertion off                   IDENTICAL
```

Boundary formats, checked shape by shape against legacy: a literal IRI, a two-element list, a one-element list and an absent format all match now. The three expression shapes do not, which is deliberate and which I think is the better reading of CWL v1.2 — the inline comment asks only that the divergence be recorded somewhere a future reader will find it, since this file's stated contract is byte-identical emission. Not blocking.

**Suite.** Fast lane: **1193 passed**, against 1171 on `master`, with the same single pre-existing Windows `run_local` failure on both. `mypy src/` is clean over 79 files, pylint is 10.00 on `infer.py` and 9.82 on `compiler.py` (9.74 on `master`; every remaining note is a pre-existing structural one), and autopep8 is clean on the touched files.

Three rounds in, the phase boundary holds: Infer is pure, the policy and catalog are immutable, and a call now contributes its child's interface on both the candidate and the insertion path.

src/sophios/ir/infer.py
darkclad: **Fixed** for the literal IRI: a scalar, a one-element list, a two-element list and an absent format all match legacy at `3179d09`. The expression carve-out that came with it does diverge from legacy, which wraps every string through `declared_formats`; I think your reading of CWL v1.2 is right, so I have left it as a note rather than a request, in a separate comment.

src/sophios/ir/infer.py
vjaganat90: Recorded in commit 4d8d9867 in the two complementary regression docstrings. The literal-IRI test now states why a Sophios-synthesized boundary input may choose the singleton-list canonical form even though authored spelling remains preserved. The expression test states the CWL v1.2 union-arm distinction, identifies legacy wrapping as the deliberate divergence, documents the marker-based opacity rule, and explains why that case intentionally does not assert IDENTICAL legacy agreement. I left the duplicated marker check in the temporary handoff unchanged rather than widening this documentation-only closeout.

PR #432, opened 2026-09-18
Spec 3: retire the legacy compiler

vjaganat90:
## Summary

This closes the compiler migration by making the typed pipeline the only internal implementation:

`Parse → Resolve → Lower → Link → Infer → Emit`

`compile_document` now returns the final immutable `WorkflowGraph` and its graph-derived artifact tree. The CLI, Python API, REST surface, schema generation, runtime-input normalization, post-processing, visualization, and the narrow `cwl_subinterpreter` entry point consume that result directly.

There is no legacy fallback or compatibility projection. `compile_workflow`, `CompilerInfo`, `RoseTree`, `NodeData`, `EnvData`, the mutable four-dictionary environment, the old inference loop, the source-level inliner, the legacy emitter, and their newly unreachable utilities are deleted. Lower also accepts only `ResolvedDocument`; the temporary syntax-only lowering path is gone.

## Compatibility boundary

Compatibility is intentionally limited to the documented Python API and CLI behavior. Public callers still receive `CompiledWorkflow`, existing CLI flags and entry points remain available, `Tools`, `wic_types.StepId`, and `Tool` remain accepted at the public boundary, and `ir.StepId` does not leak through public signatures.

The old internal call signature and result tree are not preserved. Internal consumers have one representation and one compiler path.

The inline-workflow CLI behavior now embeds already-linked child artifacts in emitted CWL. It does not rerun a separate source-rewrite implementation of workflow composition. `cwl_subinterpreter` remains a narrow adapter and does not regain authority over compiler state.

## Semantic closure

The end-to-end contract is behavioral equivalence at `UP_TO_EMBEDDING`, not byte-identical CWL. Generated paths and whether a linked child is external or embedded may differ; workflow structure, interfaces, types, bindings, requirements, opaque payloads, and observable dataflow may not.

The repository enforces both sides of retirement:

- phase composition agrees with the live compiler at `UP_TO_EMBEDDING`;
- semantic phases exchange typed immutable values and do not read mutable cross-phase state;
- retired modules, entry points, result shapes, and imports cannot reappear unnoticed;
- emitted artifact trees are the sole source for CWL, job inputs, child artifacts, and graph projections;
- every existing Spec 2 property runs through the typed compiler path.

The change removes 4,368 lines and adds 1,824, a net deletion of 2,544 lines.


src/sophios/ir/complete.py
```diff
@@ -0,0 +1,479 @@
...
+        for sink in sinks:
+            target_step, target_port = _direct_sink(graph, sink)
+            position = positions.get(target_step)
+            if position is None:
+                continue
+            step = steps[position]
+            if step.emission is None:
+                continue
+            inputs = dict(step.emission.inputs)
+            inputs.setdefault(target_port, {'source': name})
```
darkclad: `setdefault` cannot win this slot, because `lower._input_surface` has already filled it. For an `EdgeRef` it seeds `emission.inputs[port] = {Key.ALIAS: name}`, and `_materialize_bindings`'s `case EdgeRef(): pass` (line 138) deliberately leaves that in place for Link to overwrite. When the alias is discharged by `link._expose_cross_scope_inputs` rather than by a concrete `Edge`, nothing overwrites it: the binding's resolution is a `DeferredObligation`, so the `bind(..., explicit=True)` loop never touches it and only this `input_mapping` loop applies -- with `setdefault`, which finds the raw alias already sitting there.

The result is the authored sidecar construct emitted verbatim as a CWL step input, with no diagnostic, and the boundary input this same pass declared left unread.

Two `.wic` files under one root, anchor in the first and alias in the second:

```yaml
# part0.wic          # part1.wic            # root.wic
steps:               steps:                 steps:
- id: count          - id: poly             - id: part0.wic
  out:                 in:                  - id: part1.wic
  - n: !& a0               value: !* a0
                     - id: join
```

`part1`'s emitted CWL on this branch:

```yaml
steps:
- id: part1__step__1__poly
  in:
    value:
      wic_alias: a0             # on ece5181: part1__step__1__poly___value
inputs:
  part1__step__1__poly___value: # declared, and now read by nobody
    type: [int, string]
```

The parent wires `part1__step__1__poly___value` correctly, so only the child document is wrong -- which is the one cwltool will reject.

`tests/core/test_equivalences.py::test_a_meaning_preserving_rewrite_preserves_meaning[split]` catches this; it is green on `ece5181` and red here, on `{'steps': [{'id': 'count', 'out': [{'n': {'wic_anchor': 'a0'}}]}, {'id': 'poly', 'in': {'value': {'wic_alias': 'a0'}}}, {'id': 'join'}]}` with `split(((0,), (1, 2, 3, 4)))`. It reproduces through `compile_source` on real files too, so it is not an artifact of the in-memory door.

Worth noting that `infer._infer_local`'s new `bound.update(... if port.id.port in dict(step.emission.inputs))` is what keeps this quiet: the port counts as bound precisely because the stale alias is occupying `emission.inputs`.

src/sophios/ir/complete.py
```diff
@@ -0,0 +1,479 @@
...
+    output_mapping = list(graph.output_mapping)
+    authored = {port.name for port in outputs}
+    for step in graph.steps:
+        if step.emission is None:
+            continue
+        for port in step.outputs:
+            name = f'{step.emission.id}___{port.id.port}'
+            if name in authored:
+                continue
+            declaration = _output_declaration(port, step.emission.scatter)
```
darkclad: This arm is not idempotent, and `complete()` documents itself as idempotent (line 39) while `_compile_front` calls it three times -- before Link, after Link, after Infer.

Each pass takes `PurePath(target).stem` of the previous pass's output and prefixes again, so the namespace lands three times:

```
pass 1: sub.cwl                -> step___sub.cwl
pass 2: step___sub.cwl         -> step___step___sub.cwl
pass 3: step___step___sub.cwl  -> step___step___step___sub.cwl
```

The sibling arms are stable under repetition -- `f'{emission.id}/{leaf}'` and `PurePath('..') / leaf` both reproduce themselves -- so this is the only one that drifts.

On `docs/tutorials/multistep3.wic` with `relative_run_path=False`:

```
run: multistep3__step__2__append_twice.wic___multistep3__step__2__append_twice.wic___multistep3__step__2__append_twice.wic___append_twice.cwl
```

against `multistep3__step__2__append_twice.wic___append_twice.cwl` on `ece5181`. `input_output._write_artifacts_to_disk` names the file by one prefix, so after `write_artifacts_to_disk(..., relative_run_path=False)` the `run:` resolves to nothing -- I checked the written directory, the file is there under the single-prefix name and the reference misses it.

`cwl_subinterpreter.rerun_cwltool` compiles with `relative_run_path=False` (line 126) and writes with it, so this sits on a runtime path, not only in `test_cwl_embedding_independence` -- and that test deletes every `run` key before comparing, which is why it stays green.

src/sophios/ir/frontdoor.py
```diff
@@ -0,0 +1,223 @@
...
+            generated[StepId(generated_process_id(step), namespace)] = \
+                _generated_tool(step, script_dir)
+        elif step.id.endswith('.wic'):
+            # An undiscovered workflow is left unregistered rather than raising:
+            # Resolve reports it as absent from the registry, which is a
+            # diagnostic the reader can act on, not a KeyError from a loader.
+            child_path = yml_paths.get(namespace, {}).get(Path(step.id).stem)
+            if child_path is None:
+                continue
+            if child_path.resolve() in seen:
```
darkclad: The `seen` guard sits above the `workflows[(namespace, child_path.stem)] = child_source` on line 154, so it suppresses the registry entry along with the re-read. Dedup is right for the traversal, but the registry key includes the namespace, so a second call site under a different namespace needs its own entry even though the file has already been read.

```yaml
# root.wic
steps:
- id: leaf.wic
- id: leaf.wic
wic:
  steps:
    (2, leaf.wic):
      wic:
        namespace: other
```

with `leaf` discoverable under both namespaces:

```
registered workflow keys: [('global', 'leaf')]
SophiosError root.wic:3:3: error [wic013] process other/leaf is absent from the supplied registry
```

`ece5181` compiles the same input to `['root__step__1__leaf.wic', 'root__step__2__leaf.wic']`.

The comment above the guard says an undiscovered workflow is left unregistered so Resolve can report it -- but here the file was discovered and read, and the reader gets a "not in the registry" diagnostic about a file that is. Hoisting the assignment above the `seen` check looks like it covers it (it is keyed by namespace, so re-assigning under an already-present key is a no-op), unless the dedup is meant to be per-namespace for a reason I am not seeing.

tests/core/test_fuzzy_compile.py
```diff
@@ -88,17 +89,13 @@ def test_fuzzy_compile(self, data: st.DataObject) -> None:  # pylint: disable=to
...
-            yaml_tree = sophios.ast.merge_yml_trees(
-                yaml_tree_raw, {}, tools_cwl)
-            root_yml_dir_abs = yml_path.parent.absolute()
-            yaml_tree = sophios.ast.python_script_generate_cwl(
-                yaml_tree, root_yml_dir_abs, tools_cwl)
-
-            sophios.compiler.compile_workflow(yaml_tree, compiler_options, graph_settings,
-                                              yaml_tag_paths, [], [graph], {}, {}, {}, {},
-                                              tools_cwl, True, relative_run_path=True, testing=True)
+            bundle = frontdoor.bundle_from_source(
```
darkclad: This fails on the branch. `{'steps': {'0': []}}` is drawn from the schema and now reaches `parse` as written, which reports `wic003 EXPECTED_MAPPING`:

```
random_stepid.wic:2:8: error [wic003] step '0' must be a mapping, found a sequence
```

`wic003` is not in `TOLERATED_CODES`, so the draw fails the job. On `ece5181` the same input raised `ValueError: Error! If steps: tag is a Dictionary then all its values should be Dictionaries!`, which is listed verbatim in the `expected_messages` tuple at line 127.

So this is the situation the `wic026` and `wic028` comments inside `TOLERATED_CODES` already describe: a draw that was tolerated is now reported under a code, and giving a failure a code should not change which draws the job tolerates. Adding `SophiosErrorCode.EXPECTED_MAPPING` restores it -- and I would check its `wic002`/`wic004`/`wic005` siblings against the rest of that message list, since they cover the same shape errors and are presumably reachable from the same generator.

Unless the intent is that the front door should reject these outright and those `expected_messages` entries are now dead, in which case the entries they replaced should come out in the same commit so the two lists do not drift apart.

src/sophios/main.py
```diff
@@ -11,94 +11,43 @@
...
 from sophios.utils_yaml import wic_loader
 from . import input_output as io
 from . import post_compile as pc
-from . import ast, cli, compiler, inlineing, plugins, run_local, utils  # , utils_graphs
+from . import cli, compiler, plugins, run_local, utils  # , utils_graphs
+from .ir import frontdoor
+from .ir.frontdoor import SourceBundle
 from .schemas import wic_schema
-from .wic_types import (CompilerOptions, GraphData, GraphReps, GraphSettings, Json, RoseTree,
+from .wic_types import (CompilerOptions, GraphData, GraphReps, GraphSettings, Json,
```
darkclad: `StepId`, `Yaml` and `YamlTree` are all unused after this change -- pylint `W0611` three times, none of which fire on `ece5181`. `compiler.py:20` has the same with `RegistryKey`, and `cwl_subinterpreter.py:22` with `YamlTree`, plus `stepid` assigned at its line 115 and never read, plus the `import yaml` placed at line 1 above the stdlib block, which produces eight `wrong-import-order` messages.

Well inside the 9.25 floor so nothing breaks, but this PR's whole subject is deleting what stopped being reachable.

darkclad (commented):
Reviewed against the merge base `ece5181`. The retirement itself is thorough, and the `origin`-carrying port change is a real improvement over splitting `___` apart. Four behaviour regressions reproduce on this branch and not on the base, though, and two of them produce invalid CWL silently.

**Confirmed regressions** (each run both ways, on `ece5181` and on `bfb01f2`)

1. A cross-document `!*` leaks `{wic_alias: ...}` into the emitted child CWL. Inline on `ir/complete.py`.
2. `complete()` is not idempotent for `relative_run_path=False`, and `_compile_front` calls it three times, so a subworkflow's `run:` gets its prefix three times and no longer names the file that was written. Inline on `ir/complete.py`.
3. A `.wic` reached under two plugin namespaces is registered under only the first. Inline on `ir/frontdoor.py`.
4. `test_fuzzy_compile` fails on a draw the base explicitly tolerated. Inline on the test.

`tests/core/test_equivalences.py::test_a_meaning_preserving_rewrite_preserves_meaning[split]` and `tests/core/test_fuzzy_compile.py` both fail on this branch and pass on the base. The rest of `tests/core` is green (1250 passed). `test_scattering_scaling` and `test_run_local_python_api_restores_environment_after_run` fail identically on both branches in my environment (no cwltool on Windows), so those are mine, not yours. `mypy` is clean and `pylint` is 9.82 against a 9.25 floor.

**Docs left behind by the `--write_intermediate_wic` removal**

`cli.py` drops the flag, but `docs/advanced.md:73` still gives `sophios --yaml workflow.wic --generate_cwl_workflow --write_intermediate_wic` as the way to get the trees, and `docs/userguide.md:500` still describes it. `main()` parses with `parse_known_args`, so that command line does not error -- it silently does nothing, or gets passed through to the runner under `--passthrough_flags yes`. Both pages read as though the feature is still there.

**Smaller things**

- `--cwl_inline_subworkflows` no longer appends `_inline` to the output stem, so the inlined document now overwrites `{stem}.cwl` rather than sitting beside it. Self-consistent, since `run_local` follows `artifact.name`, but it is user-visible and it is not in the "Behaviour that changes" list alongside `--write_intermediate_wic`.
- `test_cwl_embedding_independence` dropped `assert len(node_data_lst[1:]) == len(yaml_forest_lst)` and gained a `continue` for includer fragments. A root whose every subworkflow classifies as a fragment now passes having compared nothing. Is the count assertion worth keeping in some form?
- New dead imports and one dead local, all flagged by pylint and none of them present on the base: `main.py` (`StepId`, `Yaml`, `YamlTree`), `compiler.py` (`RegistryKey`), `cwl_subinterpreter.py` (`YamlTree`, plus `stepid` assigned at line 115 and never read, plus the `import yaml` at line 1 tripping eight `wrong-import-order` messages). For a PR whose subject is removing what became unreachable, these are the same category.

src/sophios/ir/complete.py
vjaganat90: Right, and the seeding was the cause rather than `setdefault`. Lower no longer fills `emission.inputs` at all: every authored input naming a declared port becomes a `Binding`, and Complete writes each kind, so the seed was never load-bearing — it only made the default state invalid CWL. `setdefault` is correct as written once the slot starts empty. `test_equivalences[split]` is green again, and Emit now refuses any leftover `wic_*` key by name rather than writing it.

src/sophios/ir/complete.py
vjaganat90: Confirmed, and the arm was reading its own previous output. It now derives the leaf from the resolved identity, which is what it was always meant to spell and does not move; `multistep3` emits the single-prefix name again. I have pinned the idempotence with a test, since the phase has documented it since it was written and nothing checked it.

The larger point stands, though, and I am not fixing it here: `complete()` runs three times because it does two jobs, deriving graph facts and building the surface. Separating those so the surface is built once, inside Emit, is a structural change to the two largest files in `ir/` and this PR is already too big to carry it. It is the last of four follow-up PRs I have planned, with the idempotence property generalised over the generated space rather than the one case.

src/sophios/ir/frontdoor.py
vjaganat90: Yes. Registering now happens per `(namespace, stem)` and walking still per path — they are keyed differently on purpose, since a cycle is a cycle whichever namespace reaches it. Your reproducer is now a test.

tests/core/test_fuzzy_compile.py
vjaganat90: Added, and your question about the siblings was the useful half: `wic002`, `wic004` and `wic005` are reachable the same way, so all four went in. A free `!*` needed `wic025` too — the retired compiler accepted it by leaving the input unbound, so refusing it is the improvement and the draw is a defective document rather than a defective compiler.

I also removed the twelve message strings those codes replaced, each checked absent from `src/` first. Leaving them is how the shape codes came to be missing from the code list in the first place.

src/sophios/main.py
vjaganat90: All removed, along with the dead local and the import order. Both documentation pages went with them — `parse_known_args` meant that command line silently did nothing rather than erroring.

vjaganat90:
All four regressions are fixed, replies inline. Two were the same defect you diagnosed on `complete.py:191` — the seeded `in:` — which also accounted for the `[split]` failure.

On idempotence: agreed, and it is the finding I would keep. The three-call interleave exists because `complete()` does two jobs, deriving graph facts and building the emitted surface, and a phase that reads its own output cannot be safely repeated. I fixed the arm that drifts and pinned the property, but separating the two jobs so the surface is built once, inside Emit, is a structural change to the two largest files in `ir/` — this PR is already far past the size it should be and I would rather not carry it here.

It is planned as four follow-up PRs, in this order: name the emitted value set (it has three members today, and a type for it makes a `wic_alias` in `in:` unrepresentable rather than caught); one spelling for each derived name (`__step__{i}__{name}` is written independently in four places, `___` in six, while `NAMESPACE_SEPARATOR` sits unused in `types.py`); a distinct boundary-declaration type plus the reference check as a graph invariant; then Complete folded into Emit, with the idempotence property generalised over the generated space.

The common cause of all six defects this week is that the emitted surface has no specification. The retired compiler was one — it both produced the CWL and defined it — so deleting it removed the definition and the producer in the same commit. That is what the four PRs are for.

Also updated the body with the `--cwl_inline_subworkflows` stem change; you were right that it belonged in that list.

tests/core/test_pipeline.py
```diff
@@ -0,0 +1,152 @@
...
+    for module in PHASE_MODULES:
+        source = inspect.getsource(module)
+        tree = ast.parse(source)
+        assert not _mutable_module_bindings(source), module.__name__
+        assert not [node for node in ast.walk(tree) if isinstance(node, ast.Global)], module.__name__
+        imported = {
+            node.module for node in ast.walk(tree)
+            if isinstance(node, ast.ImportFrom) and node.module is not None
+        }
+        assert not any(name.endswith('compiler') for name in imported), module.__name__
```
darkclad: `c3cf131` deleted the retired-state half of this test, and I do not think the commit message's reasoning covers it. It argues the boundary scans stay because they "walk a transitive import graph and catch an indirect import a name check cannot see" -- true of the block that survives, but that block checks only that the five phase modules import nothing whose name ends in `compiler`. What went with the deletion was wider than that:

- `legacy_graph.py`, `inference.py` and `inlineing.py` still not existing;
- no module anywhere under `src/sophios` importing them (the surviving scan covers `PHASE_MODULES` only);
- nothing calling `compile_workflow`;
- `wic_types.py` not carrying `CompilerInfo`, `RoseTree`, `NodeData` or `EnvData` again.

None of those is implied by "no phase module imports `*compiler`". I grepped the rest of the suite and found no other guard -- the only remaining hits for those names are in prose docstrings.

That matters because the PR body lists, as one of the properties the repository enforces, that "retired modules, entry points, result shapes, and imports cannot reappear unnoticed", and `.github/workflows/lint_and_test.yml` describes this lane as "phase contracts plus the composed pipeline and retired-state boundary". After this commit the retired-state boundary is prose in two places and an assertion in none. Reintroducing `RoseTree` into `wic_types.py` is now a green build.

I have no argument with the rest of `c3cf131` -- `assert not hasattr(compiler, 'inference_rules')` and the `hasattr` sweeps over deleted API names really are asserting that a deleted name is still deleted, and `test_workflow_compile_boundary_hides_compiler_info`'s exact-field-set assertion is strictly stronger than the `hasattr` line removed next to it. It is specifically the file-existence, repo-wide import, `compile_workflow` call and `wic_types` shape scans that were carrying a claim nothing else makes.

Either restore that block (its own module would suit it better than a test named for mutable state), or drop the enforcement claim from the PR body and the CI comment so the three agree. I do not mind which.

tests/core/test_examples.py
```diff
@@ -12,21 +12,26 @@
...
+from sophios.lang.diagnostics import SophiosError
+from sophios.lang.error_codes import SophiosErrorCode
 from sophios.post_compile import verify_container_engine_config
-from sophios.wic_types import NodeData, StepId, Yaml, YamlTree, Json
+from sophios.ir.artifacts import CompilationArtifact
+from sophios.wic_types import StepId, Yaml, YamlTree, Json
 from sophios.utils_graphs import get_graph_reps
 
-from .test_setup import CorpusRegistry, corpus_registry, workflow_paths  # pylint: disable=unused-import
+from .test_setup import (CorpusRegistry, corpus_registry, workflow_paths,
```
darkclad: The `# pylint: disable=unused-import` is on line 33 now, so it no longer suppresses anything: pylint reports the import at line 32, the first line of the statement. `corpus_registry` is a deliberate pytest fixture re-export, so this one wants the comment moved back up to line 32 rather than the import removed.

While there: `9f9fd3d` cleaned up the four names I listed, but the branch is at 25 `unused-import`/`unused-variable` findings against 0 on `ece5181`, and four of those were created by this round rather than inherited:

```
src/sophios/ir/lower.py:18        InlineLiteral, RawCwlRef   (dead since _input_surface went)
tests/core/test_python_api_tool_builder.py:10  tool_builder_module   (dead since c3cf131)
tests/core/test_python_api_workflow.py:20      main_module           (dead since c3cf131)
```

The rest are in `src/sophios/utils.py` (5), `src/sophios/schemas/wic_schema.py` (6) and assorted test modules -- all files this PR trimmed heavily, so they are its leftovers too.

Not blocking, and the 9.83 score clears the 9.25 floor comfortably. Raising it only because "remove what stopped being reachable" is this PR's own subject, and `ece5181` was at zero.

darkclad (commented):
Round 2 on `0631b62`. All four regressions from the last round are fixed, each with a regression test, and several of the fixes went after the cause rather than the symptom -- `lower` no longer seeding `emission.inputs` at all is a better answer than making `setdefault` win, and `boundary_declaration` collapsing three divergent versions of the same rule was not something I asked for. Verified individually against the same repros:

| Finding | Status |
| --- | --- |
| `{wic_alias: ...}` in emitted child CWL | Fixed. The child now emits `{'source': 'part1__step__1__poly___value'}`, and `emit._emit_step` raises if a Sophios key reaches emission at all. |
| `complete()` not idempotent at `relative_run_path=False` | Fixed. `run:` is `multistep3__step__2__append_twice.wic___append_twice.cwl` and resolves against the written file. |
| One `.wic` under two namespaces | Fixed. Both keys register; the dedup is now path-keyed for the walk and namespace-keyed for the registry. |
| `test_fuzzy_compile` | Fixed, and `expected_messages` pruned to what is still reachable, so the two lists cannot drift. |

`tests/core`: 1251 passed, 2 failed -- `test_scattering_scaling` and `test_run_local_python_api_restores_environment_after_run`, both of which fail identically on `ece5181` here (no cwltool on Windows). `mypy` clean. The docs fix reads better than a flag removal usually does.

`1406920` (authored `outputSource` rewritten to the emitted step name) and `ba5aa49` (a mapping on a `string` port serialized as JSON rather than Python `repr`) are both real bugs I did not catch; the `--validate` test on the first is the right way to pin it.

One thing to settle before I approve, inline on `test_pipeline.py`, plus a lint nit inline on `test_examples.py`.

tests/core/test_pipeline.py
vjaganat90: Taking the second option: the claim is gone rather than the scan restored. The CI lane comment and this module's docstring now say what remains -- the phases' freedom from ambient state -- and neither says a retired name is watched. The body bullet went earlier, so the three agree now.

The reasoning, so it is on the record rather than only in a commit: a symbol that no longer exists cannot come back by accident. Reintroducing `RoseTree` into `wic_types.py` means somebody typing `RoseTree` into `wic_types.py`, which a reviewer sees in the diff. A test that names it fails only for that deliberate act, so it restates its own deletion and pays CI time for a diff a human is reading anyway. You are right that the surviving block is narrower than what went; I think the narrower claim is the only one worth asserting.

The file-existence checks are the clearest case: they assert three paths do not exist, in a repository where the commit that deleted them is the one under review.

tests/core/test_examples.py
vjaganat90: Fixed, and the comment was the interesting half -- `disable` on the second line of a wrapped import suppresses nothing, so it had been decorative since the import was wrapped.

Took the rest to zero rather than just the four: `utils.py`, `wic_schema.py`, `lower.py`, `benchmarks.py` and five test modules, nine in all. `ece5181` was at zero and this PR's subject is removing what stopped being reachable, so matching it seemed like the bar. Lint is 9.85 now.

src/sophios/ir/complete.py
darkclad: **Fixed** `lower` now seeds `emission.inputs` empty instead of with the source spelling, so the invalid state stops being the default rather than being overwritten later. Done in 020b73d, with 03d22d5 raising in `_emit_step` as a backstop; the split property is green.

src/sophios/ir/complete.py
darkclad: **Fixed** The leaf is derived from `emission.run.process_id.name` rather than read back out of `target`, so the arm no longer consumes its own output. Done in 7078fa0. `multistep3` now emits `multistep3__step__2__append_twice.wic___append_twice.cwl`, which resolves against the written file, and `test_completing_twice_is_completing_once` pins it.

src/sophios/ir/frontdoor.py
darkclad: **Fixed** The registry write moved above the `seen` guard, so the walk stays path-keyed and the registry namespace-keyed. Done in 905fe23. Both `('global', 'leaf')` and `('gpu', 'leaf')` register from one read.

tests/core/test_fuzzy_compile.py
darkclad: **Fixed** The four shape codes and `UNDEFINED_EDGE` joined `TOLERATED_CODES`, and `expected_messages` was pruned to what is still reachable. Done in 9a38a97. That closes the drift I was worried about, since neither list now carries an entry the other has taken over.

src/sophios/main.py
darkclad: **Fixed** Cleared here and across the other eight modules, and the branch is now at zero `unused-import`/`unused-variable` findings against zero on `ece5181`. Done in 9f9fd3d and 59fcb7a. Pylint is 9.85, slightly above the base's 9.84.

tests/core/test_pipeline.py
darkclad: **Diverted** You dropped the claim rather than restoring the scans, which was the option I said I did not mind taking. Done in 59fcb7a: the PR body bullet, the CI lane comment and this module's docstring now all describe only the ambient-state property, and the docstring records why a retired-name check was not worth keeping. The three agree, which is what I was after.

tests/core/test_examples.py
darkclad: **Fixed** The disable moved to its own `disable-next` line above the statement, with a note saying why the import is deliberate. Done in 59fcb7a, and the fixture still re-exports -- `test_examples.py` collects its 50 tests.

darkclad (approved):
LGTM on `05e79a4`. Everything from both rounds is resolved; I replied on each thread with what closed it.

Re-verified against the same repros, all four of the original regressions still fixed:

- cross-document `!*` emits `{'source': 'part1__step__1__poly___value'}`, not the alias;
- `run:` is `multistep3__step__2__append_twice.wic___append_twice.cwl` and resolves against the written file;
- one `.wic` under two namespaces registers under both;
- the fuzz lane tolerates the codes that replaced its messages.

`tests/core`: 1251 passed, 2 failed -- `test_scattering_scaling` and `test_run_local_python_api_restores_environment_after_run`, both of which fail identically on `ece5181` on my box for want of cwltool. The corpus compile lanes (`test_cwl_embedding_independence`, `test_inline_subworkflows`) are 19 passed, 3 skipped, and the skips now say which fragment they are rather than passing silently. `mypy` clean. `pylint` 9.85 with zero unused-import findings, against 9.84 and zero on the base.

On the retired-name scans: taking the claim out rather than putting the assertions back was the option I said I would not argue with, and `59fcb7a` did it in all three places at once -- PR body bullet, CI lane comment, module docstring -- with the reasoning written down instead of the claim just disappearing. That is the part I actually wanted.

`05e79a4` is a better answer to the `_inline` collision than restoring the suffix. Naming an artifact for its workflow and separating the lanes by directory keeps both properties, and the four-lanes-in-one-directory failure mode was the real risk rather than the name.

Noting your point that the big one is green rather than structurally finished. Nothing I can reach from outside shows it: behaviour is right at every door I tried, the `_emit_step` guard turns the class of failure it replaces from silent into loud, and the follow-up is yours to sequence. Approving on what the branch does today, not on the architecture being final.

